PR #13, opened 2021-11-22
Tweak the theme to match Civ3 in the ways that it hadn't been matched…

QuintillusCFC:
… yet.  Namely, get rid of the focus ring when focused, and change the color of the ibeam so you can see it again.

Also added hover/focus colors for build city buttons.  This is a common enough thing that it feels like it should not be one-off.  It is true that there are several graphics that have hover/pressed effects, and those graphics have different dimensions (and in this case, two buttons, each with different dimensions), so we'll probably still have multiple handlers.  Just, whenever we use one of those handlers, it shouldn't be coded separately of "this is how we handle this file."  I'm thinking some sort of utility class of, "hey, load this button", or perhaps better yet a Godot component that adds both buttons from a particular file.

Committing this as a branch to get in the habit of that.  Although it doesn't seem like a "going to take a bunch of commits to finish baking, needs a branch" type thing, I did notice when viewing the diffs that Godot hadn't saved the ibeam color, so that sort of thing might be caught in a PR.  We also got spaces instead of tabs in our BuildCityDialog somehow, but this commit converts them.

I'll probably merge this whenever I'm next online, if no one objects.

### DIFF
--- a/C7/C7Theme.tres
+++ b/C7/C7Theme.tres
@@ -8,6 +8,8 @@ expand_margin_top = 3.0
 expand_margin_bottom = 3.0
 
 [resource]
+LineEdit/colors/cursor_color = Color( 0, 0, 0, 1 )
 LineEdit/colors/font_color = Color( 0, 0, 0, 1 )
 LineEdit/colors/selection_color = Color( 0.709804, 0.807843, 0.709804, 1 )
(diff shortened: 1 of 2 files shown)


QuintillusCFC:
Forgot to actually open the PR after uploading the branch.  The changes look right, though, so going to go ahead and merge it.

PR #27, opened 2021-11-28
Cities on map

QuintillusCFC:
Not done but the changes from development are becoming significant and Git is becoming unhappy with them.  Want to see the diff and likely merge before adding anything else.


QuintillusCFC (commented):
The changes look exactly like I expected them to.  Not sure why local Git was being a git.

PR #31, opened 2021-11-29
FLC Import Performance

QuintillusCFC:
Boosts FLC import performance by 40% (what used to take 1 second now takes 0.6 seconds).

### DIFF
--- a/C7/project.godot
+++ b/C7/project.godot
@@ -31,6 +31,10 @@ window/size/height=768
 
 theme/custom="res://C7Theme.tres"
 
+[network]
+
+limits/debugger_stdout/max_chars_per_second=16384
+
 [rendering]
 
 environment/default_environment="res://default_env.tres"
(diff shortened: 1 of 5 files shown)


C7/project.godot
```diff
@@ -31,6 +31,10 @@ window/size/height=768
 
 theme/custom="res://C7Theme.tres"
 
+[network]
+
+limits/debugger_stdout/max_chars_per_second=16384
```
QuintillusCFC, line 36: This lets the console print more stuff.  When I was printing out all the pictures we were importing, the console couldn't handle it until I added this.  Might be too much on really slow systems, but it seems like a generally good thing and didn't cause any problems on my 9.9-year-old system.

QuintillusCFC (commented):
Looks like what I expected.  Removed a few debugging sysouts.

PR #43, opened 2021-12-04
Mountains and hills part two

QuintillusCFC:
This includes:

- Ranges of mountains/hills/volcanos
- Volcano graphics (mechanics are for 0.11 "Lilliput")
- Snow-capped mountains
- Forest/jungle variants of hills/mountains/volcanos.  These won't appear until we have forests and volcanos

It does not include landmark terrain variants, or snow volcanos (for which a PCX exists, but which don't exist in Civ3).


QuintillusCFC (commented):
Reviewed and made two changes in the most recent commit.

PR #51, opened 2021-12-12
Import real Civ3 terrain types

QuintillusCFC:
This replaces the dummy terrain types with actual, imported-from-civ3 terrain types.  This means we will now have all terrain types (including flood plains, sea, and ocean, which had not been dummied out).  It also sets the stage for easily adding more terrain attributes to C7 terrain types as we're ready for them.

I went ahead and added all the Civ3 TERR attributes, although not all convenience methods we may want for them, following the Apolyton documentation, in a few cases with comments clarifying them based on code from my editor.  One notable call-out is that there is a variable-length array in TERR, whose length is based on the number of resources in-game.  This has some special handling, and subsequent attributes use an offset from the end of that array, rather than the base offset for the TERR section.

### DIFF
--- a/C7Engine/EntryPoints/UnitInteractions.cs
+++ b/C7Engine/EntryPoints/UnitInteractions.cs
@@ -126,13 +126,13 @@ namespace C7Engine
                             return;
                         }
 
-                        var newLoc = gameData.map.tileAt(dx + unit.location.xCoordinate, dy + unit.location.yCoordinate);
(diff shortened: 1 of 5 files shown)


QuintillusCFC (commented):
Reviewed after some sleep.  Looks like what I'd expect.  Going to leave the tile print-out for now since it's our console confirmation that we're running the import.  Probably take it out when the next section is imported.

PR #52, opened 2021-12-12
Add some basic city production and growth

QuintillusCFC:
And have the city labels on the map reflect that status accurately.

So far the production chain is all hard-coded.  But it's still kind of cool to see units get produced, and be able to build another city.  It was also cool seeing that growth and production of a city size consuming unit (Settler) worked flawlessly.

I did notice that forests and marshes are currently free to move across.  Gonna have to look into why that is.


QuintillusCFC (commented):
Looks good, only minor thing is spaces vs tabs in GameData.  Now that I look at it more, I see Visual Studio has auto-detect on for that and my local VSCode appears to be using spaces, whereas our discussion in early November favored tabs.  Ugh.  Changed some VSCode settings so it should hopefully favor tabs more in the future.  Not going to try to resolve all of that in this branch.

PR #53, opened 2021-12-12
Allow setting the zoom to 100% or 50% with the Z key.

QuintillusCFC:


QuintillusCFC (commented):
Looks as expected, has been working properly for a couple days.

Noticed GitHub displays tabs as 8 spaces instead of my preferred 4.  But the indentation is tabs which is fine, just a GitHub setting to change.

PR #54, opened 2021-12-13
Reformat QueryCiv3 biq section logic to support easy data slurping

lucienmaloney:
The following BIQ data sections have their data structures completely set up: BLDG, CTZN, CULT, DIFF, ERAS, ESPN, EXPR, GOOD, TERR.


lucienmaloney:
Now also populated: TECH, TFRM, WSIZ, WCHR, TILE, CONT, SLOC, UNIT, CLNY

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+        public bool[,] TerrGood; // which resources are allowed on which types of terrain
+        public GOVTGOVT[,] GovtGovt; // relationships between governments
+
+        public unsafe BicData(byte[] bicBytes)
         {
             Bic = new Civ3File(bicBytes);
+
+            fixed (byte* bytePtr = bicBytes)
+            {
+                int offset = 736; // byte index of the first header in BIQ files
```
QuintillusCFC: This is correct as the first place at which rules/maps/custom player data/scenario properties may exist.  But there is the VER# header before that, and we might care about that someday, though I agree with skipping it for now.  Maybe updating the comment to clarify "first rule header (likely BLDG)" would be better?  My first thought on seeing this was, "why 736?  is that correct? what does that point to".

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+                string header;
+                int count = 0;
+                int dataLength = 0;
+
+                while (offset < bicBytes.Length - 12) { // Don't read past the end
```
QuintillusCFC, line 66: Just curious, why do we need the - 12?  Shouldn't it wind up right at the end?  My best guess is 12 = header + count + sizeOfData, but if that's the case, the -12 would only work if the sizeOfData is zero.

(this might also be answered by the time I read all of the code)

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+                    header = Bic.GetString(offset, 4);
+                    count = Bic.ReadInt32(offset + 4);
+                    offset += 8;
+
+                    // Section data structures are stored in the BiqSections/ folder
+                    // We can divide the BIQ sections into two types: static and dynamic
+                    // Static have a fixed length always, which means they can be read directly memory-copied into our structs with no special logic
+                    //   The static sections are: BLDG, CTZN, CULT, DIFF, ERAS, ESPN, EXPR, GOOD, TECH, TFRM, WSIZ, WCHR, TILE, CONT, SLOC, UNIT, CLNY
+                    // Dynamic sections have at least one component with varying length, and so require multiple structs and special logic
+                    //   The dynamic sections are: GOVT, RULE, PRTO, RACE, TERR, FLAV, WMAP, CITY, GAME, LEAD
```
QuintillusCFC, line 77: I like that you've documented which sections have dynamic length here.  I'm not sure I've ever seen that listed in one place (without having to check each section).

QueryCiv3/BiqSections/Bldg.cs
```diff
@@ -0,0 +1,126 @@
...
+{
+    [StructLayout(LayoutKind.Sequential, Pack=1)]
+    public unsafe struct BLDG
+    {
+        public int Length;
+
+        private fixed byte Text[128];
+        public string Description { get => Util.GetString(ref this, 4, 64); }
+        public string Name { get => Util.GetString(ref this, 68, 32); }
+        public string CivilopediaEntry { get => Util.GetString(ref this, 100, 32); }
```
QuintillusCFC, line 14: Let's see if my interpretation is correct: The variables that don't have a getter defined are the ones that take up space in memory, so the memcpy works because the only fields without a getter are the only ones that are in the Civ3 BIQ format.

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+                    switch (header) {
+                        case "BLDG":
+                            dataLength = count * sizeof(BLDG);
+                            Bldg = new BLDG[count];
+                            fixed (void* ptr = Bldg) {
+                                Buffer.MemoryCopy(bytePtr + offset, ptr, dataLength, dataLength);
+                            }
+                            break;
```
QuintillusCFC, line 85: It's probably a good idea to make a post sharing the technique for this (including a bit from one of the sections, notably the StructLayout in BLDG).  It makes sense reading through it, even though I hadn't seen it before, but I had learned about fixed pointers and unsafe code blocks a few weeks ago for C7, and am used to pointers from C.  It may not be familiar for those who don't have a C#, C++, or C background though.

We have a Godot Tips and Tricks thread, but a new Technical Knowledge Sharing (or similar) thread is probably more appropriate.

I think it is an appropriate technique here as a great way to quickly set up the data import without as much manual setup.

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+                        case "CITY":
+                            dataLength = -7;
```
QuintillusCFC, line 87: Why are we setting dataLength to -7 here?  I see it in a bunch of places, but 7 is an odd number.  There should probably be a constant with a name such as LENGTH_OF_SOMETHING that makes it clear what the 7 represents.

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+                            int govtgovtRowLength = count * sizeof(GOVTGOVT);
+
+                            fixed (void* ptr = Govt, ptr2 = GovtGovt) {
+                                byte* govtPtr = (byte*)ptr;
+                                byte* govtgovtPtr = (byte*)ptr2;
+                                byte* dataPtr = bytePtr + offset;
+                                for (int i = 0; i < count; i++) {
+                                    Buffer.MemoryCopy(dataPtr, govtPtr, 400, 400);
+                                    Buffer.MemoryCopy(dataPtr + 400, govtgovtPtr, govtgovtRowLength, govtgovtRowLength);
+                                    Buffer.MemoryCopy(dataPtr + 400 + govtgovtRowLength, govtPtr + 400, 76, 76);
```
QuintillusCFC: Minor: I can follow what's going on here having some familiarity with it, and having the Apolyton documentation up side-by-side.  But what is reading between the lines is that the GOVTGOVT Relations structure is within the GOVT structure, and the 400 is the GOVT section size before that, and the 76 is the GOVT section after that.  It might be worth having a constant named to indicate this.  GOVT_BEFORE_RELATIONS_SIZE, maybe, for the first?

(There is a bit of a theme in some of my comments of eliminating magic numbers, which is where what the numbers represent is only known if you have contextual knowledge beyond what the the code base indicates)

QueryCiv3/BiqSections/Govt.cs
```diff
@@ -0,0 +1,68 @@
...
+        public int Length;
+        public int DefaultType;
+        public int TransitionType;
+        public int RequiresMaintenance;
+        public int Toggle1; // ??? 0: Republic/Democracy, 1: Other
+        public int TilePenalty;
+        public int TradeBonus;
+
+        private fixed byte Text[352];
+        public string Name { get => Util.GetString(ref this, 28, 64); }
```
QuintillusCFC, line 24: I'll ask this here in case I forget to try it locally: is it possible to have this be Util.GetString(ref Text, 0, 64); or similar?  It seems like it would be easier than counting up the bytes before hand, and easier to verify that the offsets are right (multiples of 32 versus multiples of 32, plus 28), but I'm not sure it's possible.

QueryCiv3/BiqSections/Govt.cs
```diff
@@ -0,0 +1,68 @@
...
+        public int ImmuneTo;
+        public int DiplomatsAre;
+        public int SpiesAre;
+        public int NumberOfGovernments;
+
+        /*
+            Dynamic length gap
+            In BIQ files, there are 3 integers here for each government in NumberOfGovernments
+            Data is instead stored in 2d array of GOVTGOVT
+        */
```
QuintillusCFC, line 46: Thanks for leaving a comment here about that!

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+                                // TERR contains dynamic data, so it can't be read in as a block. Instead, read in data for each TERR
+                                for (int i = 0; i < count; i++) {
+                                    // Copy first 8 bytes into TERR struct:
+                                    Buffer.MemoryCopy(bytePtr + offset + i * terrLen, terrBytePtr, 8, 8);
+                                    // Get TerrGood flags, dynamic data which determines which resources are allowed on which terrain types
+                                    for (int j = 0; j < goodCount; j++) {
+                                        TerrGood[i,j] = Util.GetFlag(*(bytePtr + offset + i * terrLen + 8 + (j / 8)), j % 8);
+                                    }
+                                    // Copy remaining bytes into TERR struct:
+                                    Buffer.MemoryCopy(bytePtr + offset + (i + 1) * terrLen - sizeof(TERR) + 8, terrBytePtr + 8, sizeof(TERR) - 8, sizeof(TERR) - 8);
```
QuintillusCFC: Minor: It might be a good idea to use a local variable or two and/or parentheses here (and maybe in some of the other offset calculations nearby) to make things a bit clearer.  I can see here that you're starting at the bytePtr (the source we're copying from, adding the offset from the header, going to the _next_ terrain, subtracting a whole terrain length, and then adding 8 to account for what you've already done.  But I'm implicitly adding parens to show that the + 8 is outside the parens (it is not (sizeof(TERR) + 8), which would subtract too much), and it's a fair amount of terms to keep track of inline.  As a colleague once told me when I was getting a little too low-level, I am not a compiler.

QueryCiv3/BiqSections/Tile.cs
```diff
@@ -0,0 +1,58 @@
...
+        private fixed byte UnknownBuffer3[1];
+        public byte C3CTerrain;
+        private fixed byte UnknownBuffer4[2];
+        public short FogOfWar;
+
+        private fixed byte C3CBonuses[4];
+        public bool BonusGrassland      { get => Util.GetFlag(C3CBonuses[0], 0); }
+        public bool PlayerStart         { get => Util.GetFlag(C3CBonuses[0], 1); }
+        public bool SnowCappedMountain  { get => Util.GetFlag(C3CBonuses[0], 2); }
+        public bool PineForest          { get => Util.GetFlag(C3CBonuses[0], 3); }
```
QuintillusCFC: These bonuses are not actually in sequential bits.  They should be (format byte, bit):

BonusGrassland - 0, 0 [correct]
PlayerStart - 0, 3
SnowCapped - 0, 4
PineForest - 0, 5

In addition, not in Apolyton documentation, there are:

RiverNorth - 1, 0 [256 decimal]
RiverWest - 1, 1 [512 decimal]
RiverEast - 1, 2 [1024 decimal]
RiverSouth - 1, 3 [2048 decimal]
IsLandmark - 3, 1 [0x2000]

IsLandmark is the important one among these.  I am not sure exactly how the river info here interacts with the RiverConnectionInfo.  But what I do know is that RiverConnectionInfo is the edges of the tile; these are the corners.  I'm about 92% sure that if you go into the Firaxis editor, start a new map, put down some plains, and then use the river overlay to place a river only at the intersection of 4 tiles, it uses this area to store it.  Although that could use some additional verification.

QueryCiv3/BiqSections/Tile.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public byte Border;
+        public int Resource;
+        public byte TextureLocation; // Between 0 and 80 inclusive (9*9 grid)
+        public byte TextureFile; // 0: xtgc, 1: xgpc, 2: xdgc, 3: xdpc, 4: xdgp, 5: xggc, 6: wcso, 7: wsss, 8: wooo
+        private fixed byte UnknownBuffer[2];
+        public byte OverlayFlags;
+        public byte Terrain;
+        public byte BonusFlags;
+        public byte RiverCrossingFlags;
```
QuintillusCFC: Subject to confirmation/disproval, I have a comment in my editor for this byte:

`//appears to always be zero.  can use river connection info to figure out where rivers are instead`

Since there are at least three places referring to river info, we should probably be verbose and proactive about documenting what we can about them here.

QueryCiv3/BiqSections/Tile.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        private fixed byte C3COverlays[4];
+        public bool Road            { get => Util.GetFlag(C3COverlays[0], 0); }
+        public bool Railroad        { get => Util.GetFlag(C3COverlays[0], 1); }
+        public bool Mine            { get => Util.GetFlag(C3COverlays[0], 2); }
+        public bool Irrigation      { get => Util.GetFlag(C3COverlays[0], 3); }
+        public bool Fortress        { get => Util.GetFlag(C3COverlays[0], 4); }
+        public bool GoodyHut        { get => Util.GetFlag(C3COverlays[0], 5); }
+        public bool Pollution       { get => Util.GetFlag(C3COverlays[0], 6); }
+        public bool BarbarianCamp   { get => Util.GetFlag(C3COverlays[0], 7); }
```
QuintillusCFC, line 46: Additional flags:

Craters - 1, 0
Barricade - 3, 4 [0x10000000]
Airfield - 3, 5 [0x20000000]
Radar Tower - 3, 6 [0x40000000]
Outpost - 3, 7 [0x80000000]

QueryCiv3/BiqSections/Tile.cs
```diff
@@ -0,0 +1,58 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace QueryCiv3.Biq
+{
+    [StructLayout(LayoutKind.Sequential, Pack=1)]
+    public unsafe struct TILE
+    {
+        // Not 100% sure the flags are correct. The Apolyton documentation gets confusing here
```
QuintillusCFC: Thanks for calling this out.  It's always better to note where there's uncertainty, and in this case it made sure I compared to both the Apolyton docs and the documentation I had accumulated over the years in editor code.

QueryCiv3/BiqSections/Unit.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public int Length;
+
+        private fixed byte Text[32];
+        public string Name { get => Util.GetString(ref this, 4, 32); }
+
+        public int OwnerType; // 0: None, 1: Barbarian, 2: Civ, 3: Player
+        public int ExperienceLevel;
+        public int Owner;
+        public int Unit;
```
QuintillusCFC: This should be UnitType, most likely.  It's a PRTO reference, but with just Unit as the name, it's ambiguous whether it's a PRTO reference or perhaps a reference to another UNIT type (PRTO/UNIT can be confusing in general in Civ3 since they're both generically referred to as "unit" in different contexts).

QueryCiv3/BiqSections/Unit.cs
```diff
@@ -0,0 +1,27 @@
...
+        public int AIStrategy;
+        public int X;
+        public int Y;
+
+        private fixed byte Text2[57];
+        public string PTWName { get => Util.GetString(ref this, 64, 57); }
```
QuintillusCFC: Another place where Unit is confusing.  Checking my editor code, PTWName is what is used (in PTW+) for custom-named units.  I _believe_ the "regular" name is unused (can you set custom unit names in Vanilla?  I am not sure you can).  It might be worth naming this variable "CustomUnitName", or similar, with a comment that it's the "PTWUnitName".  Or perhaps just adding a comment that this one, not Name, is the one that is used.  Or maybe a comment on Name saying that this one is used instead.

Basically, something somewhere to indicate that this one is the one that is used.

QueryCiv3/BiqSections/Wchr.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace QueryCiv3.Biq
+{
+    [StructLayout(LayoutKind.Sequential, Pack=1)]
+    public unsafe struct WCHR
+    {
+        public int Length;
+        public int Climate; // 0: Arid, 1: Normal, 2: Wet, 3: Random
```
QuintillusCFC: I like the Apolyton documentation's differentiation here of "selectedClimate"/"actualClimate".  We had some confusion over "terrain"/"overlayTerrain", and now call is "baseTerrain/overlayTerrain" since when there are two variants of a variable, it's easy to confuse them without the variant name being specified.

I do like that you've included the documentation of which type is which.  I wonder if long-term we'll want enums for those?  That would probably make more sense in the C7 side than the Civ3 import side though.

QueryCiv3/Bic.cs
```diff
@@ -9,9 +10,253 @@ public class BicData
...
+                            dataLength = count * sizeof(WSIZ);
+                            Wsiz = new WSIZ[count];
+                            fixed (void* ptr = Wsiz) {
+                                Buffer.MemoryCopy(bytePtr + offset, ptr, dataLength, dataLength);
+                            }
+                            break;
+                        default:
+                            // Once every BIQ section is set up with the correct associate struct(s), the default case where a header isn't found
+                            // should never occur. However, for now, if a header isn't found, we'll just increment by 1 to keep searching
+                            dataLength = -7; // We added 8 earlier, so subtract 7
```
QuintillusCFC, line 272: Okay, I see the 7 here with a comment.  But after reading the comment (and through the offset += dataLength, I am fairly sure that this is not the right way to skip ahead.

We do indeed want to keep searching here.  Line 257 adds dataLength to offset, to continue on to the next section.  For most sections (let's choose World Size as an example), that will change it by the size of the section, times the number of instances of the section, and as you note we've already added 8 for the section header/# of instances.

But let's say we don't know about WSIZ yet, and there are 5 world sizes.  We've added 8 for WSIZ and the int 5.  We subtract 7 here.  So the next time, it's going to see "SIZ\0 ", i.e. SIZ plus a null byte from the first byte of the integer.

We don't actually want it to jump ahead by one byte though.  We should jump ahead of the entire WSIZ section.

To do that, we can leave the 8, but we also need to, for each instance of WSIZ, read the "length of section" integer (which could vary for each instance), and add that many bytes to the offset.

It's a little more complex, but should be considerably more efficient than checking each byte.  It also avoids potential errors if one of our expected headers shows up in the data of another section.  Which isn't likely, but the advantage of this header-iteration approach over just searching as we did before is that it avoids any possibility of failure for that reason.

QueryCiv3/GameSections.cs
```diff
@@ -55,7 +55,7 @@ public WrldSection(SavData sav)
         }
         public short ContinentCount { get => Data.Sav.ReadInt16(Offset1+4); }
         public int WsizID { get => Data.Sav.ReadInt32(Offset1+234); }
-        public WsizSection WorldSize { get => Data.Bic.Wsiz[WsizID]; }
+        public Biq.WSIZ WorldSize { get => Data.Bic.Wsiz[WsizID]; }
         public int Height { get => Data.Sav.ReadInt32(Offset2+8); }
```
QuintillusCFC, line 59: Correct me if I'm wrong: Aren't most of these rendered obsolete by the new sections, once we refactor references to use the new code?

It's okay if that's coming in the future, just wanted to check whether the majority of this file remains necessary since my first thought is it probably isn't anymore.

QueryCiv3/Util.cs
```diff
@@ -1,11 +1,18 @@
...
+using System.Text;
 
 namespace QueryCiv3
 {
     public class Util
     {
+        // Encoding code page ID; 1252 is Civ3 encoding for US language version
+        //   See https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=net-5.0 for other values
+        // NOTE: .NET 5 and later require a Nuget package and Encoder registration for these older encoding pages
+        static public Encoding Civ3Encoding = Encoding.GetEncoding(1252);
```
QuintillusCFC, line 14: Minor: Other code pages I know Civ uses are 1251 for Russian, and GBK for Chinese (which based on your link, the closest would be 936 in .NET).  The GBK I am only about 90% sure of; it matches screenshots I've seen of Chinese language Civ3 scenarios when I run it locally, but I don't know Chinese so I can't say with certainty that there isn't another, more complete encoding that's more appropriate.

For now, I'm thinking documenting known encodings used by Civ3 is sufficient; if the project takes off, we can use settings to swap out this variable as needed easily enough.

QuintillusCFC (changes requested):
Mostly minor comments (including a few good jobs).  I noted a few places where there are some additional flags, one case where the flags aren't accurate, and a few other suggestions, often around documenting BIQ peculiarities.

QueryCiv3/Bic.cs
lucienmaloney: Yeah, this section had _way_ too much magic and complexity, so good catch for calling me out on it. It should be a lot better now.

QueryCiv3/BiqSections/Bldg.cs
lucienmaloney: Yep, the getters don't affect the memory layout, which is entirely determined by the ints and bytes and so on. We can't define flag fields or string fields because bit types don't exist and strings are variable length, so for those sections of memory, we can just define a buffer and use getters to parse out the data we need as necessary.

QueryCiv3/Bic.cs
lucienmaloney: I'd be happy to do a write-up about the memory-struct system going on here. Can you send me a link for the right place to make the post? I assume somewhere in the CFC forums, but there's a lot of them there.

QueryCiv3/BiqSections/Govt.cs
lucienmaloney: I don't think it is possible, because fixed fields can't be passed as arguments and I don't think there's a way to automatically reference Text's offset in the overall struct automatically. But yeah, I agree it's not ideal, and it might be worth looking into at least lessening the magic and improving the readability of it. Probably low priority though.

QueryCiv3/Bic.cs
lucienmaloney: The -12 is that so when the loop is in linear-search mode, the offset isn't overflown(?) when trying to get the next potential header and count:
`header = Bic.GetString(offset, 4);
count = Bic.ReadInt32(offset + 4);`
Really, it should be -8, since those together add up to only 8, but either way, it should only be a temporary thing. The linear-search nature of the loop will only be there until the rest of the BIQ sections are defined, at which point the loop should be able to jump from one header to the next with no searching necessary. Then we can change the default case to throwing an exception, because the default would only occur in situations where something went wrong with the memory reading (bad file).

QueryCiv3/Bic.cs
lucienmaloney: Like a mentioned in one of the above comments, the -7 thing is only temporary and won't be an issue once all sections have their data population. For your WSIZ example, the 7 actually isn't involved at all. When the first WSIZ header is encountered, the offset adds 8 like you say to skip past the header and length. But then the next offset change is for the entire WSIZ section and all its members. So there isn't any byte searching at all: the offset jumps straight to the location of whatever the next header is. Linear searching (where the -7 + 8 = 1 is involved) only begins to happen currently when a header is encountered that we don't yet have the struct(s) set up for, so we don't know how far to jump ahead for the next section. But once they're all set up (hopefully I can get them all in tomorrow), linear byte searching can be removed entirely.

QueryCiv3/GameSections.cs
lucienmaloney: For right now, I haven't really looked at much of the SAV logic other than changing around some names to make sure the code still compiles, so I can't say for sure. But my plan is that once I'm done filling out the BIQ sections, I'll work on implementing the same struct-memory-copying strategy for SAV, so that'll involve the refactoring, yeah.

QueryCiv3/Util.cs
lucienmaloney: Civ encoding is not something I'm familiar with at all. I didn't actually write those comments; I was just moving the string getting logic from somewhere else into a static method of the Util class so that I could call it from the new structs. So I'll defer to whoever actually wrote the above comments on what the most appropriate action is here, hehe.

lucienmaloney:
I think I've fixed/improved all of the actionable stuff from the comments, so this should be good to merge now, I think. But if you want to just wait until I get the rest of the Biq sections in before that, that works too. I'll just end up having to make a second pull request anyway.

QueryCiv3/Bic.cs
QuintillusCFC: Right, it's the linear searching where this would occur, and my thoughts were related to the case where we don't know a header, and thus would (currently) linearly search.  My thinking was that even if we don't know the contents of a section, we know enough to be able to jump ahead in increments of _dataLength_ bytes (as _dataLength_ is always the first 4 bytes of any section, which would be much more efficient, and safer.

But if we're going to have all the rest of the sections set up this week, it's not worth optimizing that in the meantime.  That's probably the best part of this technique, it lets the basic structure get set up quite quickly.

Marking as resolved since this will be obsolete soon.

QueryCiv3/Bic.cs
QuintillusCFC: Yeah!  We have our own sub-forum at https://forums.civfanatics.com/forums/civ3-future-development.604/ .  It should be publicly accessible now.

QueryCiv3/Bic.cs
QuintillusCFC: That makes more sense with the 8, and I agree that the check (in its current form, at least) will be temporary.  We might still need a check, but it could be a check against the whole length (assuming the offset winds up being equal to the whole length after the last section, whichever section that is).

QueryCiv3/BiqSections/Bldg.cs
QuintillusCFC: Neat!  Score a point for C# as a cool language!

QueryCiv3/BiqSections/Govt.cs
QuintillusCFC: That's kind of what I expected.  I might poke around at it some time, but agree it's not a high priority, particularly as these structures are not going to change (or at least not unless we add PTW support, and we'll have other things to think about if we do that).

QueryCiv3/Bic.cs
QuintillusCFC: That is way more readable!  Thanks for updating that!

QueryCiv3/GameSections.cs
QuintillusCFC: Ah yes, I see now.  All the updates are for the BIQ contents, which in some cases (CITY, TILE, etc.) have equivalents in the SAV.  Makes sense now.

QueryCiv3/Util.cs
QuintillusCFC: Oh, yup, looks like Puppeteer wrote that a month ago.  No need to modify it then, we'll do it if/when we get to that point.  Also no reason to be familiar with code pages nowadays unless you're working on a 20 year old game, or the internals of Windows with UTF-16.

QuintillusCFC (approved):
Looks good, some nice readability improvements.  Resolved all the conversations other than the one with a link to the forums, so that one is visible.

I'd lean towards merging the changes over adding more change to the same branch.  But I'm generally the smaller-pull-request advocate on whichever team I am on.

lucienmaloney:
Yeah, good point about the more incremental PR sizes. We'll go for merge then, and I'll get started on another branch for the remaining sections. You'll have to do the merge for me though since I don't have access.

PR #55, opened 2021-12-14
Barbarians on the prowl

QuintillusCFC:
Add a basic AI for barbarians where they randomly prowl around, and give a 7% chance of a new barbarian from each barbarian camp.


JimOfLeisure:
Looks good to me. I checked out the branch and it works. Not sure if I should just merge it or what.

PR #56, opened 2021-12-16
Completed the remainder of the BIQ structs

lucienmaloney:
All the known, documented data for the BIQ format should now be accessible in the BiqData class, one way or another. Most of the stuff is stored in the arrays of section header structs themselves, but all dynamic data is stored in separate 2-dimensional arrays.

There's room for improvement to be made here. The code is a bit wet in places and better simplification and documentation would always be good. But for now, at least, this gets us access to all of the data in biq files. I've only tested it on the regular conquests.biq and the WWII one, but those at least work for sure.

### DIFF
--- a/QueryCiv3/Biq.cs
+++ b/QueryCiv3/Biq.cs
@@ -46,17 +46,17 @@ namespace QueryCiv3
             A jagged array means that the second dimension can vary between components eg. for RaceCityName, each civ can have a different # of Cities
         */
         public bool[,] TerrGood; // which resources are allowed on which types of terrain
-        public GOVTGOVT[,] GovtGovt; // relationships between governments
-        public RACECITYNAME[][] RaceCityName; // the list of city names for each civ
-        public RACEERA[,] RaceEra; // the file names for each era for each civ
-        public RACELEADERNAME[][] RaceGreatLeaderName; // the great leaders for each civ
-        public RACELEADERNAME[][] RaceScientificLeaderName; // the scientific leaders for each civ
+        public GOVT_GOVT[,] GovtGovt; // relationships between governments
+        public RACE_City[][] RaceCityName; // the list of city names for each civ
+        public RACE_ERAS[,] RaceEra; // the file names for each era for each civ
+        public RACE_LeaderName[][] RaceGreatLeaderName; // the great leaders for each civ
+        public RACE_LeaderName[][] RaceScientificLeaderName; // the scientific leaders for each civ
         public int[][] CityBuilding; // Building IDs in each city
         public int[][] WmapResource;
         public int[][] PrtoPrto; // Stealth unit targets per unit
-        public LEADPRTO[][] LeadPrto; // starting unit data for each leader
+        public LEAD_Unit[][] LeadPrto; // starting unit data for each leader
         public int[][] LeadTech; // starting tech data for each leader
-        public RULECULT[][] RuleCult; // culture level names per rule
+        public RULE_CULT[][] RuleCult; // culture level names per rule
         public int[][] RuleSpaceship; // spaceship quantity requirements per rule
         public int[][] GameCiv; // Playable civs for game
         public int[][] GameAlliance; // Civ alliances for game
@@ -271,8 +271,8 @@ namespace QueryCiv3
                             int govtLen = Bic.ReadInt32(offset) + 4;
                             dataLength = count * govtLen;
                             Govt = new GOVT[count];
-                            GovtGovt = new GOVTGOVT[count, count];
-                            int govtgovtRowLength = count * sizeof(GOVTGOVT);
+                            GovtGovt = new GOVT_GOVT[count, count];
+                            int govtgovtRowLength = count * sizeof(GOVT_GOVT);
 
                             fixed (void* ptr = Govt, ptr2 = GovtGovt) {
                                 byte* govtPtr = (byte*)ptr;
@@ -292,7 +292,7 @@ namespace QueryCiv3
                         case "LEAD":
                             dataLength = 0;
                             Lead = new LEAD[count];
-                            LeadPrto = new LEADPRTO[count][];
+                            LeadPrto = new LEAD_Unit[count][];
                             LeadTech = new int[count][];
 
                             fixed (void* ptr = Lead) {
@@ -303,8 +303,8 @@ namespace QueryCiv3
                                 for (int i = 0; i < count; i++) {
                                     Buffer.MemoryCopy(dataPtr, leadPtr, LEAD_LEN_1, LEAD_LEN_1);
                                     leadPtr += LEAD_LEN_1;
-                                    LeadPrto[i] = new LEADPRTO[Lead[i].NumberOfStartUnitTypes];
-                                    rowLength = Lead[i].NumberOfStartUnitTypes * sizeof(LEADPRTO);
+                                    LeadPrto[i] = new LEAD_Unit[Lead[i].NumberOfStartUnitTypes];
+                                    rowLength = Lead[i].NumberOfStartUnitTypes * sizeof(LEAD_Unit);
                                     fixed (void* ptr2 = LeadPrto[i]) Buffer.MemoryCopy(dataPtr + LEAD_LEN_1, ptr2, rowLength, rowLength);
                                     dataPtr += LEAD_LEN_1 + rowLength;
 
@@ -349,9 +349,9 @@ namespace QueryCiv3
                         case "RACE":
                             dataLength = 0;
                             Race = new RACE[count];
-                            RaceCityName = new RACECITYNAME[count][];
-                            RaceScientificLeaderName = new RACELEADERNAME[count][];
-                            RaceGreatLeaderName = new RACELEADERNAME[count][];
+                            RaceCityName = new RACE_City[count][];
+                            RaceScientificLeaderName = new RACE_LeaderName[count][];
+                            RaceGreatLeaderName = new RACE_LeaderName[count][];
                             /*
                                 For getting dynamic race data, we need to know the number of eras as defined earlier
                                 Presumably this means that the ERAS section of BIQ will always appear before the RACE section
@@ -360,8 +360,8 @@ namespace QueryCiv3
                                   trying to get the length of an uninitialized array, which is sufficient
                             */
                             int eras = Eras.Length;
-                            RaceEra = new RACEERA[count, eras];
-                            int raceeraRowLength = eras * sizeof(RACEERA);
+                            RaceEra = new RACE_ERAS[count, eras];
+                            int raceeraRowLength = eras * sizeof(RACE_ERAS);
 
                             fixed (void* ptr = Race, ptr2 = RaceEra) {
                                 byte* racePtr = (byte*)ptr;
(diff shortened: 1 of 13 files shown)


JimOfLeisure:
Looks good to me offhand, but I'm a bit behind. I see Quintillus is up to date and had comments on the last PR, so adding him as a reviewer.

QuintillusCFC:
Starting to review.  Debating whether to use the Review process (which hides all comments until I finish), or just leave comments so they're visible sooner.

QueryCiv3/Biq.cs
```diff
@@ -0,0 +1,549 @@
...
+namespace QueryCiv3
+{
+    public class BicData
+    {
+        public Civ3File Bic;
+        public bool HasCustomRules => Bic.SectionExists("BLDG");
+        public bool HasCustomMap => Bic.SectionExists("WCHR");
+
+        // RULE, WCHR, WMAP, and GAME all seem to only ever have a maximum of 1 header in the biq files, but I'm not certain on that
+        // If that's confirmed to be the case, they can be demoted from arrays to singular variables and the code simplified accordingly
```
QuintillusCFC, line 14: That is the case.  Apolyton indicates it, and I've never seen nor heard of anyone being able to add a second of any of these.  Not 100% sure on whether Civ3 would ignore it or crash.  If you add a sixth world map size, Civ3 crashes.  But either way it should be safe to treat it as always being one.

QueryCiv3/Biq.cs
```diff
@@ -0,0 +1,549 @@
...
+        public RACELEADERNAME[][] RaceScientificLeaderName; // the scientific leaders for each civ
+        public int[][] CityBuilding; // Building IDs in each city
+        public int[][] WmapResource;
+        public int[][] PrtoPrto; // Stealth unit targets per unit
+        public LEADPRTO[][] LeadPrto; // starting unit data for each leader
+        public int[][] LeadTech; // starting tech data for each leader
+        public RULECULT[][] RuleCult; // culture level names per rule
+        public int[][] RuleSpaceship; // spaceship quantity requirements per rule
+        public int[][] GameCiv; // Playable civs for game
+        public int[][] GameAlliance; // Civ alliances for game
```
QuintillusCFC, line 62: What do you think of having CamelCase or SCREAMING_SNAKE_CASE class names for the sub-section classes?  E.g. RaceLeaderName or RACE_LEADER_NAME instead of RACELEADERNAME?  When GOVTGOVT was the only one it seemed okay but it seems like it would be more readable with casing, and we only really use ALLCAPS because it matches Civ3.

I'm not entirely sure about this since e.g. RULECULT is kinda the same as GOVTGOVT in concept, but on the whole am leaning towards it now.

(Now that I'm writing that, I'm also thinking, does it make sense to call it BLDG instead of say Building?  But I'm leaning towards keeping BLDG, etc. since that makes it clear it's legacy Civ3, and the abbreviations are known in the Civ3 community)

I like that there are descriptions for nearly all of these, and that the rectangular/jagged distinction between some of them is called out.

QueryCiv3/Biq.cs
```diff
@@ -0,0 +1,549 @@
...
+
+                                for (int i = 0; i < count; i++) {
+                                    Buffer.MemoryCopy(dataPtr, cityPtr, CITY_LEN_1, CITY_LEN_1);
+                                    CityBuilding[i] = new int[City[i].NumberOfBuildings];
+                                    buildingRowLength = City[i].NumberOfBuildings * sizeof(int);
+                                    fixed (void* ptr2 = CityBuilding[i]) Buffer.MemoryCopy(dataPtr + CITY_LEN_1, ptr2, buildingRowLength, buildingRowLength);
+                                    Buffer.MemoryCopy(dataPtr + CITY_LEN_1 + buildingRowLength, cityPtr + CITY_LEN_1, CITY_LEN_2, CITY_LEN_2);
+
+                                    cityPtr += sizeof(CITY);
+                                    dataPtr += City[i].Length + 4;
```
QuintillusCFC, line 153: All right, I've looked at this long enough without figuring out the answer to ask... why the +4?  CITY has everything in the BIQ city object, except for the buildings.  Shouldn't we be incrementing by sizeof(CITY) + buildingRowLength?  Is City[i].Length not equal to sizeof(CITY)?  I think those are equal, since the buildings are stored separately in CityBuilding

(same applies to the following line)

This applies to a few other sections too (LEAD, PRTO).  It's probably the same thing in each case... I'm looking for a pointer whose length is 4 and needs to be accounted for, but am not finding it, and we don't have it for GOVT, for example.

QueryCiv3/Biq.cs
```diff
@@ -0,0 +1,549 @@
...
+                            dataLength = count * sizeof(EXPR);
+                            Expr = new EXPR[count];
+                            fixed (void* ptr = Expr) {
+                                Buffer.MemoryCopy(bytePtr + offset, ptr, dataLength, dataLength);
+                            }
+                            break;
+                        case "FLAV":
+                            // FLAV has two oddities compared with other sections:
+                            // 1. FLAV's count is not technically locked at 7, but is practically so. This means it can be treated as static (see FLAV.cs)
+                            count = 7;
```
QuintillusCFC, line 217: Unfortunately, some forum member named Quintillus made it practical to add additional flavors using a custom editor.

Fortunately, I haven't been able to find any evidence of anyone adding more than 7 flavors to a scenario.  If they did, they didn't publicize it in the main C&C forum.

It should be perfectly fine to have it at 7 for now, just wanted to note that since there might be one scenario lurking out there with 8 or more flavors.

QueryCiv3/BiqSections/Prto.cs
```diff
@@ -3,9 +3,174 @@
...
+        public bool AIAirTransport          { get => Util.GetFlag(Flags1[5], 1); }
+        public bool AINavalTransport        { get => Util.GetFlag(Flags1[5], 2); }
+        public bool AINavalCarrier          { get => Util.GetFlag(Flags1[5], 3); }
+        public bool AITerraform             { get => Util.GetFlag(Flags1[5], 4); }
+        public bool AISettle                { get => Util.GetFlag(Flags1[5], 5); }
+        public bool AILeader                { get => Util.GetFlag(Flags1[5], 6); }
+        public bool AITacticalNuke          { get => Util.GetFlag(Flags1[5], 7); }
+
+        public bool AIICBM                  { get => Util.GetFlag(Flags1[6], 0); }
+        public bool AINavalMissileTransport { get => Util.GetFlag(Flags1[6], 1); }
```
QuintillusCFC, line 107: Two more on the subsequent bits: AIFlag and AIKing (both strategies).

QueryCiv3/BiqSections/Prto.cs
```diff
@@ -3,9 +3,174 @@
...
+        public bool Army                    { get => Util.GetFlag(Flags1[2], 2); }
+        public bool Leader                  { get => Util.GetFlag(Flags1[2], 3); }
+        public bool ICBM                    { get => Util.GetFlag(Flags1[2], 4); }
+        public bool Stealth                 { get => Util.GetFlag(Flags1[2], 5); }
+        public bool CanSeeSubmarines        { get => Util.GetFlag(Flags1[2], 6); }
+        public bool TacticalMissile         { get => Util.GetFlag(Flags1[2], 7); }
+
+        public bool CanCarryTacticalMissiles{ get => Util.GetFlag(Flags1[3], 0); }
+        public bool RangedAttackAnimations  { get => Util.GetFlag(Flags1[3], 1); }
+        public bool TurnToAttack            { get => Util.GetFlag(Flags1[3], 2); }
```
QuintillusCFC, line 82: A few more on the subsequent bits:

LethalLandBombardment, LethalSeaBombardment, King, RequiresEscort

These are all Conquests ones IIRC.

QueryCiv3/BiqSections/Prto.cs
```diff
@@ -3,9 +3,174 @@
...
+        public bool Unload                  { get => Util.GetFlag(Flags3[4], 1); }
+        public bool Airlift                 { get => Util.GetFlag(Flags3[4], 2); }
+        public bool Pillage                 { get => Util.GetFlag(Flags3[4], 3); }
+        public bool Bombard                 { get => Util.GetFlag(Flags3[4], 4); }
+        public bool Airdrop                 { get => Util.GetFlag(Flags3[4], 5); }
+        public bool BuildArmy               { get => Util.GetFlag(Flags3[4], 6); }
+        public bool FinishImprovements      { get => Util.GetFlag(Flags3[4], 7); }
+
+        public bool UpgradeUnit             { get => Util.GetFlag(Flags3[5], 0); }
+        public bool Capture                 { get => Util.GetFlag(Flags3[5], 1); }
```
QuintillusCFC, line 136: Four empty bytes, and then:

Telepad, Teleportable, Stealth Attack, Charm, Enslave, Collateral Damage, Sacrifice, Science Age

QueryCiv3/BiqSections/Prto.cs
```diff
@@ -3,9 +3,174 @@
...
+        public bool ClearForest             { get => Util.GetFlag(Flags3[8], 7); }
+
+        public bool ClearJungle             { get => Util.GetFlag(Flags3[9], 0); }
+        public bool PlantForest             { get => Util.GetFlag(Flags3[9], 1); }
+        public bool ClearPollution          { get => Util.GetFlag(Flags3[9], 2); }
+        public bool Automate                { get => Util.GetFlag(Flags3[9], 3); }
+        public bool JoinCity                { get => Util.GetFlag(Flags3[9], 4); }
+        public bool BuildAirfield           { get => Util.GetFlag(Flags3[9], 5); }
+        public bool BuildRadarTower         { get => Util.GetFlag(Flags3[9], 6); }
+        public bool BuildOutpost            { get => Util.GetFlag(Flags3[9], 7); }
```
QuintillusCFC, line 154: One more bit: BuildBarricade

QueryCiv3/BiqSections/Race.cs
```diff
@@ -3,9 +3,136 @@
...
+        public int FreeTech3;
+        public int FreeTech4;
+
+        private fixed byte Flags[16];
+        public bool Militaristic                { get => Util.GetFlag(Flags[0], 0); }
+        public bool Commercial                  { get => Util.GetFlag(Flags[0], 1); }
+        public bool Expansionist                { get => Util.GetFlag(Flags[0], 2); }
+        public bool Scientific                  { get => Util.GetFlag(Flags[0], 3); }
+        public bool Religious                   { get => Util.GetFlag(Flags[0], 4); }
+        public bool Industrious                 { get => Util.GetFlag(Flags[0], 5); }
```
QuintillusCFC, line 83: Two more:

Agricultural (6)
Seafaring (7)

QuintillusCFC (commented):
A few more fields to add from Conquests.  At least one question.

QueryCiv3/Biq.cs
lucienmaloney: Okay, awesome. Once I've finished the Sav structs, I'll do a pass back over Biq to simplify this kind of stuff and clean it up with some of the better techniques I'm figuring out.

QueryCiv3/Biq.cs
lucienmaloney: Yeah, having it be ALLUPPERCASE was just a convention I was copying over from the Civ 3 headers in a way that didn't really make sense. But now I've reworked it in a way I think is more clear. I've added underscores, and now only have full uppercase in places where lengths match.

eg. RULECULT becomes RULE_CULT because the outer index of RULE_CULT has the same length as RULE and the inner index has the same length as CULT. However, RACECITY becomes RACE_City because while the outer index matches the length of RACE, the inner index in this context does not match that of CITY. I think this is a pretty good and slightly more clear setup, but let me know what you think.

lucienmaloney: Oh, and I didn't change the pointer names to match the new names of the structs, but I'll do that in the future Biq cleanup PR.

QueryCiv3/BiqSections/Prto.cs
lucienmaloney: Do you mean 4 empty bits instead of 4 empty bytes? So telepad would be GetFlag(Flags3[5], 6)? It might be better if you add the flags in here just because I'm not sure I'd do it right.

QueryCiv3/Biq.cs
lucienmaloney: Got it. For now this should be fine then, but I'll open up an issue and let this be a low priority improvement to be made in the future.

QueryCiv3/Biq.cs
lucienmaloney: The Length attribute as defined in Biq files records the length of everything until the end of the section _except_ for itself. That is, it is one 32-bit integer short of being the actual full length, which is why adding an extra 4 is needed. But you're right that it is acting as a magic number here and should have better documented behavior. I'll make those improvements in the cleanup PR.

QueryCiv3/Biq.cs
QuintillusCFC: That looks much better!  Easy to read at a glance now.  The variable names are probably okay as-is, and in some cases (military/scientific leaders) there are more variables than structs anyway.

QueryCiv3/Biq.cs
QuintillusCFC: Oh, right, the .Length is reading the property in the CITY object, which is from the BIQ, which doesn't include its own length in the total.  That's definitely an area where I've made off-by-four errors a few times over the years in my editor.

I'm fine with magic-reduction being in a cleanup PR.  For that purpose though, I still think sizeof(CITY) is _probably_ the same as City[i].Length + 4, right?  If so, that seems like a simpler/more readable route.

QueryCiv3/BiqSections/Prto.cs
QuintillusCFC: Oops, yeah, I talk about bytes more frequently than bits and wrote the wrong one.  So GetFlag(Flags3[5], 6) would be correct for Telepad.  Seeing if I can add them now...

QuintillusCFC: Well, I followed the guide WildWeazel linked to at https://stackoverflow.com/questions/27567846/how-can-i-check-out-a-github-pull-request-with-git to check out your code, which worked.  But when I pushed, it pushed to a new branch on _this_ repo at https://github.com/C7-Game/Prototype/commits/BiqStructs .  Not altogether surprised it didn't push to your repo, somewhat surprised it pushed to a new branch on this repo.

Going to mark this as resolved, and it should be feasible to either cherry pick my commit there for the cleanup branch, or just open a PR for that branch and the one-commit one-file change.

QuintillusCFC (approved):

PR #60, opened 2021-12-17
Flood the swamps.  They now show up on the map as marshes.

QuintillusCFC:
Creating as a draft, as a reminder to clean up duplication between this one and forest/jungle.  It is fully functional for marsh displaying, so if you're hankering for marshes, you can take it for a spin.

### DIFF
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -324,6 +324,9 @@ public class ForestLayer : LooseLayer {
 
 public class MarshLayer : LooseLayer {
 	public static readonly Vector2 marshSize = new Vector2(128, 88);
+	//Because the marsh graphics are 88 pixels tall instead of the 64 of a tile, we also need an addition 12 pixel offset to the top
+	//88 - 64 = 24; 24/2 = 12.  This keeps the marsh centered with half the extra 24 pixels above the tile and half below.
+	readonly Vector2 MARSH_OFFSET = (float)0.5 * marshSize + new Vector2(0, -12);
 	
 	private ImageTexture largeMarshTexture;
 	private ImageTexture smallMarshTexture;
@@ -347,7 +350,7 @@ public class MarshLayer : LooseLayer {
 				marshTexture = largeMarshTexture;
 			}
 			Rect2 jungleRectangle = new Rect2(randomMarshColumn * marshSize.x, randomJungleRow * marshSize.y, marshSize);
-			Rect2 screenTarget = new Rect2(tileCenter - (float)0.5 * marshSize + new Vector2(0, -12), marshSize);
+			Rect2 screenTarget = new Rect2(tileCenter - MARSH_OFFSET, marshSize);
 			looseView.DrawTextureRectRegion(marshTexture, screenTarget, jungleRectangle);			
 		}
 	}
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -252,20 +252,6 @@ public class ForestLayer : LooseLayer {
 	}
 	
 	public override void drawObject(LooseView looseView, Tile tile, Vector2 tileCenter) {
-		TerrainType northeastType = tile.neighbors[TileDirection.NORTHEAST].baseTerrainType;
-		TerrainType northwestType = tile.neighbors[TileDirection.NORTHWEST].baseTerrainType;
-		TerrainType southeastType = tile.neighbors[TileDirection.SOUTHEAST].baseTerrainType;
-		TerrainType southwestType = tile.neighbors[TileDirection.SOUTHWEST].baseTerrainType;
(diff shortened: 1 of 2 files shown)


C7/MapView.cs
```diff
@@ -336,6 +322,37 @@ public class ForestLayer : LooseLayer {
...
+			if (tile.neighborsCoast()) {
+				randomMarshColumn = tile.xCoordinate % 5;
+				marshTexture = smallMarshTexture;
+			}
+			else {
+				randomMarshColumn = tile.xCoordinate % 4;
+				marshTexture = largeMarshTexture;
+			}
+			Rect2 jungleRectangle = new Rect2(randomMarshColumn * marshSize.x, randomJungleRow * marshSize.y, marshSize);
+			Rect2 screenTarget = new Rect2(tileCenter - (float)0.5 * marshSize + new Vector2(0, -12), marshSize);
```
lucienmaloney: Shifting the target origin by half the marsh size makes sense to me, but why does there need to an extra shift 12 pixels up? I assume there's a connection with the fact that the marsh height is 88 pixels, 24 pixels more than the usual size of 64, and 24 is double 12. So, ultimately, the delta-y is 32 pixels, same as a normal-sized texture. A comment or named variable might help make this more clear.

Honestly, this is less of me having an issue with the code than it is me being consistently annoyed with how inconsistent civ 3 terrain textures work. I think the whole system needs to reworked and consolidated into a new type of file structure.

C7GameData/Tile.cs
```diff
@@ -50,6 +50,22 @@ public MapUnit findTopDefender()
         
         public static Tile NONE = new Tile();
 
+        public bool neighborsCoast() {
```
lucienmaloney: Minor nitpick, but the indentation here is a bit wonky. At some point, someone will probably have to bite the bullet an go through and standardize all the many different styles used throughout the different files, but that's less important right now than actually writing code.

C7GameData/Tile.cs
```diff
@@ -50,6 +50,22 @@ public MapUnit findTopDefender()
...
+            TerrainType northeastType = neighbors[TileDirection.NORTHEAST].baseTerrainType;
+		    TerrainType northwestType = neighbors[TileDirection.NORTHWEST].baseTerrainType;
+		    TerrainType southeastType = neighbors[TileDirection.SOUTHEAST].baseTerrainType;
+    	    TerrainType southwestType = neighbors[TileDirection.SOUTHWEST].baseTerrainType;
+
+		    TerrainType[] neighborTerrains = { northeastType, northwestType, southeastType, southwestType };
```
lucienmaloney: Perhaps having a way to easily pull out just the diagonal members of neighbors as some sort of getter would make this cleaner. I'm not exactly sure what the best approach is to that, but having to manually get each of the 4 types and then define an array to iterate through them is awkward.

lucienmaloney (approved):
I've added some minor comments on style, but the functionality is fine and working, which is the important thing.

C7GameData/Tile.cs
QuintillusCFC: Gee, that is wonky.  A mix of spaces and tabs in the same line even!  Fired it up in Notepad++ so I could see them differently, and converted the whole file to tabs.

I think part of the reason we have this problem is that VSCode has a setting "Editor: Detect Indentation" that is enabled by default.  Puppeteer and I had agreed to set tabs, and figured out how to do it, we thought, but I at least didn't know that I was only setting it for one file, and it was auto-detecting based on other files.

I've un-checked that setting now, so going forward my code should be tab-indented.

C7GameData/Tile.cs
QuintillusCFC: Added one.  That'll be nice to have if it turns out there is another diagonal case, which there probably will be.  It reads a bit nicer in the current _neighborsCoast_ method as well.

C7/MapView.cs
QuintillusCFC: Yup, that's what's going on.  I moved the offset into a constant and added a comment above it to make it clearer.  It _is_ confusing.  I have a bunch of these magic offsets in my editor code as well.

QuintillusCFC:
Noticed that the "isHilly" method was blowing up for hills next to the edge of the map.  Traced it to a non-null default name for TerrainType.  Couldn't figure out why it didn't blow up on Development.  Eventually realized that the default was not null on Development.  Fired up git blame, realized I fixed that 24 minutes before the first commit to this branch.

Yeah, branches hanging out for a month is annoying.  Things like that happen as a result.  If the turnaround time were a few days, that investigation would have never had to happen.

Merging now to avoid future wild goose chases, and since there's already approval.

PR #61, opened 2021-12-17
Non-Square Maps Now Openable

QuintillusCFC:
1. Fix switcheroo in ImportCiv3 that resulted in non-square maps not being openable.
2. Somewhat improved logging in GameData if somehow invalid tile coordinates are passed in.
3. Non-null default for terrain name, which prevents Tile.NONE from causing a blow-up in the case that you ask whether it's hilly (could happen with hills on edge of map, will be obsolete with the key refactor).
4. Increase Godot printout support, as while debugging I hit the old limit.

### DIFF
--- a/C7/project.godot
+++ b/C7/project.godot
@@ -27,13 +27,18 @@ GlobalSingleton="*res://GlobalSingleton.cs"
 
 window/size/height=768
 
+[global]
+
+limit=false
+limits=false
+
 [gui]
 
 theme/custom="res://C7Theme.tres"
 
 [network]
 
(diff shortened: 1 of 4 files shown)


QuintillusCFC:
GitHub doesn't really show it well, but the new commit is from rebasing.  I think it looks weird (two commits on Conversation, one on Commits) because I rebased from local Development after pulling that branch and switching back, rather than from origin/Development.  Then I had to force-push for GitHub to accept the update.

lucienmaloney (approved):
I tested out loading the WWII scenario, and it works now. This is ready to merge.

PR #63, opened 2021-12-19
Unit animation

### LINKED ISSUE #35: Set civ colors on units
Arguably Dutch, but we might want this for Carthage since being able to tell who's units are whose will be useful for prototype combat.

### LINKED ISSUE #36: Render unit animations

maxpetul:
Implements unit animations and draws them with a custom shader that works off of flic color palettes (instead of requiring the textures to be de-palettized when loaded). Also tints units by civ color and draws the original game's animated cursor under the selected unit. There's still some work to be done regarding how animations will be integrated with the rest of the code. For now animations are triggered by calling startAnimation in AnimationTracker and the calling code can wait for the animation to complete by passing a callback function. I don't like this as a long term solution, for more details, see my post here: https://forums.civfanatics.com/threads/shaders-color-palettes-and-unit-animations.674376/#post-16199160.


C7/Game.cs
```diff
@@ -436,8 +448,15 @@ private void UnitButtonPressed(string buttonName)
...
-			popupOverlay.ShowPopup("buildCity", PopupOverlay.PopupCategory.Advisor);
+			EngineStorage.animTracker.startAnimation(
+				OS.GetTicksMsec(),
+				CurrentlySelectedUnit.guid,
+				MapUnit.AnimatedAction.BUILD,
+				(unitGUID, action) => {
+					PopupOverlay popupOverlay = GetNode<PopupOverlay>("CanvasLayer/PopupOverlay");
+					popupOverlay.ShowPopup("buildCity", PopupOverlay.PopupCategory.Advisor);
+					return false;
+				});
```
QuintillusCFC, line 465: Don't know if this is going to cause an issue (if so, it can be delayed till later), but in Civ we get the popup first before the animation.  Currently, if you cancel out, the Settler remains at the end of the animation.

C7/Game.cs
```diff
@@ -346,23 +348,33 @@ public override void _UnhandledInput(InputEvent @event)
...
+					switch (eventKey.Scancode - (int)Godot.KeyList.Kp1) {
+					case 0: dir = TileDirection.SOUTHWEST; break;
+					case 1: dir = TileDirection.SOUTH;     break;
+					case 2: dir = TileDirection.SOUTHEAST; break;
+					case 3: dir = TileDirection.WEST;      break;
+					case 4: return; // Key pad 5 => don't move
+					case 5: dir = TileDirection.EAST;      break;
+					case 6: dir = TileDirection.NORTHWEST; break;
+					case 7: dir = TileDirection.NORTH;     break;
+					case 8: dir = TileDirection.NORTHEAST; break;
```
QuintillusCFC: Can we adjust it by Kp0 or add 1, or something, so that "case 4" actually means pressing 4?  It's odd that the 5 key is case 4 when we're dealing with numbers, and there's been some effort to make them line up, just not exactly.

C7/Game.cs
```diff
@@ -420,8 +432,8 @@ private void UnitButtonPressed(string buttonName)
 		}
 		else if (buttonName.Equals("fortify"))
 		{
-			UnitInteractions.fortifyUnit(CurrentlySelectedUnit.guid);
-			GetNextAutoselectedUnit();
+			UnitInteractions.fortifyUnit(CurrentlySelectedUnit.guid, OS.GetTicksMsec());
+			GetNextAutoselectedUnit(); // This skips the animation but if we don't do this the unit will stay selected
```
QuintillusCFC, line 443: I know it may well be temporary, but the animation could be not-skipped with the EngineStorage.animTracker.startAnimation method, right?  (It's also not that important right now, with all the units so close to each other)

C7/Util.cs
```diff
@@ -107,11 +107,83 @@ static public Pcx LoadPCX(string relPath)
...
+	// Loads a Flic and also converts it into a sprite sheet
+	public static (FlicSheet, Flic) loadFlicSheet(string filePath)
+	{
+		var flic = new Flic(Util.Civ3MediaPath(filePath));
+
+		var texPalette = Util.createPaletteTexture(flic.Palette);
+
+		var countColumns = flic.Images.GetLength(1); // Each column contains one frame
+		var countRows = flic.Images.GetLength(0); // Each row contains one animation
+		var countImages = countColumns * countRows;
```
QuintillusCFC, line 163: Potentially controversial opinion: Can we specify the types?  For some it's obvious (the last three ints, the Flic), but sometimes it isn't.  What type is texPalette?  If it were specified here, I might not have to check, even in the case where I read loadFlicSheet before createPaletteTexture (which happens to be the case in this out-of-order review).

IMO specifying the types improves readability, and thus maintainability over the long term, at a very meagre cost in the short term.

(This is also more of a going-forward thing than a change-all-of-them-that-already-exist thing)

C7/Util.cs
```diff
@@ -107,11 +107,83 @@ static public Pcx LoadPCX(string relPath)
...
+						    pixelIndex = pixelRow * countColumns * flic.Width + pixelCol;
+						allIndices[pixelIndex] = flic.Images[row, col][y * flic.Width + x];
+					}
+
+		var imgIndices = new Image();
+		imgIndices.CreateFromData(countColumns * flic.Width, countRows * flic.Height, false, Image.Format.R8, allIndices);
+		var texIndices = new ImageTexture();
+		texIndices.CreateFromImage(imgIndices, 0);
+
+		return (new FlicSheet { palette = texPalette, indices = texIndices, spriteWidth = flic.Width, spriteHeight = flic.Height }, flic);
```
QuintillusCFC, line 183: Do I interpret this method correctly that it's creating one giant texture with all the frames of the Flic in it, and the FlicSheet takes as parameters two textures, one consisting of that giant texture, and the other a palette?

This might be the definition of "sprite sheet", but as someone not acquainted with animation terminology, it was a bit surprising to see that's what all the math in lines 165-181 was doing.

C7/Util.cs
```diff
@@ -107,11 +107,83 @@ static public Pcx LoadPCX(string relPath)
...
+		var img = new Image();
+		img.CreateFromData(16, 16, false, Image.Format.Rgb8, flatPalette);
+		var tex = new ImageTexture();
+		tex.CreateFromImage(img, 0);
+		return tex;
+	}
+
+	// Creates textures from a PCX file without de-palettizing it. Returns two ImageTextures, the first is 16x16 with RGB8 format containing the
+	// color palette and the second is the size of the image itself and contains the indices in R8 format.
+	public static (ImageTexture palette, ImageTexture indices) loadPalettizedPCX(string filePath)
```
QuintillusCFC, line 136: Another thing that falls kind of in the "coding standards" camp - do we want to have methods that return two things, in general?  I'm trying to figure out what this method does, and it seems to have two responsibilities, and it's not clear why it needs to return two things.  It doesn't help that the one call to this function (in MapView.cs) seems to ignore the second return value.

Part of that also might be that I don't have a background in animation (and am thus also pondering why we're using R8 format, which according to the Godot documentation only uses the red channel for storage).  But while I'll occasionally reach for a tuple in a language that doesn't make returning two parameters easy, in general I find it's easier to follow the single-responsibility principle and read my own code later with only one return parameter.

C7Engine/EntryPoints/AnimationTracker.cs
```diff
@@ -0,0 +1,99 @@
...
+		public MapUnit.AnimatedAction action;
+		public OnAnimationCompleted callback;
+	}
+
+	private Dictionary<string, ActiveAnimation> activeAnims    = new Dictionary<string, ActiveAnimation>();
+	private Dictionary<string, ActiveAnimation> completedAnims = new Dictionary<string, ActiveAnimation>();
+
+	public void startAnimation(ulong currentTimeMS, string unitGUID, MapUnit.AnimatedAction action, OnAnimationCompleted callback)
+	{
+		ulong animDurationMS = 500; // Hard-code durations to 0.5 sec for now. Ultimately we'll want to figure this out based on the INI file.
```
QuintillusCFC: Yeah, I was going to note that the Build City animation is way too quick.  Fine with saving that for the future though.

C7Engine/EntryPoints/AnimationTracker.cs
```diff
@@ -0,0 +1,99 @@
...
+		activeAnims[unitGUID] = aa;
+		completedAnims.Remove(unitGUID);
+	}
+
+	public void endAnimation(string unitGUID, bool triggerCallback = true)
+	{
+		ActiveAnimation aa;
+		if (triggerCallback && activeAnims.TryGetValue(unitGUID, out aa)) {
+			var forget = aa.callback(unitGUID, aa.action);
+			if (! forget)
```
QuintillusCFC: I'm confused by why this is called `forget`.  We call the OnAnimationCompleted callback, and it returns a boolean.  If that boolean is false, we put the ActiveAnimation in completedAnims; if it's true, we don't.  But why exactly?  The variable name doesn't make that clear.

C7Engine/EntryPoints/AnimationTracker.cs
```diff
@@ -0,0 +1,99 @@
...
+	public (MapUnit.AnimatedAction, double) getCurrentActionAndPeriod(string unitGUID, ulong currentTimeMS)
+	{
+		ActiveAnimation aa;
+		if (! activeAnims.TryGetValue(unitGUID, out aa))
+			aa = completedAnims[unitGUID];
+
+		var durationMS = (double)(aa.endTimeMS - aa.startTimeMS);
+		if (durationMS <= 0.0)
+			durationMS = 1.0;
+		var period = (double)(currentTimeMS - aa.startTimeMS) / durationMS;
```
QuintillusCFC: If I'm reading the math right, isn't the `period` variable actually the percentage of the way through the animation that the code would be?  E.g. if startTimeMS is 100, and endTimeMS is 400, and currentTimeMS is 200, durationMS would be 300, and period would be 100/300 = 1/3.

If so, I don't think `period` is the right name for it, as that suggests a recurring interval.  The `progress` variable used at a higher level than this in UnitInteractions.cs seems like a more accurate name.

QuintillusCFC (changes requested):
Reviewed Game.cs, Util.cs, EngineStorage.cs, and AnimationTracker.cs.  Still have the others yet to do.  Which means only about 25% of it is reviewed :(.  Well, slightly more, I looked at parts of MapView before realizing I needed to go deep first.

Submitting comments for what I have looked at so far.

C7/Game.cs
maxpetul: Sure, it would make more sense if the cases lined up with the numbers on the keys. The only reason they don't is that switch statement was created from an array so it inherited starting from zero.

C7Engine/EntryPoints/AnimationTracker.cs
maxpetul: "forget" stores whether or not the animation is to be forgotten (i.e. discarded) after it's completed, or if it's kept around in completedAnims. I wrote all of this code about callbacks & completed animations sloppily since I'm intended to tear it all out and rearchitect things, and hopefully soon.

C7/Game.cs
maxpetul: Yes. I wrote that before I added the completion callback to AnimationTracker.

C7/Util.cs
maxpetul: In my opinion it makes sense in this case to have two return values since they're two parts loaded from the same file. It might help to rename this function so it's clear that's what it's doing.

About the R8 format, it's for the texture that stores palette indices so we only need one byte per pixel there. Actually it's not ideal but it's the best format Godot offers. What would be ideal would be a format that corresponds to OpenGL's GL_R8 so we could get the palette index values as integers 0 - 255 directly in the shader, but Godot's R8 format corresponds to GL_RED which can only be read as floating point values.

C7/Game.cs
maxpetul: Unfortunately I programmed all that while misremembering how the original game works. In fact the only reason I gave AnimationTracker the ability to keep completed animations around is so that the settler can be kept on the last frame of the build animation while the name input popup is open. I'm aware of the bug where the settler gets stuck on that frame if the player cancels out of the popup. The fix would be to call AnimationTracker.endAnimation when the popup is cancelled. I looked into doing that but it would require adding another signal from the popup to the main Game object and I don't know how to do that without looking it up so I didn't bother.

C7Engine/EntryPoints/AnimationTracker.cs
maxpetul: I was using "progress" to store only values 0.0 to 1.0 so I wanted a different name for that variable. "period" isn't ideal, but it sort of makes sense if you think of it as the number of periods the animation has gone through, but it's a floating point so it stores fractional periods as well. I'll rename it to "repetitionCount" for clarity.

C7/Util.cs
maxpetul: Yes, that's correct. I'll think about adding a comment to better explain what this block is doing, care to suggest one? It's probably easier as an "outsider" to know what needs to be explained, for example I never would have thought of defining the word sprite sheet.

QuintillusCFC (approved):
Approving in conjunction with the threading changes.  We've been discussing in on CFC and that branch (which builds on this one) addressed the high-level concerns; I'd rather focus on improving any low-level concerns incrementally in the future than keep animation in its own branches.

PR #64, opened 2021-12-21
If C7 fails to load a file, display an error message with potentially…

QuintillusCFC:
… helpful information and let the user return to the main menu, rather than crashing.

You can replicate this by trying to open a BIQ file.  Or a Word file or any other one, although in cases other than SAV/BIQ, when it tries to use the JSON loader, we seem to get a more generic "Failure has occurred while loading a type" message.  Still better than crashing though.


QuintillusCFC (commented):
This needs to be refactored to use the newfangled popup style before being merged.

PR #66, opened 2021-12-22
This fixes the "New Game" functionality not working, due to C7 JSON not loading.

QuintillusCFC:
Git Bisect pointed me to the VS2019 commit, and after double-checking, somehow that is the relevant one.  This commit consists of a reversion of that commit, plus including a regenerated c7-static-map-save.json file.

Please verify locally that this fixes the issue for you.  Still investigating why the VS change caused the issue, my guess is the change from net472 to netstandard2.0 in C7.csproj, though I don't know why that would have caused the issue.  But maybe Godot uses the sln and it was something in there?


QuintillusCFC:
Restored the changes to C7.sln as they are not causing any issues locally.  This seems to be sufficient for Visual Studio to compile the project, provide IntelliSense, etc., while being compatible with Godot.  It also makes more sense that the .csproj change was the one causing a JSON issue, given that there are some known changes in .netstandard 5.0 relative to .net4.7.2.

JimOfLeisure (approved):
This works for me in my environment. I can't speak for Visual Studio users.

Ulrich-Schmidt (approved):
Hmm, hard to say whether it fixes the .sav loading problem for me, because it doesn't compile... But I won't hold you guys up.

QuintillusCFC:
@Ulrich-Schmidt Does it not compile in Godot, or Visual Studio, or both?  I just realized Puppeteer and I were compiling the static-json-builder using completely different commands, so there are some differences going on.

With my local Visual Studio 2019, a clean and build works, when I open the C7.sln file.  Visual Studio reports that it is version 16.4.1, with Microsoft .NET Framework Version 4.8.03761.  I did notice that the Visual Studio version seems to have gone from 15 to 16, although I thought we'd both said we were on 2019, so maybe that has something to do with it.

maxpetul (approved):
This fix works for me too.

QuintillusCFC:
Merging to fix the New Game issue.  I'd still like to figure out what's going on with Visual Studio so Lanzelot is back in commission.  I wonder if it's VS2017 vs VS2019?  Or a Mono vs .NET issue.

JimOfLeisure:
As mentioned in the forum, I just installed VS 2019, and this PR works. I opened the .sln in VS, pulled this merged commit as head, and I could use the build option in VS with 11 warnings and no errors. I could build from Godot successfully and open a new game.

VS did drop a backup sln file and an "Upgrade.htm" file which seems to suggest it did something to the .sln, but git sees no changes, only the 2 new files.

PR #67, opened 2021-12-24
Github action readme update

JimOfLeisure:
for better documentation of what exactly is happening
and external resources we rely on

### DIFF
--- a/.github/workflows/readme.md
+++ b/.github/workflows/readme.md
@@ -8,7 +8,7 @@ I've been using test-action.yml to try out GitHub action features before integra
 
 ## export-c7.yml
 
-This build action uses [the goto-ci:mono-\<version> Docker image](https://hub.docker.com/r/barichello/godot-ci) which is based on [the latest mono Docker image](https://hub.docker.com/_/mono) which together contain the mono SDK, headless Godot, and Godot export templates.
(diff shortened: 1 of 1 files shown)


.github/workflows/readme.md
```diff
@@ -8,6 +8,8 @@ I've been using test-action.yml to try out GitHub action features before integra
 
 ## export-c7.yml
 
+This build action uses [the goto-ci:mono-\<version> Docker image](https://hub.docker.com/r/barichello/godot-ci) which is based on [the latest mono Docker image](https://hub.docker.com/_/mono) which together contain the mono SDK, headless Godot, and Godot export templates.
```
QuintillusCFC: This is supposed to be godot-ci:mono, not goto-ci, right?  Looks like it's godot in the export-c7.yml file.

Also is it supposed to be <version>?  That I could see being intentional since the version will presumably change over time, but it looks a bit temporary and the goto has me wondering.

JimOfLeisure: The docker image is barichello/godot-ci . Click the link and read the mono section. Docker images have versions, the default being :latest , but you can pick specific versions, and :mono-3.4 is the image version we want to export with Godot Mono 3.4 . 

And oh, I put goto instead of godot? Oops. I'll fix that in the PR.

QuintillusCFC: Oh, I see, it does have the version in the export-c7.yml file as well.

QuintillusCFC (approved):

QuintillusCFC:
This has been approved for over a week, so it is being "automatically" merged.

Issue #68, opened 2021-12-29
Display menu popup when main menu button is pressed


pcen:

QuintillusCFC:
This was fixed when #69 was merged.  Closing.

PR #69, opened 2021-12-29
Add menu popup

pcen:
Changes for https://github.com/C7-Game/Prototype/issues/68

- creates a separate class for popups
- simplifies the functionality of PopupOveraly to just render current popup
- the following features need to be added to popups in a future PR:
  - game menu needs italics for hotkey description, and underline for first letter when shortcut key is available


C7/C7Game.tscn
```diff
@@ -96,55 +96,55 @@ __meta__ = {
 
 [node name="DownButton" type="Button" parent="CanvasLayer/SlideOutBar/VBoxContainer"]
 margin_right = 79.0
-margin_bottom = 40.0
```
pcen: I think these changes were not intentional and should be excluded?

QuintillusCFC: If you didn't intend to change them, that is correct.  Sometimes the margins can be knocked off by accidentally dragging the mouse around the scene and moving something, which is my best guess for what occurred as the differences are not very large.

C7/UIElements/Popups/Popup.cs
```diff
@@ -0,0 +1,158 @@
...
+        for (; leftOffset < width - HTILE_SIZE; leftOffset += HTILE_SIZE)
+        {
+            image.BlitRect(center, new Rect2(new Vector2(0, 0), new Vector2(center.GetWidth(), center.GetHeight())), new Vector2(leftOffset, vOffset));
+        }
+
+        leftOffset = width - HTILE_SIZE;
+        image.BlitRect(right, new Rect2(new Vector2(0, 0), new Vector2(right.GetWidth(), right.GetHeight())), new Vector2(leftOffset, vOffset));
+    }
+
+    protected void AddBackground(int width, int height, int vOffset = 0)
```
QuintillusCFC, line 113: It might be worth adding a note on when to use vOffset on AddBackground versus a top margin.  I realized that the code is using margin when there isn't an advisor being displayed, and vOffset when there is, since margin would push the advisor down, and that is not desired.  But I'm not sure I would have realized that had I not written the initial popup code.

(This also accounts for why the EscapeQuitPopup's height is reduced.  Overall, it probably makes sense to have it the way you've done it, so that the height is the actual visible area rather than potentially included some "invisible" area as I'd written it, but I'm still thinking a comment describing the correct use of vOffset would make it less likely to have a mix of styles going forward)

C7/UIElements/Popups/PopupOverlay.cs
```diff
@@ -11,31 +11,25 @@ public class PopupOverlay : HBoxContainer
...
 
 	Control currentChild = null;
-	
-	const int HTILE_SIZE = 61;
-	const int VTILE_SIZE = 44;
-	private readonly static int BUTTON_LABEL_OFFSET = 4;
-	private static ImageTexture InactiveButton;
-	private static ImageTexture HoverButton;
+
+	public const string NodePath = "/root/C7Game/CanvasLayer/PopupOverlay";
```
QuintillusCFC, line 15: Nifty that it can be referred to in absolute terms this way.  Makes sense, but hadn't realized you could do that.

If we're going to use it as a constant, though, it probably makes sense to replace the three instances in Game.cs with it as well (they are right above the lines you've already modified in Game.cs)

C7/UIElements/Popups/GameMenu.cs
```diff
@@ -0,0 +1,45 @@
...
+    public override void _Ready()
+    {
+        base._Ready();
+
+        ImageTexture thisTexture = new ImageTexture();
+        Image image = new Image();
+        image.Create(370, 300, false, Image.Format.Rgb8);
+        image.Fill(Color.Color8(0, 0, 0, 0));
+        thisTexture.CreateFromImage(image);
+        this.Texture = thisTexture;
```
QuintillusCFC: Reading through this, I thought, "why are we doing this here?", and realized it's essentially boilerplate we have to do for every popup.  Can we toss it in another helper method, similar to AddBackground?  SetupDimensions or something of that sort?  I don't know that we can combine it with AddBackground since there are edge cases like needing 10 pixels to the right on right-aligned ones (although... shouldn't that be a margin thing?  Or done automatically for right alignment?), but we ought to be able to cut 6 lines down to 1, I'm pretty sure.

(Also could be punted for the future PR)

C7/UIElements/Popups/GameMenu.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        AddHeader("Main Menu", 10);
+
+        AddButton("Map", 60, "map");
+        AddButton("Load Game (Ctrl-L)", 85, "load");
+        AddButton("New Game (Ctrl-Shift-Q)", 110, "newGame");
+        AddButton("Preferences (Ctrl-P)", 135, "preferences");
+        AddButton("Retire (Ctrl-Q)", 160, "retire");
+        AddButton("Save Game (Ctrl-S)", 185, "save");
+        AddButton("Quit Game (ESC)", 210, "quit");
```
QuintillusCFC, line 26: I saw the note about italic, underline, and so forth.  I'm also wondering as we move towards Babylon, should we even note the shortcuts before they are available?  We (in some cases "I") tossed a few things in early just to add some interactivity, or the illusion thereof, but maybe we should present a more accurate picture of what's available.  At this time, I believe only ESC for Quit Game is implemented.

QuintillusCFC (commented):
Looks like a nice improvement over the initial code we had for this area, and it's cool seeing the game menu built on those improved foundations.  A few comments but basically approved, other than reverting the dimension changes in the .tscn.

C7/UIElements/Popups/Popup.cs
```diff
@@ -105,6 +101,14 @@ protected void AddBackground(int width, int height, int vOffset = 0)
 
     private TextureRect CreateBackground(int width, int height)
     {
+        TextureRect rect = new TextureRect();
+
+        if (backgroundCache.ContainsKey((width, height)))
```
QuintillusCFC, line 124: Just wondering, were you able to notice a difference with this caching?  I always like to be able to document if a performance improvement made a difference, since otherwise it's easy to chase things that don't yield improvements but might increase complexity, and occasionally I've even made what I thought would be an improvement, but actually made things worse.

I don't see any problem with having this cache, so it doesn't matter much, just something to think about.

QuintillusCFC:
Non-code-related question: Do you have a CFC username?  I know WildWeazel mentioned there were a couple potential new contributors _somewhere_ on the forum, but am having trouble finding that post, and don't recall if you are one of the ones we were expecting.

C7/UIElements/Popups/Popup.cs
pcen: will add a comment documenting this 👍

C7/UIElements/Popups/GameMenu.cs
pcen: It should be a margin for the 10 pixel padding as well, I didn't want to add too many changes at once, but it probably makes sense to do this in the same pr

C7/UIElements/Popups/GameMenu.cs
pcen: from a development perspective I think it's nice to have so it's easy to hook up buttons to code once it's ready (ie. save when we can serialize a game to .SAV format), but if we don't want them for releases I can remove too

C7/UIElements/Popups/Popup.cs
pcen: On my laptop the initial creation of popups spends ~2ms on creating the background texture, which is bypassed on subsequent renderings of popups with the same dimensions. I can't really notice a difference running it, but it does cut out those 2ms. I added the optimization because there's a comment in the code somewhere saying that it's expensive to create the background with the current method. If the 2ms isn't worth the added complexity I can remove.

pcen:
> 

I don't have a CFC account, I'm only in the discord as pub

pcen:
> Non-code-related question: Do you have a CFC username? I know WildWeazel mentioned there were a couple potential new contributors _somewhere_ on the forum, but am having trouble finding that post, and don't recall if you are one of the ones we were expecting.

I don't have a CFC account, but my username in the discord server is pub

C7/UIElements/Popups/Popup.cs
QuintillusCFC: Cool, good to hear there are some savings.  I wonder if that comment is from the very first iteration of popups?  It used to be a lot worse.  Either way, that's 1/8th of a 60 FPS target saved, so it adds up.

QuintillusCFC:
> > Non-code-related question: Do you have a CFC username? I know WildWeazel mentioned there were a couple potential new contributors _somewhere_ on the forum, but am having trouble finding that post, and don't recall if you are one of the ones we were expecting.
> 
> I don't have a CFC account, but my username in the discord server is pub

Oh, awesome!  I tend to forget to check the Discord as well as the forums, I see your posts there now.

C7/UIElements/Popups/BuildCityDialog.cs
```diff
@@ -8,7 +8,7 @@ public class BuildCityDialog : Popup
 	public BuildCityDialog()
     {
 		alignment = BoxContainer.AlignMode.End;
-		margins = new Margins();
+		margins = new Margins(right: 10); // 10px margin on the right
```
QuintillusCFC, line 11: Trying things out manually, I noticed the build city dialog wasn't 10 pixels in from the right.  Playing around with it, I figured out that it's actually required to be right: -10 to be 10 pixels left of the right edge.

This might be related, but how are the margins actually applied?  I see them declared on Popup.cs, and changing them has an outcome, so they must be... inherited from Godot somehow?  But that's a bit confusing since Margins is a class defined _within_ Popup.cs, so how could Godot be inheriting that behavior?

Not sure if that confusion also contains a hint as to why right must be -10 instead of +10, but the confusion is why I can't tell you why it needs to be -10 (and wouldn't have guessed it if I hadn't tried 50 instead of 10 and noticed it went the unexpected way).

C7/UIElements/Popups/GameMenu.cs
QuintillusCFC: That looks like a good change!

C7/UIElements/Popups/BuildCityDialog.cs
pcen: They're taken from the Margins class and applied to the Popupoveraly manually in the popup overlay code before adding the actual Popup as a child. This might not be necessary though if we can set margins on the Popup despite it not being the top level component. As for the negative margins I'm not sure since whatever we set in margins is set directly, so I'll look into that some more

QuintillusCFC: @pcen - What are your thoughts on setting the margin to -10, and opening a separate issue to investigate the why and potentially making it more intuitive?  I'd like to build upon your changes in this PR with another popup, and this doesn't seem like a big enough issue to justify keeping the entire popup refactor out of the Development branch.

pcen: Makes sense, I'll make those changes and we can get this merged

QuintillusCFC:
Resolved the conflict that arose when Flintlock's animation feature was merged in; it was pretty straightforward.  Since it's been a week, I'm going to go ahead and merge as-is and make the 10px adjustment separately.

PR #70, opened 2021-12-31
Set up the logic and structs for full Sav data slurping

lucienmaloney:
This refactoring of the Sav parsing should give us access to a lot of new data- hopefully all the data that is needed to progress in the near future. The PR title says this achieves "full" slurping, but that's full only in the sense of incorporating all that is known from available documentation. Because Savs are more complicated and less fully mapped out than Biqs, there are a good number of flags yet to be filled and "UnknownBuffer"s yet to be populated. However, fortunately, there are no gaps. The Sav slurper can successfully go from beginning to end of the file in the ones I tested without any skipping ahead or linear searching.


C7GameData/ImportCiv3.cs
```diff
@@ -36,31 +36,35 @@ static public C7SaveFormat ImportSav(string savePath, string defaultBicPath)
...
-                if (civ3Tile.isPineForest) {
+                if (civ3Tile.PineForest) {
                     c7Tile.isPineForest = true;
                 }
                 c7Save.GameData.map.tiles.Add(c7Tile);
+                i++;
             }
             // This probably doesn't belong here, but not sure where else to put it
-            c7Save.GameData.map.RelativeModPath = civ3Save.MediaBic.Game[0].ScenarioSearchFolders;
+            // c7Save.GameData.map.RelativeModPath = civ3Save.MediaBic.Game[0].ScenarioSearchFolders;
```
QuintillusCFC, line 67: Is this replaced somewhere?  It's used in Game.cs, where it's passed to Civ3Map, where it's used to load PCX files on the scenario search path.

I'm not saying it belongs here either, but it looks suspicious just being commented out.

QueryCiv3/SavSections/Radt.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace QueryCiv3.Sav
+{
+    [StructLayout(LayoutKind.Sequential, Pack=1)]
+    public unsafe struct RADT
+    {
+        // It seems that RADT, VLOC, and OUTP all have completely identical structs
```
QuintillusCFC, line 9: Yeah, that makes sense, fundamentally they're all just locations on a map with possible player control, and the difference is their type and function.

QueryCiv3/Sav.cs
```diff
@@ -1,125 +1,371 @@
...
-                CityOffset += 0x228 + (CityList[i].CitizenCount * 300) + (Bic.Bldg.Length * 8) + 0x30;
+                int* header;
+                scan = bytePtr + BIQ_SECTION_START + BiqSectionLength;
+                byte* end = bytePtr + savBytes.Length;
+
+                while (scan < end) {
+                    header = (int*)scan;
+
+                    // Every header in civ 3 files is exactly 4-chars long, which means they can be represented as 32-bit integers instead of strings
+                    // Switching off of these hex values is substantially faster than string switching, but comes at the expense of readability
```
QuintillusCFC, line 141: "is substantially faster" - How much faster is this, really?

From a conceptual standpoint, it makes sense that int comparisons are quicker than string comparisons.  And if we were iterating through every single byte checking, "does this match?" - a naive approach I once used when searching for SAV structures in memory, knowing what I was looking for but not knowing where it was - I'd expect it to be noticeable.

But here, it looks like we're making a couple dozen comparisons based on the number of sections?  If so, I'm skeptical that it would add up to much at all, relative to all the other work that has to be done.

It is well-documented with the section headers right there, and this is not likely to be a highly-modified piece of code, so it's fine to leave it as-is.  It just matched my "micro-optimization that's probably undetectable in real life" sensor.

QuintillusCFC (approved):
One comment that might be important in ImportCiv3.cs.  Otherwise ought to be fine.  Given the size of the changes, I only really focused on the logic rather than the structure of the specific SAV sections - we can fill in any gaps, including the flags that are outstanding, when we need them.

QuintillusCFC (approved):
Ran into needing this branch again when trying to dust off the rivers code.  I'm still curious about the question from 5 days ago, but am going to go ahead and approve and merge it because the lack of it is more problematic than that question.  Also resolved the conflict and tested to verify it still loads SAV maps, including from custom scenarios.

WildWeazel:
RelativeModPath does not appear to be set anywhere else in the code. @lucienmaloney was it causing a problem?

PR #71, opened 2022-01-01
Improve PCX loading and Godot conversion performance

lucienmaloney:
There were a number of small and moderate gains to be made in the speed of the code controlling PCX loading. Most of these were minor changes, like instead of calculating length as width * height in every for loop iteration, the length could be moved outside. The biggest change was replacing (relatively) expensive SetPixel function calls with a single CreateFromData call.

Further PCX performance progress can be made by moving from array indexing to pointers. Unfortunately, like in QueryCiv3, this would require unsafe code, so I've avoided making that step for now.

### DIFF
--- a/C7/PCXToGodot.cs
+++ b/C7/PCXToGodot.cs
@@ -6,12 +6,6 @@ public class PCXToGodot : Godot.Object
 {
 	private readonly static byte CIV3_TRANSPARENCY_START = 254;
 
-	private static ImageTexture getImageTextureFromImage(Image image) {
-		ImageTexture Txtr = new ImageTexture();
-		Txtr.CreateFromImage(image, 0);
-		return Txtr;
-	}
-
 	public static ImageTexture getImageTextureFromPCX(Pcx pcx) {
 		Image ImgTxtr = ByteArrayToImage(pcx.ColorIndices, pcx.Palette, pcx.Width, pcx.Height);
 		return getImageTextureFromImage(ImgTxtr);
@@ -20,6 +14,77 @@ public class PCXToGodot : Godot.Object
 	public static ImageTexture getImageTextureFromPCX(Pcx pcx, int leftStart, int topStart, int croppedWidth, int croppedHeight) {
 		Image image = getImageFromPCX(pcx, leftStart, topStart, croppedWidth, croppedHeight);
 		return getImageTextureFromImage(image);
+	}
+
+	/**
+	 * This method is for cases where we want to use components of multiple PCXs in a texture, such as for the popup background.
+	 **/
+	public static Image getImageFromPCX(Pcx pcx, int leftStart, int topStart, int croppedWidth, int croppedHeight) {
+		int[] ColorData = loadPalette(pcx.Palette);
+		int[] BufferData = new int[croppedWidth * croppedHeight];
+
+		int DataIndex = 0;
+
+		for (int y = topStart; y < topStart + croppedHeight; y++) {
+			for (int x = leftStart; x < leftStart + croppedWidth; x++) {
+				BufferData[DataIndex] = ColorData[pcx.ColorIndexAt(x, y)];
+				DataIndex++;
+			}
+		}
+
+		return getImageFromBufferData(croppedWidth, croppedHeight, BufferData);
+	}
+
+	public static ImageTexture getImageFromPCXWithAlphaBlend(Pcx imagePcx, Pcx alphaPcx) {
+		return getImageFromPCXWithAlphaBlend(imagePcx, alphaPcx, 0, 0, imagePcx.Width, imagePcx.Height);
+	}
+
+	//Combines two PCXs, one used for the alpha, to produce a final output image.
+	//Some files, such as Art/interface/menuButtons.pcx and Art/interface/menuButtonsAlpha.pcx, use this method.
+	public static ImageTexture getImageFromPCXWithAlphaBlend(Pcx imagePcx, Pcx alphaPcx, int leftStart, int topStart, int croppedWidth, int croppedHeight, int alphaRowOffset = 0) {
+		int[] ColorData = loadPalette(imagePcx.Palette);
+		int[] AlphaData = loadAlphaPalette(alphaPcx.Palette, ColorData);
+		int[] BufferData = new int[croppedWidth * croppedHeight];
+
+		int AlphaIndex;
+		int DataIndex = 0;
+
+		for (int y = topStart; y < topStart + croppedHeight; y++) {
+			AlphaIndex = (y - alphaRowOffset) * imagePcx.Width + leftStart;
+			for (int x = leftStart; x < leftStart + croppedWidth; x++) {
+				BufferData[DataIndex] = ColorData[imagePcx.ColorIndexAt(x, y)] | AlphaData[alphaPcx.ColorIndices[AlphaIndex]];
+				DataIndex++;
+				AlphaIndex++;
+			}
+		}
+
+		Image OutImage = getImageFromBufferData(croppedWidth, croppedHeight, BufferData);
+		return getImageTextureFromImage(OutImage);
+	}
+
+	private static Image ByteArrayToImage(byte[] colorIndices, byte[,] palette, int width, int height, int[] transparent = null, bool shadows = false) {
+		int[] ColorData = loadPalette(palette, shadows);
+		int[] BufferData = new int[width * height];
+
+		for (int i = 0; i < width * height; i++) {
+			BufferData[i] = ColorData[colorIndices[i]];
+		}
+
+		return getImageFromBufferData(width, height, BufferData);
+	}
+
+	private static Image getImageFromBufferData(int width, int height, int[] bufferData) {
+		Image image = new Image();
+		var Data = new byte[4 * width * height];
+		Buffer.BlockCopy(bufferData, 0, Data, 0, 4 * width * height);
+		image.CreateFromData(width, height, false, Image.Format.Rgba8, Data);
+		return image;
+	}
+
+	private static ImageTexture getImageTextureFromImage(Image image) {
+		ImageTexture Txtr = new ImageTexture();
+		Txtr.CreateFromImage(image, 0);
+		return Txtr;
 	}
 
 	private static int[] loadPalette(byte[,] palette, bool shadows = false) {
@@ -59,70 +124,5 @@ public class PCXToGodot : Godot.Object
 		}
 
 		return AlphaData;
-	}
-
-	private static Image getImageFromBufferData(int width, int height, int[] bufferData) {
-		Image image = new Image();
-		var Data = new byte[4 * width * height];
-		Buffer.BlockCopy(bufferData, 0, Data, 0, 4 * width * height);
-		image.CreateFromData(width, height, false, Image.Format.Rgba8, Data);
-		return image;
-	}
-
-	/**
-	 * This method is for cases where we want to use components of multiple PCXs in a texture, such as for the popup background.
-	 **/
-	public static Image getImageFromPCX(Pcx pcx, int leftStart, int topStart, int croppedWidth, int croppedHeight) {
-		int[] ColorData = loadPalette(pcx.Palette);
-		int[] BufferData = new int[croppedWidth * croppedHeight];
-
-		int DataIndex = 0;
-
-		for (int y = topStart; y < topStart + croppedHeight; y++) {
-			for (int x = leftStart; x < leftStart + croppedWidth; x++) {
-				BufferData[DataIndex] = ColorData[pcx.ColorIndexAt(x, y)];
-				DataIndex++;
-			}
-		}
-
-		return getImageFromBufferData(croppedWidth, croppedHeight, BufferData);
-	}
-
-	private static Image ByteArrayToImage(byte[] colorIndices, byte[,] palette, int width, int height, int[] transparent = null, bool shadows = false) {
-		int[] ColorData = loadPalette(palette, shadows);
-		int[] BufferData = new int[width * height];
-
-		for (int i = 0; i < width * height; i++) {
-			BufferData[i] = ColorData[colorIndices[i]];
-		}
-
-		return getImageFromBufferData(width, height, BufferData);
-	}
-
-	public static ImageTexture getImageFromPCXWithAlphaBlend(Pcx imagePcx, Pcx alphaPcx) {
-		return getImageFromPCXWithAlphaBlend(imagePcx, alphaPcx, 0, 0, imagePcx.Width, imagePcx.Height);
-	}
-
-	//Combines two PCXs, one used for the alpha, to produce a final output image.
-	//Some files, such as Art/interface/menuButtons.pcx and Art/interface/menuButtonsAlpha.pcx, use this method.
-	public static ImageTexture getImageFromPCXWithAlphaBlend(Pcx imagePcx, Pcx alphaPcx, int leftStart, int topStart, int croppedWidth, int croppedHeight, int alphaRowOffset = 0) {
-		int[] ColorData = loadPalette(imagePcx.Palette);
-		int[] AlphaData = loadAlphaPalette(alphaPcx.Palette, ColorData);
-		int[] BufferData = new int[croppedWidth * croppedHeight];
-
-		int AlphaIndex;
-		int DataIndex = 0;
-
-		for (int y = topStart; y < topStart + croppedHeight; y++) {
-			AlphaIndex = (y - alphaRowOffset) * imagePcx.Width + leftStart;
-			for (int x = leftStart; x < leftStart + croppedWidth; x++) {
-				BufferData[DataIndex] = ColorData[imagePcx.ColorIndexAt(x, y)] | AlphaData[alphaPcx.ColorIndices[AlphaIndex]];
-				DataIndex++;
-				AlphaIndex++;
-			}
-		}
-
-		Image OutImage = getImageFromBufferData(croppedWidth, croppedHeight, BufferData);
-		return getImageTextureFromImage(OutImage);
 	}
 }
--- a/C7/PCXToGodot.cs
+++ b/C7/PCXToGodot.cs
@@ -5,65 +5,98 @@ using ConvertCiv3Media;
 public class PCXToGodot : Godot.Object
 {
 	private readonly static byte CIV3_TRANSPARENCY_START = 254;
-	
-	public static ImageTexture getImageTextureFromPCX(Pcx pcx) {
-		Image ImgTxtr = ByteArrayToImage(pcx.ColorIndices, pcx.Palette, pcx.Width, pcx.Height);
+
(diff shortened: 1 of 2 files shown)


ConvertCiv3Media/ReadPcx.cs
```diff
@@ -43,44 +43,48 @@ public byte ColorIndexAt(int x, int y)
...
                     for (int j = 0; j < RunLen; j++) {
                         // Add pixel copy if it's not a junk byte
-                        if (!(JunkByte && LineIdx % BytesPerLine == BytesPerLine - 1)) {
-                            ListImage.Add(PcxBytes[PcxIdx]);
-                            ImgIdx++;
+                        if (JunkByte) {
+                            if (++LineIdx == BytesPerLine) {
+                                LineIdx = 0;
+                                ImgIdx--;
+                            }
```
QuintillusCFC: I'm not really sure if the new code is any clearer than the old code, which admittedly was not very clear.  So I took a look at what I wrote in Java back in 2012 to get an idea of what might be clearer.  A few thoughts:

 - JunkByte is confusing.  I initially thought it was "is this byte a junk byte?", but it's just a true/false of whether there are junk bytes.  Suggestions below, based on the other thoughts...
 - Prefix/Postfix operators (++x, x++), when combined with an operation on the same line, are IMO confusing.  I always have to check "does it add first when it's ++x or x++?"  I'm guessing it must be adding first before comparing to BytesPerLine here, but not having to think about that would simplify the relatively complex code a bit.  I'd rather have an extra line than have to think about that in other words.  (If I were primarily a low-level programmer this probably wouldn't be of concern, but IIRC only Flintlock is really a low-level programmer on the team)
 - In the new code, we're still adding the byte to ColorIndices if it's a junk byte; in the old code we are not.  I'm 99% sure the old code is correct in that regard.
 - My Java code appears to not handle junk bytes at all.  Why doesn't it fail?  Probably because all the Civ3 graphics have an even width.  This would also be why the above didn't have any visual side effects.

So let's see if I can re-write this following my suggestions, while still having it be nicer than the old version.

```
for (int j = 0; j < RunLen; j++) {
    if (!(JunkByte && (LineIdx == BytesPerLine - 1))) {
        ColorIndices[ImgIdx] = PcxBytes[PcxIdx];
        ImgIdx++;
    }
    LineIdx++;
    if (LineIdx == BytesPerLine) {
        LineIdx = 0;
    }
}
```
Hmm... that wound up being almost the same as the old code.  So maybe the suggestion should be "add a helper function to make things more clear", e.g.

```
for (int j = 0; j < RunLen; j++) {
    if (shouldAddPixel(JunkByte, LineIdx, BytesPerLine)) {
        ...

bool shouldAddPixel(bool JunkByte, int LineIdx, Int BytesPerLine) {
    return !(JunkBytes && (LineIdx == BytesPerLine - 1));
}
```
That could then be used in the literal case as well.

Also maybe DiscardLastByte instead of JunkByte?  It's bothering me that the current name implies the current byte is a junk byte when that isn't the case.

ConvertCiv3Media/ReadPcx.cs
```diff
@@ -43,44 +43,48 @@ public byte ColorIndexAt(int x, int y)
...
+                        ColorIndices[ImgIdx++] = PcxBytes[PcxIdx];
                     }
                     PcxIdx++;
                 } else {
                     // Add as literal pixel if it's not a junk byte
-                    if (!(JunkByte && LineIdx % BytesPerLine == BytesPerLine - 1)) {
-                        ListImage.Add(PcxBytes[PcxIdx]);
-                        ImgIdx++;
-                    }
+                    ColorIndices[ImgIdx++] = PcxData;
```
QuintillusCFC: See note above: This should be inside the conditional check for junk bytes.

ConvertCiv3Media/ReadPcx.cs
QuintillusCFC: Haha, I see now why my Java code doesn't worry about junk bytes.  It uses 2D indexing, and simply reads in the junk bytes into its array.  But then when it goes to convert that 2D array of indexes to an image, it only reads the x indices through xSize, rather than through BytesPerLine.

That wouldn't work with our 1D array of indices, and I suspect when I read the other file I'll see that the 1D aspect is important for performance.  But it's an interesting tradeoff of where the complexity lies.

C7/PCXToGodot.cs
```diff
@@ -5,65 +5,102 @@
 public class PCXToGodot : Godot.Object
 {
 	private readonly static byte CIV3_TRANSPARENCY_START = 254;
-	
-	public static ImageTexture getImageTextureFromPCX(Pcx pcx) {
-		Image ImgTxtr = ByteArrayToImage(pcx.ColorIndices, pcx.Palette, pcx.Width, pcx.Height);
+
+	public static ImageTexture getImageTextureFromImage(Image image) {
```
QuintillusCFC: Should this be private?  My first thought upon seeing it as the first method was, "why would we pass a generic Godot image into the PCXToGodot function?"  I then saw that this is really a helper method to de-duplicate code from the following two methods.

(If it should be private, it probably also makes sense to move it after the next two methods, so it's not as highly-visible when someone's looking for options of what to call.  I don't always put public methods first, but when I do it's because I prefer to make them easy to find)

C7/PCXToGodot.cs
```diff
@@ -5,65 +5,102 @@
...
+	// BufferData should be as large as the largest PCX file needed to be loaded in pixels
+	// For now, I will assume no PCX file is larger than 4k resolution
+	// It's efficient to use a shared BufferData, ColorData, and AlphaData instead of having to reinitialize these arrays for every Image load
+	// However, this will become a problem if (when?) we move towards parallelization
+	// But for now, it's an improvement
+	public static int[] BufferData = new int[3840 * 2160];
+	public static int[] ColorData = new int[256];
+	public static int[] AlphaData = new int[256];
+
+	public static void loadPalette(byte[,] palette, bool shadows = false) {
```
QuintillusCFC: Same as above: this should probably be private.  That makes it clear there's no need to check for external references.

QuintillusCFC: Same with loadAlphaPalette, getImageFromBufferData.

As a reviewer, this also helps make it clear what the top level is.  If it's new and it's private, I know that it's a helper method and I shouldn't start by reviewing that method until I know how it's used at a higher level.

I also generally prefer that helper methods be below the methods that call them, for the same basic reason.  Without guidance, I generally explore files top-to-bottom, with a preference for starting at the entry points (public/protected methods).  Back in the old days (C89, for instance), helper methods had to be defined first so the compiler could find them, but nowadays I prefer human readability.

C7/PCXToGodot.cs
```diff
@@ -5,65 +5,102 @@
...
+	}
+
+	// BufferData should be as large as the largest PCX file needed to be loaded in pixels
+	// For now, I will assume no PCX file is larger than 4k resolution
+	// It's efficient to use a shared BufferData, ColorData, and AlphaData instead of having to reinitialize these arrays for every Image load
+	// However, this will become a problem if (when?) we move towards parallelization
+	// But for now, it's an improvement
+	public static int[] BufferData = new int[3840 * 2160];
+	public static int[] ColorData = new int[256];
+	public static int[] AlphaData = new int[256];
```
QuintillusCFC: Is there a measurable improvement to not allocating 8.5 MB or so of data?  To me this seems like a micro-optimization that will burn someone when they unsuspectingly try to open two PCX files at once (granted, via threads, but at some point someone will likely try it).  And since most PCX files are a lot smaller than 4K, it's a lot less than 8 MP per file... pollution.pcx would only be saving a bit over 200KB of allocation.

If there's data to show that this makes a big difference, maybe... but even then there should be some locks around the entry points to avoid this being a footgun.  And as one of the goals is to be more stable and have fewer limitations than Civ3, I'd rather err on the side of safety.

(Dynamic buffer data allocation would also cover that inevitable 8K texture from 2032... it may seem far-fetched but I've already seen a full-screen texture in a Civ3 mod from China, so I wouldn't be shocked if someone with one of those lovely 5K LG or Dell monitors tries something like that)

C7/PCXToGodot.cs
```diff
@@ -5,65 +5,102 @@
...
+
+		for (int i = CIV3_TRANSPARENCY_START; i < 256; i++) {
+			ColorData[i] &= 0x00ffffff;
+		}
+
+		if (shadows) {
+			for (int i = 240; i < 256; i++) {
+				ColorData[i] = ((255 - i) * 16) << 24;
+			}
+		}
```
QuintillusCFC, line 33: This looks like a nice improvement to shadow readability to me.  @JimOfLeisure Does this make sense to you, as the author of the shadow code and most knowledgeable person on it?

C7/PCXToGodot.cs
```diff
@@ -5,65 +5,102 @@
...
+
+		int yEnd = topStart + croppedHeight;
+		int xEnd = leftStart + croppedWidth;
+		int Index;
+		int DataIndex = 0;
+
+		for (int y = topStart; y < yEnd; y++) {
+			Index = y * pcx.Width + leftStart;
+			for (int x = leftStart; x < xEnd; x++) {
+				BufferData[DataIndex++] = ColorData[pcx.ColorIndices[Index++]];
```
QuintillusCFC: Minor, but why not just move Index = to inside the x loop, and calculate it once there?

```
for (int x = leftStart; x < xEnd; x++) {
    Index = y * pcx.Width + leftStart + x;
}
```

It's probably a readability versus math overhead thing.  TBH, I'd just as soon leave one call to ColorIndexAt, which is designed for this purpose:

```
		for (int y = topStart; y < yEnd; y++) {
			for (int x = leftStart; x < xEnd; x++) {
				BufferData[DataIndex++] = ColorData[pcx.ColorIndexAt(x, y)];
```

Yes, there's slightly technically more multiplication, but can the difference be measured as noticeable?  There's still a 75% decrease with the calculations versus the old method as well.

If there _is_ a significant difference, it may make sense to leave it as-is, but I've learned that it takes an awful lot of multiplication to matter on a modern CPU, so I'm inclined to go with the more readable option without data on why it matters in that case.

C7/PCXToGodot.cs
```diff
@@ -73,30 +110,24 @@ public class PCXToGodot : Godot.Object
...
-		}
-		OutImage.Unlock();
+		loadPalette(imagePcx.Palette);
+		loadAlphaPalette(alphaPcx.Palette);
 
-		ImageTexture Txtr = new ImageTexture();
-		Txtr.CreateFromImage(OutImage, 0);
-		return Txtr;
+		int yEnd = topStart + croppedHeight;
+		int xEnd = leftStart + croppedWidth;
```
QuintillusCFC: Minor, but in cases like this where yEnd/xEnd are only used in one place and are simple, I don't think there's any gain to separating them out.  If we're reading line 122, it's quicker to read, "y goes from topStart to topStart + croppedHeight... okay, so it's a range of size croppedHeight starting at topStart", than to read, "y goes from topStart to yEnd... now what's yEnd?  Oh right, it's topStart + croppedHeight... so the difference is croppedHeight, so it's a range from topStart of length croppedHeight".  It just winds up being a different line to read to read what could be on one line easily.

Obviously there comes a point when separating things out _does_ make sense (see the junk byte comments in Pcx.cs), but if it's just an offset from the start, I think it's easier to have it in-line.  As for any performance difference, an optimizing compiler will see that topStart/croppedHeight never change, and make the upper bound of the for loop be a constant, so the performance will be identical.

C7/PCXToGodot.cs
```diff
@@ -73,30 +110,24 @@ public class PCXToGodot : Godot.Object
...
+		int Index;
+		int AlphaIndex;
+		int DataIndex = 0;
 
+		for (int y = topStart; y < yEnd; y++) {
+			Index = y * imagePcx.Width + leftStart;
+			AlphaIndex = (y - alphaRowOffset) * imagePcx.Width + leftStart;
+			for (int x = leftStart; x < xEnd; x++) {
+				BufferData[DataIndex++] = ColorData[imagePcx.ColorIndices[Index++]] | AlphaData[alphaPcx.ColorIndices[AlphaIndex++]];
+			}
```
QuintillusCFC, line 109: See the comment from getImageFromPCX about "does keeping track of Index make sense here?"  It should be consistent between the two methods regardless.

QuintillusCFC (commented):
Pretty sure there's a bug around junk bytes being added, see comment (although AFAIK Civ3 only uses even-width PCX files?  so it _probably_ won't matter, at least until we find that one odd-width file).

Otherwise primarily stylistic/readability comments.  I want to be careful about sacrificing readability for micro-optimizations without data showing the micro-optimizations made a significant difference, and I suspect we may be going too far (I also found micro-optimizations did very little for FLC, whereas avoiding SetPixel did a ton, as you mentioned here too).  The most important such comment may be one on the shared structs in PCXToGodot.cs - they seem like a hazard waiting to happen, so I'd like to see data showing that they really do add up in performance, and even if so some safety locks would be great.

ConvertCiv3Media/ReadPcx.cs
lucienmaloney: Ah, that is a fun way to circumvent the junk byte problem! But yeah, they do matter. Commenting out that part of the code causes the program to crash when trying to load a sav (though the main menu works fine), so there are at least some odd pcx files.

I admit this isn't any clearer than the old code and is probably worse in fact to read. I went a little bit crazy with the micro-optimizations here, since reverting all changes to this main PCX loop (except the switching from List to array) only slows the "Game scene load time" down by about 40-50ms. Now, this isn't a full measure of the slowdown, since PCX files are loaded and will be loaded in other places, but even then, it's not a crazy difference. Personally, my style-vs-performance tradeoff calculus still favors keeping the changes, because having measurable differences left on the table eats away at me, but I also completely get why that's not necessarily the right call for a long-term project.

lucienmaloney: As far as working correctly, the new code is actually producing the same correct results as the old. It's just doing it in a very sneaky and not-at-all-sufficiently-documented way. Like you say, when the loop goes into single-pixel mode instead of run-of-pixels mode, it sets the next ColorIndices value, regardless of whether it's a junk byte or not. But then it *does* check whether that byte was a junk value, and if so, it decrements the ImgIdx. What this means is that on the next loop iteration, the same pixel will be written to, correcting post-hoc any mistakes. The reason why the ColorIndices doesn't overflow when the end of it is reached and there is still an extra junk byte left to go is that the outer loop stops when width * height number of pixels have been read. So yeah, this is an equivalent change to the code, but perhaps a poorly-designed one.

C7/PCXToGodot.cs
lucienmaloney: This is another case of my design decisions being wonky, yep, where I set everything to public by default and only make things private when I remember or realize that it won't be accessed outside the class.

C7/PCXToGodot.cs
lucienmaloney: Ah, I tend to default to the C++ ways of writing callees before callers (because I'm too lazy to write function declarations), but I'm fine if this is the preferred style here.

C7/PCXToGodot.cs
lucienmaloney: Looks like the performance difference here is even more minor than the main Pcx loop changes, so this can be reverted for readability's sake.

C7/PCXToGodot.cs
lucienmaloney: Good point about the compiler optimization- that didn't occur to me at the time. And my unscientific test of running the branch code 5 times vs the reverted code 5 times didn't reveal any performance changes here, so this is also revertible.

C7/PCXToGodot.cs
lucienmaloney: My gut is that of all the changes commented on so far, this is the one that will cause the biggest performance drop if reverted. However, I need to actually refactor these static variables back into local ones, which is a change I'm a bit too sleepy to test out right now. And yeah, it's not great in terms of future expectations, so maybe the old approach would be better here.

lucienmaloney:
Okay, so overall, I agree that I injected too much of my own micro-heavy style into this part of the code, so scaling things back a bit is probably correct. The good news, at least, is that I'm fairly confident I didn't insert any bugs into the code. But in any case, I'll loop into the case of whether the static memory allocation is actually doing anything for performance.

C7/PCXToGodot.cs
JimOfLeisure: Looks good to me. I wrote that a while back. I seemed to be coding shadows on the fly where here he's building a discrete palette with alpha-channel shadows. That works.

ConvertCiv3Media/ReadPcx.cs
QuintillusCFC: Oh, I see now.  It decrements ImgIndex, so the next (non-junk) byte will overwrite the junk byte.  So yes, it will work (and I really ought to manually verify that, too...).

Yeah, I'd say if we get 95-96% of the savings without the micro-optimizations, it isn't worth keeping the ones that make it harder to read, even in a slow way.  Some of them don't affect readability, so we'd probably wind up with 98% or so of the improvement.  I try to keep in mind Kernighan's quip about debugging being more difficult than writing code, and being as clever as possible when writing it means you can't debug it, and also remembering that whoever comes later may not be as clever with regards to whichever areas we have expertise in.

I also try to apply Amdahl's Law, which generally points me to macro-optimizations.  E.g. I've been loading saves from The Great War locally, and it takes about 8 seconds.  I haven't broken down why it takes so long, but I'm pretty sure most of the different isn't in PCX loading.  Once this is merged, we can apply some high level metrics and see what's next in the order of making loads take awhile... although we probably should get around to the SAV pull request before that point since that may have an impact as well.

C7/PCXToGodot.cs
QuintillusCFC: Honestly, one of the really nice things about Java and C# is not having to write function declarations.  It's a small thing, but it's one of the things I have to remind myself about when I go back and write some C.  Taking a look at some of my 2019 C code, sure enough, the callees are first, probably because I didn't want to have to write a .h file/include more definitions in the .h file.

It's also kind of a combined-effect thing.  I may well not have commented on the order if the public interfaces were the only public things, and if it had been in top-down order, I may have overlooked that the local methods weren't private reading top-down.

lucienmaloney:
Having the buffers be redeclared locally didn't cause as much of a drop as I expected, so that's been reverted too, along with the other loopy stuff and the fixing of function access specifiers and orders.

QuintillusCFC (approved):
Looks good and noticeably quicker!

WildWeazel:
any reason not to merge this one?

QuintillusCFC:
Nope, I think it's ready to merge.  Just waiting for Caro-Kann to merge it, since it isn't obviously blocking anything else.  Although it may be time to stop waiting and just merge it.

WildWeazel:
I'm fine with anyone merging once comments are resolved

PR #72, opened 2022-01-16
Case insensitive media paths

maxpetul:
Makes the game work on case sensitive filesystems by enabling Civ3MediaPath to ignore case when necessary. Without this change the game wouldn't work for me on Linux.

### DIFF
--- a/C7/Util.cs
+++ b/C7/Util.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 using Godot;
 using ConvertCiv3Media;
 
@@ -57,9 +58,12 @@ public class Util
 		if (System.IO.File.Exists(fullPath))
 			return fullPath;
 
-		// If that didn't work, do a case-insensitive search starting at the root path and stepping through each piece of the extension.
+		// If that didn't work, do a case-insensitive search starting at the root path and stepping through each piece of the extension. Skip
+		// this step if the root directory doesn't exist or if running on Windows. Skipping this step on Windows avoids a potential problem
+		// with mixed forward and back slashes in the path.
 		string tr = null;
-		if (System.IO.Directory.Exists(exactCaseRoot)) {
+		if ((! RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) &&
+		    System.IO.Directory.Exists(exactCaseRoot)) {
 			tr = exactCaseRoot;
 			foreach (string step in ignoredCaseExtension.Replace('\\', '/').Split('/')) {
 				string goal = (tr + "/" + step).Replace("//", "/"); // Sometimes tr will already end in a slash so goal will end up
--- a/C7/Util.cs
+++ b/C7/Util.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Collections.Generic;
 using Godot;
 using ConvertCiv3Media;
--- a/C7/Util.cs
+++ b/C7/Util.cs
@@ -41,6 +41,42 @@ public class Util
 		// Assuming 64-bit platform, get vanilla Civ3 install folder from registry
 		return (string)Microsoft.Win32.Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Infogrames Interactive\Civilization III", "install_path", defaultPath);
 	}
+
+	// Checks if a file exists ignoring case on the latter parts of its path. If the file is found, returns its full path re-capitalized as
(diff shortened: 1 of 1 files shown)


QuintillusCFC (changes requested):
I like the idea of this, and the approach of enumerating and finding a case-insensitive match is better than anything I'd thought of to solve this problem.

However, these changes break C7 on Windows.  I added some debugging where it prints out what it's looking for; an abbreviated version that illustrates the problem is here:

```
Goal is D:\Civilization III/Conquests
Candidate: D:\Civilization III\00000000.016
...
Candidate: D:\Civilization III\Conquests
Failed to find match for tr = D:\Civilization III

```

It's a forward-versus-backward-slash problem.  On Windows the file enumeration returns backward slash, but the code here is searching using a string match and forward slash (checking whether a file exists using forward slash works on Windows, but this is just a string match).

Java has `System.getProperty("file.separator");` that can be used to look up which file separator is appropriate on a system; I don't know if C# has an equivalent but if so, line 65 could be altered to use that instead of forward-slash everywhere.

Alternately, I had wondered, should we skip this on Windows/Mono, since it's doing extra work that Windows/Mono will handle automatically anyway?  Figuring out how to skip it on Windows/Mono would also allow C7 to work on both systems, even if another irregularly-capitalized file slips in.

maxpetul:
How annoying. It works on my Windows system, I tested before submitting the PR and tested again just now.  This might be an issue of .NET Core versus Mono. Looking back through the Babylon progress thread, you said you were using .NET Core, is that still the case? I was and still am using it, on both Windows and Linux.

I thought using forward slashes wouldn't be a problem since that's what the original code was doing. Anyway the easiest solution I think is to bypass the case insensitive search when running on Windows. I'll implement something to that effect but I'll need you to test it since I'm not seeing the issue for whatever reason.

QuintillusCFC:
Interestingly, on my Windows 10 laptop, it works both with and without your most recent commit to skip case insensitive searching on Windows.  It is using "dotnet CLI" as its build tool, so it is using .NET Core, and cannot find the build tools if I switch it to Mono (they must not be intalled).

I'll check on my desktop, too.  It's running Windows 8.1, so it also could be related to a Windows 10 change (and one that makes things work better for a change!  amazing).  I did post that I was using "dotnet CLI" in December, but I didn't specify in that thread whether I was on my desktop at the time.  I probably was, but better to double check.

QuintillusCFC:
Remembered to check this tonight.  What do you know, it still didn't work on my desktop, despite it also using "dotnet CLI".  But I had my laptop fired up as well, so I dug into it side by side.

It turns out, the difference was that on my desktop, the install path from the registry is "D:\Civilization III", and on the laptop it is "D:\Software\Games\Steam\steamapps\common\Sid Meier's Civilization III Complete\\".  Note that only the laptop's install path ends in a backslash.  Since the code doesn't add the backslash in this branch as it did before, it was looking at paths such as ` D:\Civilization IIIConquests/Conquests/Sounds/Menu/Menu1.mp3`, which weren't going to work.

I'd suggest modifying the GetCiv3Path() method like this:

```
		// Look up in Windows registry if present
		path = Civ3PathFromRegistry("");
		if (path != "") {
			if (path.EndsWith("\\"))
				return path;
			return path + "\\";
		}
```
That fixes the root cause; another place in the code that used that method could have run into the exact same problem, so it's better to have consistency that the GetCiv3Path() method will always return it with an ending slash, whether the registry has that slash or not.

maxpetul:
What other place in the code do you mean? I found two places where GetCiv3Path() is called then a forward slash is tacked onto the path it returns, Util.cs:19 and GlobalSingleton.cs:15. I wonder why that code didn't break, it may be that Godot is fixing the path in the case of the load dialog. My Civ 3 Install_Path registry entry ends in a backslash BTW.

We should standardize whether or not GetCiv3Path() returns a path with an ending slash, but that should also include the case where the path is read from an environment variable. The tricky thing there is we won't know whether to look for a forward or back slash. Another option is to use System.IO.Path.Combine instead of assembling paths manually, since that should handle all these problems for us. For now I've done that in FileExistsIgnoringCase, that should fix things at least the immediate problem.

QuintillusCFC:
I just meant any other place, thinking "there could be other places in the future".

I agree that we should standardize what format GetCiv3Path will return, and good call on making it consistent with when it returns an environment variable.  My thoughts are that either \ or / should be acceptable, but if there is neither, add a / .  On the other hand, I didn't know that System.IO.Path.Combine was a thing.  Switching to the branch on my desktop to verify right now... I expect it to work, but I did the last time, too.

QuintillusCFC (approved):
Verified that it works when there is no trailing (back)slash on the Civ3 registry path, with the new System.IO.Path.Combine changes!

PR #76, opened 2022-02-04
Water 1.0

QuintillusCFC:
Start differentiating land from water.  Human units can only move on land (sorry, no human boats yet).  Barbarians will build galleys if they are coastal, and galleys will only move on the coast, whereas barbarian warriors will stay on land.  Barbarian camps will also always be on land instead of sometimes being on water.

Also has a few stylistic changes as identified in part with Rider.  Some of these are whitespace fixes, so if the two halves of the comparison look the same, that's probably what's going on.

Finally, changed the wait time for AI turns to 1 second from 2.  Now that there is barbarian activity, we don't really need a whole two second wait between turns.

### DIFF
--- a/C7GameData/Tile.cs
+++ b/C7GameData/Tile.cs
@@ -2,6 +2,7 @@ namespace C7GameData
 {
 	using System;
 	using System.Collections.Generic;
+	using System.Linq;
 	public class Tile
 	{
 		// ExtraInfo will eventually be type object and use a type descriminator in JSON to determine
@@ -73,6 +74,11 @@ namespace C7GameData
 			Random rnd = new Random();
 			int index = rnd.Next(8);
 			return (TileDirection)(Enum.GetValues(TileDirection.NORTH.GetType())).GetValue(index);
+		}
+
+		public Tile RandomLandNeighbor() {
+			List<Tile> landNeighbors = neighbors.Values.ToList().Where(tile => !tile.baseTerrainType.isWater()).ToList();
+			return landNeighbors[new Random().Next(landNeighbors.Count)];
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


maxpetul:
Instead of creating a new Random object every time you need a random number, you should instead use the Random object stored in GameData (named "rng"). The problem with creating a new object each time is that it gets a new seed based on the system clock which will prevent us from implementing the preserve-random-seed option and more importantly will prevent us from keeping the stream of "random" numbers consistent for all players in a multiplayer game.

C7GameData/Tile.cs
```diff
@@ -74,6 +75,22 @@ public override string ToString()
 			int index = rnd.Next(8);
 			return (TileDirection)(Enum.GetValues(TileDirection.NORTH.GetType())).GetValue(index);
 		}
+
+		public Tile RandomLandNeighbor() {
+			List<Tile> landNeighbors = neighbors.Values.ToList().Where(tile => !tile.baseTerrainType.isWater()).ToList();
```
maxpetul: It's a minor thing but you don't need the first call to ToList here or the corresponding one in RandomCoastNeighbor. "Where" works on any IEnumerable, which is implemented by the values collection, so there's no need to gather the values into a list.

QuintillusCFC:
Good call, I had forgotten about Preserve Random Seed.  Also makes sense that we'll want it to be consistent in MP, should we get to that point.

I've pushed some changes for that; one note is that because our "Quick Start" option loads from the JSON that currently lacks an RNG, there's a hack to set the RNG once, in the load, if it's null.  Once we get a more built-out standard rule set/random game generation that shouldn't be necessary, but I don't want to scope creep this PR to incorporate that.

maxpetul (approved):

PR #77, opened 2022-02-05
Unit prototype revamp

QuintillusCFC:
1.  Instead of creating a new unit prototype every time we create a unit, have them be created at game start, and reference those prototypes.  This applies for initial unit spawning, barbarian unit spawning, and city production.
2. IProducable abstract class for things that can be built in cities.  Right now, only units, but that will change over time.
3. Basic city production AI is now in its own class in the AI folder.  Method interface will change, at this point the goal of the change is to have it not be stuck in the EntryPoint to the engine, but be its own component.

Note that this builds on the Water branch, so that one should be merged into Development before this one is merged into Water.  Otherwise we'll wind up with a review for Water that includes both and is twice as large.

### DIFF
--- a/C7GameData/GameData.cs
+++ b/C7GameData/GameData.cs
@@ -108,6 +108,7 @@ namespace C7GameData
             warrior.defense = 1;
             warrior.movement = 1;
             warrior.iconIndex = 6;
+            warrior.shieldCost = 10;
 
             UnitPrototype settler = new UnitPrototype();
             settler.name = "Settler";
@@ -116,6 +117,8 @@ namespace C7GameData
             settler.movement = 1;
             settler.iconIndex = 0;
             settler.canFoundCity = true;
+            settler.shieldCost = 30;
+            settler.populationCost = 2;
 
             UnitPrototype worker = new UnitPrototype();
             worker.name = "Worker";
@@ -124,6 +127,8 @@ namespace C7GameData
             worker.movement = 1;
             worker.iconIndex = 1;
             worker.canBuildRoads = true;
+            worker.shieldCost = 1;
+            worker.populationCost = 1;
 
             UnitPrototype chariot = new UnitPrototype();
             chariot.name = "Chariot";
@@ -131,6 +136,7 @@ namespace C7GameData
             chariot.defense = 1;
             chariot.movement = 2;
             chariot.iconIndex = 10;
+            chariot.shieldCost = 30;
 
             UnitPrototype galley = new SeaUnit();
             galley.name = "Galley";
@@ -138,6 +144,7 @@ namespace C7GameData
             galley.defense = 1;
             galley.movement = 3;
             galley.iconIndex = 29;
+            galley.shieldCost = 30;
             
             unitPrototypes["Warrior"] = warrior;
             unitPrototypes["Settler"] = settler;
--- a/C7GameData/GameData.cs
+++ b/C7GameData/GameData.cs
@@ -105,55 +105,14 @@ namespace C7GameData
         }
         private void CreateDefaultUnitPrototypes()
         {
-            UnitPrototype warrior = new UnitPrototype();
-            warrior.name = "Warrior";
-            warrior.attack = 1;
-            warrior.defense = 1;
-            warrior.movement = 1;
-            warrior.iconIndex = 6;
-            warrior.shieldCost = 10;
-
-            UnitPrototype settler = new UnitPrototype();
-            settler.name = "Settler";
-            settler.attack = 0;
-            settler.defense = 0;
-            settler.movement = 1;
-            settler.iconIndex = 0;
-            settler.canFoundCity = true;
-            settler.shieldCost = 30;
-            settler.populationCost = 2;
-
-            UnitPrototype worker = new UnitPrototype();
-            worker.name = "Worker";
-            worker.attack = 0;
-            worker.defense = 0;
-            worker.movement = 1;
-            worker.iconIndex = 1;
-            worker.canBuildRoads = true;
-            worker.shieldCost = 1;
-            worker.populationCost = 1;
-
-            UnitPrototype chariot = new UnitPrototype();
-            chariot.name = "Chariot";
-            chariot.attack = 1;
-            chariot.defense = 1;
-            chariot.movement = 2;
-            chariot.iconIndex = 10;
-            chariot.shieldCost = 30;
-
-            UnitPrototype galley = new SeaUnit();
-            galley.name = "Galley";
-            galley.attack = 1;
-            galley.defense = 1;
-            galley.movement = 3;
-            galley.iconIndex = 29;
-            galley.shieldCost = 30;
-            
-            unitPrototypes["Warrior"] = warrior;
-            unitPrototypes["Settler"] = settler;
-            unitPrototypes["Worker"] = worker;
-            unitPrototypes["Chariot"] = chariot;
-            unitPrototypes["Galley"] = galley;
+			unitPrototypes = new Dictionary<string, UnitPrototype>()
+			{
+				{ "Warrior", new UnitPrototype { name = "Warrior", attack = 1, defense = 1, movement = 1, iconIndex =  6, shieldCost = 10 }},
+				{ "Settler", new UnitPrototype { name = "Settler", attack = 0, defense = 0, movement = 1, iconIndex =  0, shieldCost = 30, populationCost = 2 }},
+				{ "Worker",  new UnitPrototype { name = "Worker",  attack = 0, defense = 0, movement = 1, iconIndex =  1, shieldCost = 30, populationCost = 1 }},
+				{ "Chariot", new UnitPrototype { name = "Chariot", attack = 1, defense = 1, movement = 2, iconIndex = 10, shieldCost = 20 }},
+				{ "Galley",  new SeaUnit       { name = "Galley",  attack = 1, defense = 1, movement = 3, iconIndex = 29, shieldCost = 30 }},
+			};
         }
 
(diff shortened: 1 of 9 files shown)


C7GameData/GameData.cs
```diff
@@ -105,23 +100,25 @@ public Player CreateDummyGameData()
 
             return humanPlayer;
         }
-
-        private void CreateStartingDummyUnits(Player humanPlayer)
+        private void CreateDefaultUnitPrototypes()
```
WildWeazel, line 106: protip: you can do this with a lot less code using collection initialization:
```
unitPrototypes = new Dictionary<string, UnitPrototype>()
{
  { "warrior", new UnitPrototype { name="warrior", attack=1, defense=1, movement=1, iconIndex=6, shieldCost=10 }
  ...
}
```

C7GameData/IProducable.cs
```diff
@@ -0,0 +1,13 @@
+namespace C7GameData
+{
+	/**
+	 * Represents something that can be produced by a city.
+	 * Known examples are Buildings and UnitPrototypes.
+	 */
+	public abstract class IProducable
```
WildWeazel: I'd prefer to declare these kinds of things as Interfaces with Properties, which are less brittle and better encapsulated. *deflects blame to Jon Skeet* https://csharpindepth.com/articles/PropertiesMatter

WildWeazel: also it seems to be spelled Producible

C7GameData/UnitPrototype.cs
```diff
@@ -21,5 +20,14 @@ public override string ToString()
         {
             return name + " (" + attack + "/" + defense + "/" + movement + ")";
```
WildWeazel: Some quick searching suggests this is optimized to be more efficient than it would in Java, but I think interpolation is the preferred method these days:
`return $"{name} ({attack}/{defense}/{movement})"`
Kind of a stylistic thing though. I find it more readable but open to suggestions.

C7GameData/GameData.cs
QuintillusCFC: Nifty!  That's way better than the style Rider was suggesting.  I've switched it over, and fixed Chariots being too expensive at the same time.

C7GameData/UnitPrototype.cs
QuintillusCFC: That makes sense.  I tend to default to the simple style because every language has its own syntax for the interpolated style, so I tend to forget which one is the correct one to use in the language I'm currently writing in, and spend more time getting sidetracked by that than it would take to write the non-interpolated version.  But it looks readable enough with interpolation in C#, even with the parentheses in this string making it harder to read than it would be in most cases.

Fun fact, string concatenation being slow in Java hasn't been true for about 15 years, at least outside of some edge cases.  It was either in Java 1.4 or 1.5 (the latter being released in 2004) when they added compiler optimizations to use the StringBuilder class under the hood (maybe StringBuffer if it was 1.4), which ended the performance problem of re-initializing a String for every item being +'ed together.  There are still a lot of old sites, or sites quoting old sites, that say string concatenation performance matters in Java, but in practice these days the right answer is "use whatever reads easier" 95% of the time.  The other 5% is typically when you're concatenating a _lot_ of strings for debug loggers, using an older logging framework, and not checking if debug logs are enabled before creating your strings.  Then it doesn't really matter how you're doing it, just that the volume can slow things down if it's e.g. printing a debug statement for every pixel of a 20 megapixel PCX file.

No idea if that applies to C# though, so it might make a more significant difference here.

I've changed this one and will try to remember this for future string interpolation events, now that I know the C# syntax for it.

C7GameData/IProducable.cs
QuintillusCFC: As the (misspelled!) name suggests, I initially had this as an interface.  But I'm having trouble getting this interface-with-properties thing working.  I can set it up like this:

```
	public interface IProducable
	{
		string name { get; set; }
		int shieldCost { get; set; }
		int populationCost { get; set; }
	}
```

But then any attempt to access name/shieldCost/etc. fails.  E.g. `return $"{name} ({attack}/{defense}/{movement})";` says "Cannot resolve symbol 'name'".

So I'd switched to abstract class so I could access the properties from sub-classes.

I have read on Microsoft documentation that, "Beginning with C# 8.0, an interface may define a default implementation for members, including properties."  So maybe we're on too old of a C# version?

QuintillusCFC: (I'm also the opposite of a Property expert; they're one of the few things in C# that doesn't have a fairly direct equivalent in Java, so the ins and outs of when/where they can be used aren't intuitive to me yet)

QuintillusCFC: Okay, I think I figured out how to do it via StackOverflow: https://stackoverflow.com/questions/1593413/how-to-implement-a-property-in-an-interface

It seems kind of weird that it's required to put the property on the interface but also the implementing class.  Then again, I guess that was true with Java before Java 9 (which added default implementations on interfaces), too.

I see why Jon argues for properties, especially the practical benefits.  But I guess I'm still a bit confused as to why we prefer interfaces to abstract classes here.  It sort of makes sense, as in "a unit fits the interface of something that's producible".  But the abstract class phrasing, "a unit is a producible thing", also seems to fit.  In lieu of being able to think of another parent class for UnitPrototype or Building, I'm half inclined to favor the easier syntax of the abstract class over the flexibility of the interface.

Abstract classes versus interfaces are, admittedly, something that I've found myself swapping both ways in projects such as my Civ editor.  Sometimes I start with an interface and decide not having to duplicate functionality would be really nice and the "is a" metaphor fits well enough to go abstract class.  Sometimes I start abstract class and realize the metaphor isn't as solid as I'd hoped.

Anyway, I've switched it back to an interface for now (commit incoming).  Still curious about the C# 8/code duplication of properties questions though.

C7GameData/UnitPrototype.cs
WildWeazel: Good to know about Java! I haven't written it since (Android's approximation of) 7. In C# they seem to be pretty equivalent inline, but if you're doing dynamic concatenation like you mentioned you should use StringBuilder.

C7GameData/IProducable.cs
WildWeazel: Interesting, I knew about default implementations but have not worked with them. We should be on C# 8, otherwise I don't think your snippet above would even compile with fields in an interface.

Conceptually interface makes sense here because Producible describes something that it does or has via public attributes, rather than an inheritable implementation. Since C# doesn't do multiple inheritance and it's gross anyway, IMO it's best to avoid inheritance (via interfaces or composition as appropriate) until you really need to. I imagine there are other things that some Producibles would rather extend later, even though as you said those are probably Producible themselves.

WildWeazel: Oh and also, don't worry about it right now since we probably need to reformat everything once we figure out coding standards, but properties are always supposed to be capitalized like methods.

QuintillusCFC: Yeah, I think that makes sense.  There is no IProducible that isn't also something else, so it's an interface.

Yeah, we've kind of all been bringing out own patterns to the project.  I started a thread to try to standardize some things in December, but that's about when my enthusiasm waned, so we didn't decide on much.  This is definitely an artifact of the Java standards being different than the C# ones, and my muscle memory still being in Java.

C7GameData/UnitPrototype.cs
QuintillusCFC: Good to know that there's a StringBuilder in C# as well!

QuintillusCFC:
@WildWeazel Do you want to make any more changes on this, or approve + merge it?  From the conversations it sounds like we're probably good to go for now, but they're still open.  I'd like to merge this one if there aren't more things to change as an increasing amount of code builds on it.

(We also might want to standardize who resolves comments, if your thought was I'd resolve them; on most teams I've been on the person who opened them resolves them unless they indicate they don't need to)

WildWeazel (approved):

WildWeazel:
The way I'm used to is the owner of the issue/branch is responsible for all commits unless they ask for help. For doing the merge, I've been on teams where it's assumed the branch owner will merge once approved, and others where there's an integration czar who performs all merges.

QuintillusCFC:
That makes sense in terms of fixing the issues.  Overall I think the model of the comment-leaver being the default person to resolve them makes more sense than the comment-fixer though; I've seen enough fixes for comments over the years where it's like, well, it sort of fixes that issue, but only in a specific case, or with a side effect.  On the other hand the second check does add more time, hence the "I know you can fix this, no need for me to re-check" option.

Integration czar... I haven't been on a team with one but knew of some adjacent teams that had one on their SVN project where they didn't merge anything until right before a release.  No idea how they didn't introduce any bugs with all the conflicts that must have caused, I guess that's why they had someone who specialized in it.

WildWeazel:
Ha, that sounds both miserable and vaguely familiar. This team has someone constantly tending CI and merges, but it's safety critical flight software so I guess protecting master kind of comes with the territory. Anyway I'm a fan of shared code ownership especially in open source, but I also hate the "are you doing it or waiting on me" ambiguity when there's no clear protocol. :shrug:

QuintillusCFC:
Yeah, as I recall you work in an area where some extra precautions are well-justified.  Most of the place I've worked have been less critical, although some have dealt both with web UIs were minor issues weren't super-important, and financial transactions where they were.  That was interesting as a consulting team telling them, "We know you hired us to help you move faster on the web component, and we know you're happy that things are moving faster now, but you really need to slow down and take things more cautiously on the financial transactions side."

And yes... the ambiguity over who's doing it is harder to figure out than in the office (and sometimes it slips by then...).

PR #78, opened 2022-02-06
Static variable in Util to hold Civ3Root and FileDialog and button to select it

### LINKED ISSUE #48: Adding dialog to select Civ3 Home if not detected

JimOfLeisure:


QuintillusCFC:
Could we add a check so that it only adds the link on the main page if the install location cannot be auto-detected via the Windows Registry?  It looks strange having "Select Civ3 Home Folder" display right over the top of the Tower of Babble PCX file.

I'm rebuilding my Linux VM currently, so I haven't tested it there yet.  Looks plausibly correct though.

QuintillusCFC:
So I realized that you are trying to hide it if it's already detected, i.e. if DisplayTitleScreen goes according to plan.  Thus, why was I seeing it?  I noticed it was printing the "Could not set up the main menu" message despite the main menu being clearly visible, so I added a , ex to that message so it printed the error.  It printed:

```
Could not set up the main menuSystem.NullReferenceException: Object reference not set to an instance of an object.
  at MainMenu.DisplayTitleScreen () [0x0013a] in C:\Development\Civ Related Projects\Prototype\C7\MainMenu.cs:57
```

After adding a few prints, line 57 was `SetCiv3Home.Visible = false;`

So, it seems like for some reason the DisplayTitleScreen method doesn't have proper access to SetCiv3Home.  Not sure why yet.

JimOfLeisure:
Huh, weird. I can replicate the issue on my Windows box. Checking into it. I don't see an access issue as it's a method in the class where the button is defined, and it doesn't freak out on build. But *something* is happening here....

JimOfLeisure:
Ok, fixed it. The button reference was getting assigned after the first DisplayTitleScreen run, so it doesn't exist if the menu succeeds the first time.

JimOfLeisure:
Made another commit. The first run of DisplayTitleScreen() didn't need the tailing initialization, but it hit me that it made no sense for DisplayTitleScreen() to not be at the very end of _Ready() lest we stub our toe on something like this again in the future. No reason to run code until the init is finished.

QuintillusCFC (approved):
Looks good from a Windows standpoint an in terms of the recent changes fixing the issue.  Haven't had a chance to run it off of Windows yet.

I did discover that GetCiv3Path() is returning null if you are on Windows and the registry doesn't have Civ3 in it, and that causes things to blow up, but that's been the case since before this change set, we just hadn't noticed because we all have Civ3 installed.

I'll write that up separately.

PR #82, opened 2022-02-10
Fix non square maps not working.

QuintillusCFC:
They broke with SAV-slurping support, but it appears the root cause had been there all along - a height/width swap.  This gets them working again.


JimOfLeisure (approved):
Ah, yeah, that would do it. Oops.

PR #83, opened 2022-02-10
Add the ability to open BIQ files.

QuintillusCFC:
This is likely limited to BIQ files with custom rules and custom maps
at this point.  The same amount of data is imported as for SAVs -
terrains, tiles, and map size.

Some of this, but not all, can be combined across BIQs and SAVs.  E.g.
terrains types can be combined, but tiles cannot be (e.g. overlays and
terraforms may have changed).


QuintillusCFC:
Closing in favor of #84 .

PR #84, opened 2022-02-10
Allow Opening BIQ Files

### LINKED ISSUE #108: Open BIQ Files Directly (in addition to SAVs)

QuintillusCFC:
1. Go to Load Game
2. Navigate to the scenarios folder
3. Choose a BIQ file, then click Open
4. It loads in-game.

Known issues: For some reason all pine forests are snow-covered.

Recommended to merge https://github.com/C7-Game/Prototype/pull/82 first, as it fixes non-square maps, which is a lot of BIQs.

### DIFF
--- a/C7GameData/ImportCiv3.cs
+++ b/C7GameData/ImportCiv3.cs
@@ -34,7 +34,6 @@ namespace C7GameData
             }
 
             // Import tiles
-            //TODO: This looks very swapped.
             c7Save.GameData.map.numTilesTall = civ3Save.Wrld.Width;
             c7Save.GameData.map.numTilesWide = civ3Save.Wrld.Height;
             int i = 0;
--- a/C7GameData/ImportCiv3.cs
+++ b/C7GameData/ImportCiv3.cs
@@ -37,8 +37,8 @@ namespace C7GameData
                     BaseTerrainFileID = civ3Tile.TextureFile,
                     BaseTerrainImageID = civ3Tile.TextureLocation,
                 };
-                int y = i / (civ3Save.Wrld.Width / 2);
-                int x = (i % (civ3Save.Wrld.Width / 2)) * 2 + (y % 2);
+                int x, y;
+                (x, y) = GetMapCoordinates(i, civ3Save.Wrld.Width);
                 Tile c7Tile = new Tile
                 {
                     xCoordinate = x,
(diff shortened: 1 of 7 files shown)


WildWeazel:
This is largely duplicated from ImportSav. Can we extract the logical blocks to common methods? eg. InitGameData, ReadBytes, ReadTerrain, ReadTiles...

QuintillusCFC:
Good point, and largely yes.  The major exception being tiles, since there are separate SAV and BIQ tile objects whose data may be different - although even there, we can standardize the variable names and structure more than they are currently.

QuintillusCFC:
All right, this ought to be good to go again.  Also wired up the Load Scenario button.  It turns out the Tile part that will differ between SAV and BIQ is a significant part of what we're importing, but as we import more BIQ rules, there should be more shared import code.

C7GameData/ImportCiv3.cs
```diff
@@ -68,11 +61,65 @@ static public C7SaveFormat ImportSav(string savePath, string defaultBicPath)
...
+			int i = 0;
+			foreach (QueryCiv3.Biq.TILE civ3Tile in theBiq.Tile)
+			{
+				Civ3ExtraInfo extra = new Civ3ExtraInfo
+				{
+					BaseTerrainFileID = civ3Tile.TextureFile,
+					BaseTerrainImageID = civ3Tile.TextureLocation,
+				};
+				int y = i / (theBiq.Wmap[0].Width / 2);
+				int x = (i % (theBiq.Wmap[0].Width / 2)) * 2 + (y % 2);
```
WildWeazel: is one of these supposed to use Height? same in the method above.

QuintillusCFC: Nope, that's correct.  The x uses modulus instead of division, so e.g. if i = 73 and width = 100, it will be 73 % 50 = 23.  Then times 2 = 46.  Plus y %2 = 1, X = 47, Y = 1 would be the 73rd tile (base zero) in the list.

Definitely not intuitive with the isometric grid, but it appears to be correct (other than using BIQ data for the width, it's actually copied from code Lucien wrote in December).

It probably does make sense to have it be shared across the two areas though; I'm adding a shared function now.

QuintillusCFC (commented):
Self-reviewed as this is the longest-open non-approved PR at 7 days, and changes have already been made in response to @WildWeazel 's comments.  Noticed that the helper methods in ImportCiv3.cs are inconsistent in whether they take the BIQ or the C7SaveFormat first, so I'm going to standardize them as Civ3 (source) first, C7 (destination) second.  Other than that, I think it's ready to go, and I'm looking forward to being able to open BIQs in Development.

PR #85, opened 2022-02-13
BFS (Breadth-First Search) Pathing

### LINKED ISSUE #99: AI Cannot navigate around bays/inlets
In PR #86 , the settlers can spread out across the continent... except some get stuck trying to go around bays due to the shortest path being across water.

This will improve their pathing so they can go around such obstacles.

Covered in PR #85 .

QuintillusCFC:
(Please take a look at SlightlyImprovedAI first, if it hasn't already been reviewed).

This branch implements breadth-first-search for pathing units.  This is an early step; an outline of thoughts for future improvements is included in PathingNotes.txt, and more sophisticated algorithms will be added in future branches.

Compared to the simple as-the-crow-flies pathing previously used, this has a major advantage in that the AI won't get stuck if the as-the-crow-flies pathing would have taken it over water.  It can figure out a way around.

Also of note is the first test using a test framework, for part of the BFS algorithm.


QuintillusCFC:
This is supposed to be a draft, but isn't... seeing if I can get it that way by closing it and re-opening.

QuintillusCFC (commented):
As noted, there are some future improvements to be made, but this will accomplish its goal.  Merging so it doesn't sit in the queue forever.

PR #86, opened 2022-02-13
Slightly Improved AI

### LINKED ISSUE #97: Settler AI should consider where other settlers are going
Covered in PR #86 .  The AI should not send settlers where other settlers are already going.

### LINKED ISSUE #98: Settlers should be willing to settle outside of their home city area
In the first iteration, settlers would only settler near their home city.

They should be willing to travel farther afield, as all the good locations near their home city will eventually be taken.

Covered in PR #86 .

QuintillusCFC:
Two main improvements versus BasicOpponents:
 - The AI will consider sites Settler locations near cities other than its first city, and will take into account where other Settlers are going so they don't wind up settling at the same tile or immediately next to one another.
 -  The AI will now fortify some units in its cities.  It isn't very good at this yet, but at least it's not totally random whether a city will have defenses.

This should be a significant improvement over BasicOpponents for settling, but the AI still often fails to colonize its whole continent due to the pathing being too simple, using only a greedy algorithm.  The follow-up on this, #85 , will address that problem.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -66,7 +66,8 @@ public class Game : Node2D
 
 			Toolbar = GetNode<Control>("CanvasLayer/ToolBar/MarginContainer/HBoxContainer");
 			Player = GetNode<KinematicBody2D>("KinematicBody2D");
-			GetTree().Root.Connect("size_changed", this, "_OnViewportSizeChanged");
+			//TODO: What was this supposed to do?  It throws errors and occasinally causes crashes now, because _OnViewportSizeChanged doesn't exist
+			// GetTree().Root.Connect("size_changed", this, "_OnViewportSizeChanged");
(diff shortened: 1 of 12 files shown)


QuintillusCFC (commented):
Self-reviewed.  I'm reminded that we could benefit from a logging framework, and wrote up Issue #118 to cover that.  Otherwise:

1. The check for not building cities on mountains will conflict with the changes to not use hard-coded terrain names in decisions.  That's pretty much an argument for getting this merged in, though.
2. Probably want to refactor PlayerAI into a few sub-classes at some point; there are already sub-methods.  But that should be its own PR.
3. Probably a case of "a .NET expert would have a better way", but being a Java expert, the use of goto statements still seems like the best option I know of in .NET.  We can change 'em later.

PR #87, opened 2022-02-13
Rivers 1.0

### LINKED ISSUE #57: River support
Display rivers on the map.

TBD: How much we care about "special" river features like corner-only rivers, waterfalls, and deltas at this point in time.

QuintillusCFC:
This includes basic river displaying support, on par with my editor, but lacking some advanced features such as flood plains overlays and deltas.  I think it's an appropriate level for a near-term Babylon release, with enhancements for the Carthage release.

There's also some BIQ-opening code that got in this branch, as I was using BIQs to test the river code due to convenience.  We should probably get that branch approved and merged first.  But if you want to explore the river part of the branch, go ahead.


QuintillusCFC (commented):
Self-reviewed.  Looks good; C7 TILE has been being branched in multiple directions so getting it merged would be good.

QuintillusCFC:
Aaand only after leaving a comment does it say there are conflicts.  Was kinda wondering why it hadn't already.

PR #109, opened 2022-02-14
Added BonusShield (bonus grassland) data only

JimOfLeisure:
- Not sure why we're using if statement, but going with the flow
- This needs to be verified, but I'm reasonbly confident this is right
- Note that bonus will need to be hidden when under overlay terrain
- Also note that mountains and hills may have bonus shields under them

**I'm just throwing this up real quick to show where the bonus grassland info is. We may or may not want to merge it as-is.**


QuintillusCFC:
Looks correct.  What are your thoughts on basing it on the branch that this PR has: https://github.com/C7-Game/Prototype/pull/84, and also extracting it for BIQ files?  I'd like to try to keep BIQ and SAV support in sync where possible going forward, but unfortunately they do have two different TILE structures.

JimOfLeisure:
> but unfortunately they do have two different TILE structures.

Well, that's the thing. I haven't looked at BIQ maps, so I don't know where it is in the BIQ tile. I'm also just now peeking at your BIQ reader.

A wild thought: Perhaps we can have ~~an interface ITileType for the tile~~, and the BIQ and SAV sections can implement the same interface so the differences are handled there and the C7 GameMap code doesn't have to worry about whether it's talking to a BIQ tile or SAV tile. Updated idea: just have one Tile inherit from the other and override as necessary.

QuintillusCFC:
We sort-of have that in a way?  At least in that there's a C7 tile, and the C7 code, including the map, only deals with it, and it has/will have the properties of both types of Civ3 tiles.  But there's the two types of Civ3 tiles.  I did think, but what if the SAV one is an extension of the BIQ one?

But there's the problem that as they are now, they're loaded directly into memory, as whole C-style structs. - `[StructLayout(LayoutKind.Sequential, Pack=1)]` .  Which is efficient, but means I'm far from sure that it's possible for the SAV one to inherit - and even if it can, it will have to have all the same data, plus then some additional, which I'm not sure it does.

Overall I'm inclined to think that we're doing all right having the import script be the only place that has to worry about the differences.  Although if you know of a way to shoe-horn them together and save us nearly half the TILE work, I won't stop you.

`Well, that's the thing. I haven't looked at BIQ maps, so I don't know where it is in the BIQ tile. I'm also just now peeking at your BIQ reader.`

Right, I tend to forget that I'm the BIQ person here.  It turns out it's in the same byte, same position, three bits down from SNOW_CAPPED_MOUNTAIN in the same direction.  Pulled up the Poly docs and they didn't help, but it was in my editor code.

JimOfLeisure:
Well, in answer to the original question, I'm fine with whatever. I just pushed the code to make sure we know where the BG info is.

Sean-Brown:
the PR looks pretty straight forward, just resolve the merge conflict and it should be good to go

QuintillusCFC:
The remaining item was applying the changes for BIQs as well.  I just resolved the conflict and added that.  But that means I probably shouldn't be approving what's now in large part my changes...

Sean-Brown:
Your changes look fine too, just some indentation issues that should be fixed before merging.

QuintillusCFC:
Fix the indentation issues... gonna make sure there's an issue for fixing that more globally (and/or in Visual Studio).  Going to merge since there's no reason not to.

PR #111, opened 2022-02-15
Fixes #79 error when winreg key not present

### LINKED ISSUE #79: GetCiv3Path() returns null if Civ3 isn't in the Windows registry, and that causes issues
Stumbled upon this while reviewing Jim's PR.

Steps to reproduce:

1. Find a Windows box without Civ3 installed, or (more plausibly, since we all have Civ3 installed), rename the key HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Infogrames Interactive\Civilization III to something else temporarily, so C7 won't find it.
2. Run C7

Expected: You are able to select the Civ3 home.
Actual: C7 crashes.  Output:

```
Unhandled Exception:
System.ArgumentNullException: Value cannot be null.
Parameter name: path1
  at System.IO.Path.Combine (System.String path1, System.String path2) [0x00003] in <1400ea11d6fd49d4939be686a9549fb4>:0
  at Util.FileExistsIgnoringCase (System.String exactCaseRoot, System.String ignoredCaseExtension) [0x00001] in C:\Development\Civ Related Projects\Prototype\C7\Util.cs:60
  at Util.Civ3MediaPath (System.String relPath, System.String relModPath) [0x0008e] in C:\Development\Civ Related Projects\Prototype\C7\Util.cs:109
  at MainMenuMusicPlayer._Ready () [0x00002] in C:\Development\Civ Related Projects\Prototype\C7\MainMenuMusicPlayer.cs:14
```

The fact that it can't find the main menu music is a red herring, the real problem is that it's even trying to despite the registry not containing the expected key.  The relevant code in GetCiv3Path():

```
                // Look up in Windows registry if present
		path = Civ3PathFromRegistry("");
		if (path != "") return path;
```

It turns out that the method we use in Civ3PathFromRegistry to check the registry returns null if it doesn't find anything, but out code expects it to return the empty string.

Of note is that I've tested this on Windows 8.1 64-bit.  I know the Windows registry format has not been entirely the same for the past 30 years, so it's possible that it does return the empty string on another version of Windows.  I have separate registry-parsing code for pre-and-post-Vista in my editor, so it's not a wholly theoretical possibility.

JimOfLeisure:
I originally misunderstood the default value parameter; it only triggers when the key exists but the value doesn't, which isn't likely. So returning null for both cases of key with no value and of no key.

Tested on Windows with renamed value, renamed key, and proper entries present. Doesn't break Mac.

### DIFF
--- a/C7/Util.cs
+++ b/C7/Util.cs
@@ -31,17 +31,18 @@ public class Util
 		if (path != null) return path;
 
 		// Look up in Windows registry if present
-		path = Civ3PathFromRegistry("");
-		if (path != "") return path;
+		path = Civ3PathFromRegistry();
+		if (path != null) return path;
 
 		// TODO: Maybe check an array of hard-coded paths during dev time?
 		return "/civ3/path/not/found";
 	}
 
-	static public string Civ3PathFromRegistry(string defaultPath = "D:/Civilization III")
+	static public string Civ3PathFromRegistry()
 	{
 		// Assuming 64-bit platform, get vanilla Civ3 install folder from registry
-		return (string)Microsoft.Win32.Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Infogrames Interactive\Civilization III", "install_path", defaultPath);
+		// Return null if value not present or if key not found
+		return (string)Microsoft.Win32.Registry.GetValue(@"HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Infogrames Interactive\Civilization III", "install_path", null);
 	}
 
 	// Checks if a file exists ignoring case on the latter parts of its path. If the file is found, returns its full path re-capitalized as


QuintillusCFC (approved):
Looked over and tested on Windows, we now get the "Unable to get graphics" message if it's missing from the registry instead of crashing.  Thanks for taking out the D:\Civilization III too, that must have slipped in!

JimOfLeisure:
> Thanks for taking out the D:\Civilization III too, that must have slipped in!

I left it in there from way back when the dir was hard-coded and would tend to get updated for whichever dev pushed last. I needed a default value so had to pick something. I think now the default path is "/civ3/not/found" or similar, just to have something to return. I guess we could generate an exception and use try/catch, but I'm not sure how much that actually gains us for the effort.

QuintillusCFC:
I'm going to go ahead and merge this since it's approved and that closes one of the issues...

Issue #32, opened 2021-11-29
General-purpose mp3 support (without duplicating the main menu approach everywhere)


QuintillusCFC:
Once we have another mp3, general-purpose-ize our mp3 support.

QuintillusCFC:
Moving to Babylon as it's definitely not in Aztec.

QuintillusCFC:
De-assigning to Carthage as it doesn't match up with any of our proposed Carthage priorities.

PR #112, opened 2022-02-15
Allow volume configuration via INI file

### LINKED ISSUE #34: Main volume/music volume support in .ini file
Add a configurable INI file if it isn't already present.  If there's a built-in C# library to handle them, great, if not, it's an INI file, what could go wrong with building our own?

Add a volume and music volume options to it.  Probably an integer from 0 to 100.  Have the sound react to this value, at least for music.

Borderline whether it's a good first issue, but the risk is low so why not?

QuintillusCFC:
Valid values are 0 to infinity, although it sounds really odd due to clipping if you set it to over 9000 (it might also break your windows if you have powerful enough speakers).  I imagine we will allow 0-100 in game, although there's an argument to be made for 0-200, similar to VLC, for those with weak speakers, e.g. some business laptops.

It will create the INI file on first run, with the default value of 100.  You can then change it, with 0 disabling it entirely.

Future improvements:
 - GUI control in-game
 - Right now the main menu music is on the main audio bus.  We should give it is own bus so we can control its volume separately from everything else.  This relates to issue #32 .

To verify: That it works with a compiled/deployed system, not just within the Godot editor.

### DIFF
--- a/C7Engine/C7Settings.cs
+++ b/C7Engine/C7Settings.cs
@@ -1,0 +1,40 @@
+using System.IO;
+using IniParser.Exceptions;
+
+namespace C7Engine
+{
+	using IniParser;
+	using IniParser.Model;
+	
+	public class C7Settings {
+		private const string SETTINGS_FILE_NAME = "C7.ini";
+		public static IniData settings;
+
+		public static void LoadSettings() {
+			try {
+				settings = new FileIniDataParser().ReadFile(SETTINGS_FILE_NAME);
+			}
+			catch(ParsingException ex) {
+				//First run.  The file doesn't exist.  That's okay.  We'll use sensible defaults.
+				settings = new IniData();
+				SaveSettings();
+			}
+		}
+
+		public static void SaveSettings() {
+			new FileIniDataParser().WriteFile(SETTINGS_FILE_NAME, settings);
+		}
+
+		public static void SetValue(string section, string key, string value)
+		{
+			settings[section][key] = value;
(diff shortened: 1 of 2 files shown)


QuintillusCFC:
Verified that it works as expected in the exported version.

(Also, for the question of, "shouldn't we put the INI file somewhere else?", the answer is probably yes, but that can be decided in a future PR.  The main goal of this one is making the volume configurable)

WildWeazel:
There's a broader conversation to be had there: what kind of things should be in an INI (or equivalent) vs mod, where should it be found / should there be a hierarchy of them, etc.

C7Engine/C7Settings.cs
```diff
@@ -0,0 +1,40 @@
...
+			}
+		}
+
+		public static void SaveSettings() {
+			new FileIniDataParser().WriteFile(SETTINGS_FILE_NAME, settings);
+		}
+
+		public static void SetValue(string section, string key, string value)
+		{
+			settings[section][key] = value;
```
WildWeazel, line 30: I haven't looked at the API- does this create the section/key if they don't already exist?

QuintillusCFC: It does.  After running this branch on a fresh computer, I see the following in the AI file:

```
[audio]
musicVolume = 100
```

I haven't dug into how they got it to work like that, but I like what I've seen so far.

WildWeazel (approved):

PR #115, opened 2022-02-18
This allows importing Civ3 scenarios with non-default terrain names (…

### LINKED ISSUE #58: Refactor terrains rendering to not rely on hard-coded names.
Right now, imported games from non-English variants of Civ will not render the map correctly as the names are hard-coded to English.

Puppeteer and I have discussed this on the forum and have agreed that a key is a decent way to implement this, where the key may be the English name.  This still allows for a more user-friendly approach to editing scenario files than the integer approach that Civ uses.

QuintillusCFC:
…including defaults in non-English versions), and having them work.

There's a lookup table for the Civ3 types, and the Name property will henceforth only be used for display purposes.  See the issue (#58 ) for a test scenario in German.


QuintillusCFC:
Note: I need to re-gen the JSON file from my desktop; for some reason my laptop can't find the .NET 4.7.2 Targeting Pack to do so even after re-installing it.  But other than that, I think this is ready to review.

WildWeazel (approved):
Sure, I'll approve a 13000 line diff :D
It might be a good idea to tag generated files as binary so git doesn't try to diff/merge them.

This branch is relevant to a comment I just made in Feature Requests about possibly using tags to indicate terrain features, ie. potential HasTag(tag), NeighborsTag(tag) functions.

QuintillusCFC:
Huh, I didn't know it was possible to mark text files as binary in Git.  I agree that 13000 line diffs are unhelpful.  But the teams I'd worked on had resolved that by not committing generated files; we never knew about .gitattributes.  I've attempted to mark it as binary here, although it still shows up for now... not sure if that means I did it wrong, or it's because I marked it after making the JSON changes.  Going to merge anyway, before this branch picks up more conflicts.

PR #119, opened 2022-02-18
AI Won't Build Cities Right by Cities Belonging to Other AIs

### LINKED ISSUE #117: With Multiple AI, They Shouldn't Settle Cities on the Same Tiles
If you have multiple AIs (in the BFSPathing branch), and they start on the same continent, they'll build cities on the same tiles as each other, or directly next to tiles with cities belonging to other AIs.

They shouldn't; there can only be one city per tile.

QuintillusCFC:
Teach the AI not to build cities where other AIs have already built cities.  If a settler is heading towards a destination, and that tile or a neighboring tile has a city, then it will seek a new destination instead.  This new destination will take into account its current location, i.e. it will prefer locations closer to it rather than walking back across the empire, if a location nearby is available.

This branch adds two new AI players (for a total of 3).


QuintillusCFC (commented):
For the main task, it looks okay, but it's going to have conflicts with Flintlock's changes, for example in PlayerAI.cs.  That isn't all bad, some of the conflicts like the unit movement ones will be better afterwards.  But I'm not going to try to rebase it right now.  Hopefully it'll be merged tomorrow.

QuintillusCFC:
Tested things out with the latest changes.  This change appears to be working as expected.  I did notice that with the latest architectural changes, the AI isn't sending out galleys anymore - which goes back to what I mentioned in #148 .  But that's not related to the settler changes here.  It was actually kind of fun watching the AI duke it out with animations, the Dutch and Greeks had a big border war that went back and forth as they randomly attacked each other.

PR #122, opened 2022-02-24
Threaded arch, animations, and combat

maxpetul:
This PR contains what I've been working on for the past couple of months, starting with the threaded engine architecture (described here: https://forums.civfanatics.com/threads/architecture-thread.674254/#post-16206357), including work on unit animations (enabling the engine to wait on them, spawning explosion effects, and other small things), and up to an initial implementation of unit combat.


C7/C7.ini
```diff
@@ -0,0 +1,2 @@
+[audio]
+musicVolume = 100
```
QuintillusCFC: Minor: We shouldn't commit this file as doing so will cause our settings to get bundled as the default (this is mostly on me as I didn't remember to add it to .gitignore when it was added).

C7Engine/EntryPoints/UnitInteractions.cs
```diff
@@ -84,85 +75,21 @@ private static MapUnit UnitWithAvailableActions(MapUnit unit)
...
-                    unit.movementPointsRemaining = 0;
-                    return;
-                }
-            }
-            Console.WriteLine("Failed to find a matching unit with guid " + guid);
+            new MsgSkipUnitTurn(guid).send();
         }
 
-        public static void waitUnit(string guid)
+        public static void waitUnit(GameData gameData, string guid)
```
QuintillusCFC, line 91: Shouldn't this one also create a message to the engine?  I've been using the wait action right away to make the camera center on my units (which your other PR will make unnecessary), and noticed it isn't working anymore, and that later on if I select a fortified unit, wait it, and then fortify it, the turn doesn't register as being over (whereas if I just select it and fortify it, it does).  Haven't dug all the way into why (started with functional testing), but I suspect it may be related to there not being a MessageToEngine for wait whereas there are for other actions?

C7Engine/MapUnitExtensions.cs
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	public static void fortify(this MapUnit unit)
+	{
+		unit.facingDirection = TileDirection.SOUTHEAST;
+		unit.isFortified = true;
+		unit.animate(MapUnit.AnimatedAction.FORTIFY, false);
+	}
+
+	public static bool fight(this MapUnit unit, MapUnit defender)
```
QuintillusCFC, line 24: I like how these read, in particular this one, with the animations integrated.

C7GameData/AnimatedEffect.cs
```diff
@@ -0,0 +1,14 @@
+// My intention is that eventually AnimatedEffect will be a class with some useful stuff in it. Right now it's just an enum that lets the engine
+// communicate to the UI which Civ 3 special effect it wants to trigger. The UI must provide the file names for the animations, unfortunately they
+// can't be guessed based on the enum names (like MapUnit.AnimatedAction) because the files aren't all stored in predictable folder locations.
+namespace C7GameData
+{
+public enum AnimatedEffect {
+	Hit,
+	Hit2,
+	Hit3,
+	Hit5,
```
QuintillusCFC, line 13: Potentially silly question: What happened to Hit4?

More serious comment: It might be good to have a comment explaining why there are Hit, Hit2, Hit3, and Hit5.  This is probably documented on CFC somewhere?  But as someone who doesn't know the nitty-gritty on this, what's likely to happen is I'll see one of the code references using Hit5, dig down here, and still not really know why it called that one instead of one of the others.

QuintillusCFC:
Looked at C7.ini, MessageToEngine.ini, MapUnitExtensions.cs, AnimatedEffect.cs, GameData.cs, and MapUnit.cs.  Left a couple comments but the only significant one is the one about Wait no longer working.

C7/C7.ini
maxpetul: I usually try not to commit extraneous things but this one must have slipped through. I'll delete it.

C7Engine/EntryPoints/UnitInteractions.cs
maxpetul: I'd rather not require an engine message to wait units because in multiplayer we either have to (1) sync that message with the server and suffer the latency or (2) carve out an exception where that message, unlike all the others, is not synchronized. I think it's better to think of the methods that wait units as a little utility library that the engine provides for the UI independent of the gameplay stuff.

As for why the UI doesn't always recognize when all units have been moved, we'll need to go over all that logic carefully at some point and work it out. I've noticed the same thing happening occasionally even though I don't use the wait command. There's a lot of room for edge cases in there since unit fortification happens asynchronously and we are (or at least I am) trying to allow for units to be fortified and selected at the same time. For example (I don't know if you've seen it) I had to add that awkward KeepCSUWhenFortified variable to the Game class. It's necessary so the autoselector doesn't prevent the player from selecting fortified units but adds another piece of mutable state that might be interacting badly with other things.

C7GameData/AnimatedEffect.cs
maxpetul: You'd have to ask someone at Firaxis what happened to Hit4. Those names come from the animation files in civ3PTW/Art/Animations/Trajectory. I wasn't sure how we should go about letting the engine tell the UI which effect animation to trigger, so to start I did the lazy thing and put an enum with one value for each Civ 3 animation file. Eventually we'll probably want a more abstract specification, i.e. the engine will just say to spawn a "hit" animation and let the UI pick one randomly from the Civ 3 ones. The hard part is that has to be made extensible. We don't want to assume the Civ 3 animations are available, and modders might want to add new effect animations or pick one non-randomly, etc.

QuintillusCFC: Interesting, that's a directory I'd never studied.  Looking at the files in FLICster, I'm not sure if "random" would be right though.  For example, hit.flc is 20 frames with a time of 833, and hit2.flc is 18 frames with a time of 692 (milliseconds?).

The comment helps, and we're probably good for now... it seems to be working.  But we might need to not only make it extensible, but see if we can figure out why there are 4 hit INI/FLC options... maybe different ones are used in different cases.

C7Engine/EntryPoints/UnitInteractions.cs
QuintillusCFC: I think there's a challenge in that the wait action requires some state to be stored to know which actions have been waited.  That does not need to be sychronized in multiplayer, but waited units aren't selected until all other ones have been selected, so the fact that a unit has been waited needs to be stored somewhere (and I added that when I first added the functionality).

I saw the KeepCSUWhenFortified, but don't fully understand the goal.  Is the idea like in Civ4 where you can select a fortified unit and it isn't automatically un-fortified?  My understanding is that's important in Civ4 because fortification bonuses grow over several turns, but it isn't the case in Civ3.  That is something we could add, although I hadn't seen it on the radar anywhere.

I guess the bigger view, though, is that this is a regression; Wait works properly in Aztec, but with this change it won't work properly in Babylon.

maxpetul: The wait command was broken because it wasn't autoselecting the next unit after the currently selection was set to wait. I must have carelessly deleted the call to GetNextAutoselectedUnit, I reinserted it and now waiting works again.

KeepCSUWhenFortified is to deal with the fact that fortification happens asynchronously and I also want to allow for units to be selected while fortified. The code originally called GetNextAutoselectedUnit after the fortify command was issued, but that doesn't work now that fortification is async b/c the unit won't actually be fortified until the message is processed. So I deleted that call to GetNextAutoselectedUnit and inserted some code that's run every tick that checks if the currently selected unit is fortified and, if so, calls it then. But that prevents fortified units from being selected so I had to add KeepCSUWhenFortified to stop that from happening as needed. I don't really like KeepCSUWhenFortified and we can probably ditch it later when/if we have a way for the engine to let the UI know the fates of messages, i.e. when they've been processed or if they've been rejected.

QuintillusCFC: Wait looks good now.

It does feel a bit like the block in line 169 of Game.cs is becoming a bit complicated.  After unwinding the parentheses, I'm pretty sure it means:

Get the next unit if:

 - There is a unit selected, AND
 - Either:
    - It's fortified and we should get the next unit when it's fortified (Civ3 default)
    - Both:
      - It's out of movement points OR it's dead (in other words, it's out of actions)
      - It's not animating

We might want to make some helper methods to reduce the complexity of that conditional (especially if my understanding is not accurate).

Is the concern with "that doesn't work now that fortification is async b/c the unit won't actually be fortified until the message is processed" that it might return the already-selected unit because it hasn't processed the command?  It feels like an early sign of network syncing nightmares to come.  Although I wonder if it's possible to bypass the concern somehow?  E.g. if it were going over a real network, and using TCP, I'm 99% sure that TCP enforces the order of messages.

QuintillusCFC (approved):
The parts I've looked over in detail look good (one possible refactoring mentioned, but that can be punted past Babylon), and functionally it is working well.

WildWeazel:
commenting so I can remember to come back to better understand what's going on here

PR #124, opened 2022-03-08
Initialize camera location on game start to capital or starting settler

maxpetul:


QuintillusCFC (approved):
Looks good, seems to work... probably will be able to notice it more once the zoom starts at 100% too.

PR #125, opened 2022-03-08
Visual Studio 2019 Debugging Support

QuintillusCFC:
See the thread at https://forums.civfanatics.com/threads/visual-studio-2019-impressions.675739/

This gets Visual Studio Community 2019 working including debugging support.

Visual Studio 2017 is not supported by the Godot extension, and on Visual Studio 2022 the extension + workaround to get it working breaks IntelliSense, so 2019 is the only version supported.


WildWeazel:
Will it be possible to debug from both Community and Code (and Rider?) with the same project files?

QuintillusCFC:
Yup for Rider!  This is purely additive, Rider ought to still work just as described in [this post](https://forums.civfanatics.com/threads/jetbrains-rider-impressions.675190/#post-16220505).  Rider actually gets half a point because it didn't need any changes to the project structure to debug, as Visual Studio does.

I'm not sure about VSCode, as I've yet to get debugging working there (that was my primary motivation for trying Rider in the first place).  I have seen a hint that it may be possible, but don't know the details yet.

QuintillusCFC:
Merging since WW noted he has no objections at CFC (although he doesn't have VS 2019), and merging will allow VS debugging - helping both me and potential future VS developers.

PR #126, opened 2022-03-09
Horse taming

### LINKED ISSUE #127: Horses are Everywhere
If you load the static JSON, every tile has horses.

This is due to a serialization bug, combined with object equality.

QuintillusCFC:
This reduces horse presence to just the tiles where they are supposed to appear.

This is done by changing resources on tiles within the JSON to be a reference, rather than a whole embedded object.  The whole embedded object was problematic because object equality did not hold between each instance when loaded from the flat JSON file (it did when everything was done in-memory, which is why the problem wasn't present for loading BIQ and SAV files).

I've gone with the "key" approach rather than using the name variable as the key because Civ using the name as an identifier in many places was one of the reasons internationalization of Civ was difficult.  By having separate identifiers, I18N of a scenario will be much easier - you could simply jump in the JSON and change the names of all the resources, and everything else would keep working.  (Later on, a similar approach to units will also enable renaming them without having to rename all their folders and files)

### DIFF
--- a/C7/MainMenu.tscn
+++ b/C7/MainMenu.tscn
@@ -89,8 +89,6 @@ window_title = "Open a File or Directory"
 resizable = true
 mode = 3
 access = 2
-current_dir = "/Development/Civ Related Projects/Prototype/C7"
-current_path = "/Development/Civ Related Projects/Prototype/C7/"
 
 [connection signal="pressed" from="CanvasLayer/SetCiv3Home" to="." method="_on_SetCiv3Home_pressed"]
 [connection signal="dir_selected" from="CanvasLayer/SetCiv3HomeDialog" to="." method="_on_SetCiv3HomeDialog_dir_selected"]
--- a/C7/MainMenu.tscn
+++ b/C7/MainMenu.tscn
@@ -89,8 +89,8 @@ window_title = "Open a File or Directory"
 resizable = true
 mode = 3
 access = 2
-current_dir = "/Users/jim/code/src/godot/Prototype/C7"
-current_path = "/Users/jim/code/src/godot/Prototype/C7/"
+current_dir = "/Development/Civ Related Projects/Prototype/C7"
+current_path = "/Development/Civ Related Projects/Prototype/C7/"
 
 [connection signal="pressed" from="CanvasLayer/SetCiv3Home" to="." method="_on_SetCiv3Home_pressed"]
 [connection signal="dir_selected" from="CanvasLayer/SetCiv3HomeDialog" to="." method="_on_SetCiv3HomeDialog_dir_selected"]
--- a/C7GameData/ImportCiv3.cs
+++ b/C7GameData/ImportCiv3.cs
@@ -1,6 +1,5 @@
 using System.Collections.Generic;
 using System;
-using System.Linq;
 
 namespace C7GameData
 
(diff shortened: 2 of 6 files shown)


C7/MainMenu.tscn
```diff
@@ -89,8 +89,8 @@ window_title = "Open a File or Directory"
 resizable = true
 mode = 3
 access = 2
-current_dir = "/Users/jim/code/src/godot/Prototype/C7"
-current_path = "/Users/jim/code/src/godot/Prototype/C7/"
+current_dir = "/Development/Civ Related Projects/Prototype/C7"
+current_path = "/Development/Civ Related Projects/Prototype/C7/"
```
maxpetul: These variables shouldn't be modified as part of this PR, but really they shouldn't have been included in MainMenu.tscn in the first place. I tried it myself and found that they will get automatically reset to match the current system each time the game is launched. I think this started with the "Static variable in Util" PR that was merged recently. Could we avoid storing these, or at least move them to C7.ini?

C7GameData/ImportCiv3.cs
```diff
@@ -63,6 +63,7 @@ public static C7SaveFormat ImportSav(string savePath, string defaultBicPath)
                 c7Tile.riverSouthwest = civ3Tile.RiverSouthwest;
                 c7Tile.riverNorthwest = civ3Tile.RiverNorthwest;
                 c7Tile.Resource = resourcesByIndex[civ3Tile.ResourceID];
+				c7Tile.ResourceKey = resourcesByIndex[civ3Tile.ResourceID].Key;
```
maxpetul, line 65: This line doesn't match the others on my screen due to tabs vs spaces. We really ought to go through and tabify everything once and for all.

C7/MainMenu.tscn
QuintillusCFC: You're right, they shouldn't be there.  I have removed them in my new commit.  But they come back every time I start the program... they seem to be the path where the project lives (mine is what it states, except with C: in front).

What I'm not sure of is where they're coming from.  I don't see any code references that would generate them, nor do I see any changes in https://github.com/C7-Game/Prototype/pull/78/files that seem to explain it.

C7GameData/ImportCiv3.cs
QuintillusCFC: Spaces vs tabs :(.  I have now standardized the rest of this file on tabs.  I wonder if there's a good way to do it in one fell swoop without manually replacing them?  I could write a program to do that, but there's probably already a good way.

C7/MainMenu.tscn
maxpetul: I did some investigation and figured out where those variables are coming from. They belong to the Godot FileDialog node named "SetCiv3HomeDialog" that was added by the "Static variable in Util" PR. Have a look at the blame for MainMenu.tscn: https://github.com/C7-Game/Prototype/blame/Development/C7/MainMenu.tscn. You can find those variables in the editor if you look under the "Inspector" tab for that node. There's a little arrow next to each one that you can click to reset them to the current C7 project directory, and what must be happening is that Godot is resetting them each time the project is built.

I don't know what's the best way to fix this. We could create the file dialog at run time in C# code, that would work but it's not very elegant. Maybe there's a switch somewhere we could flip to stop Godot from resetting the vars. I search online briefly and didn't find anything.

C7GameData/ImportCiv3.cs
maxpetul: I don't know of an easy way to do it. I thought we only needed to change a handful of files, but searching the repo for strings of 8 spaces turned up almost 100 matching files, including all of the QueryCiv3 module.

C7/MainMenu.tscn
maxpetul: Anyway this isn't really related to the PR. I'll open an issue regarding these variables. The PR itself looks good.

maxpetul (approved):

C7/MainMenu.tscn
QuintillusCFC: Interesting.  I see what you mean now that I pull it up in Godot's editor.  I checked to see why the BIQ/SAV dialog doesn't have the same issue, and it's instantiated in code.  I agree with putting it in a separate issue, I'll add it to Carthage.

C7GameData/ImportCiv3.cs
QuintillusCFC: Yeah... there was no consistency on that early on.  It didn't help that when I thought I'd solved it in my VSCode settings, so all new code would be tabs, I didn't realize that the setting was only applied to the file I had been editing, so it continued to add spaces in other files for a while.

It looks like there's a somewhat okay to do this in VSCode (https://stackoverflow.com/questions/33974681/how-can-i-convert-tabs-to-spaces-and-vice-versa-in-an-existing-file).  I'm thinking we should have a "Carthage" branch off of Development, for any intermediate code that we do before Babylon releases (which won't be merged before then).  That could then be a good place to have one big commit to convert them.

I'm a little bit afraid of doing it before you merge your current PR though, in case it causes conflicts all over the place.

Sean-Brown:
> I wonder if there's a good way to do it in one fell swoop without manually replacing them?

Sounds like the job of a linter. On other projects I've used the `husky` package and added a pre-commit hook to run the linter, though that only works on staged files. I'm sure whatever linter package we'd choose would have a command line option for running that through all the files.

PR #147, opened 2022-03-14
Add credits file to display in the Credits scene

### LINKED ISSUE #145: Add credits text
Add a text file with contributors, licenses, etc to be displayed on the credits screen and bottom of the main menu

WildWeazel:
Raw contents of C7/credits.txt displayed in a simple Label. I was going to use RichTextLabel with BBcode but ran into formatting issues and just went with what worked. Exact text is TBD as we're awaiting a few more preferred names, and I'm not sure if there is any other software we need to credit.

### DIFF
--- a/.github/workflows/export-c7.yml
+++ b/.github/workflows/export-c7.yml
@@ -56,13 +56,13 @@ jobs:
         mkdir -v -p build/${FOLDER_NAME}
         cd ${EXPORT_NAME}
         godot -v --export${DEBUG_BUILD} "Windows Desktop" ../build/${FOLDER_NAME}/${EXPORT_NAME}.exe
-    - name: Copy Hard-Coded Save
+    - name: Copy Static Files
       run: |
-        if cp ${EXPORT_NAME}/c7-static-map-save.json build/${FOLDER_NAME}/
+        if cp ${EXPORT_NAME}/c7-static-map-save.json build/${FOLDER_NAME}/ && cp ${EXPORT_NAME}/credits.txt build/${FOLDER_NAME}/
         then
-          echo "successfully copied json save to build"
+          echo "successfully copied static files to build"
         else
-          echo "::warning title=Json Save Copy Failed::Failed to add the hard-coded save to the build"
+          echo "::warning title=Static File Copy Failed::Failed to add the static files to the build"
         fi
     - name: Upload Artifact
       uses: actions/upload-artifact@v2
@@ -102,13 +102,13 @@ jobs:
         mkdir -v -p build/${FOLDER_NAME}
         cd ${EXPORT_NAME}
         godot -v --export${DEBUG_BUILD} "Linux/X11" ../build/${FOLDER_NAME}/${EXPORT_NAME}.x86_64
-    - name: Copy Hard-Coded Save
+    - name: Copy Static Files
       run: |
-        if cp ${EXPORT_NAME}/c7-static-map-save.json build/${FOLDER_NAME}/
+        if cp ${EXPORT_NAME}/c7-static-map-save.json build/${FOLDER_NAME}/ && cp ${EXPORT_NAME}/credits.txt build/${FOLDER_NAME}/
         then
-          echo "successfully copied json save to build"
+          echo "successfully copied static files to build"
         else
-          echo "::warning title=Json Save Copy Failed::Failed to add the hard-coded save to the build"
+          echo "::warning title=Static File Copy Failed::Failed to add the static files to the build"
         fi
     - name: Create Tar Archive
       run: |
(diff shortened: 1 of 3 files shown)


QuintillusCFC:
Looks great when run from the editor.  When run from an exported version, it crashes due to being unable to find the file :(.

I looked at what JimOfLeisure did to get the static-json working, and he has a "Copy Hard-Coded Save" step in the build process (https://github.com/C7-Game/Prototype/blob/Development/.github/workflows/export-c7.yml) that does that.

Considering that we'll very soon have three resources to copy (static json, credits, background image), I think we're rapidly approaching where we should have a "resources" folder that contains all of our resources so we can just copy it over in one step.  I'm okay with band-aiding it for Babylon and having a Carthage task to make a more robust solution... but we need at least the band-aid or the credits won't be visible.

WildWeazel:
Agreed. I'll add error handling to this, which should be standard anyway. A unified resource packaging solution will mesh nicely with mod prototyping in Carthage.

QuintillusCFC:
That will fix the crash, but the credits still won't display to the end users in Babylon.  I'm okay with going ahead with it if you are, but if the point was to make them visible it is still defeated.

WildWeazel:
Right, I still need to modify the workflow to export it. And add any other licenses - we're using Blast, anything else? @JimOfLeisure @maxpetul

JimOfLeisure:
Blast is the only licensed thing I know of. There is a public domain OpenSimplex implementation in https://github.com/C7-Game/Prototype/blob/517f42ca14b24a1a3596db6994802cc1ff06f2ef/C7GameData/MapGenNoise.cs copied from https://gist.github.com/digitalshadow/134a3a02b67cecd72181 . Being public domain it doesn't require attribution, but FYI.

(That is not currently in use; it is there presuming we'll want to be able to generate maps independent of Godot code which has its own OpenSimplex implementation, such as a standalone game server.)

JimOfLeisure:
Oh, wait, there is actually MoonSharp and an INI reader library we're grabbing via NuGet. MoonSharp is not currently part of the distributable; it's in _Console/LuaCiv3 but will eventually find its way into C7 as the Lua interpreter. The INI reader is being used for unit .INI files and I think C7 config.

Off to look through the .csproj files and look up licenses...

JimOfLeisure:
MoonSharp license (unknown if standard): https://github.com/moonsharp-devs/moonsharp/blob/master/LICENSE

ini-parser license (MIT/expat): https://github.com/rickyah/ini-parser/blob/development/LICENSE

We're using ini-parser in the exported C7 now, but MoonSharp is only in one of the console apps so far.

I didn't find any other external libraries in the .csproj files.

WildWeazel:
I added all license/attribution text, and then went back to RichTextLabel because I needed it to scroll even though not all BBcode is working. Apparently the default font doesn't have styling and I couldn't get it to use our custom font, but that's polish we can add later.
Changed workflow step "Copy Hard-Coded Save" to "Copy Static Files" and added credits.txt.
Assuming the build works this should be ready to go now.

JimOfLeisure:
A follow-up on [the Noto_Sans font](https://fonts.google.com/noto/specimen/Noto+Sans/about) license: @WildWeazel and I had a conversation. The Noto Sans zip I have from 2021 (containing files from 2013) has an Apache 2.0 license included with no owner or copyright date included, and that is the attribution included in the repo at https://github.com/C7-Game/Prototype/blob/Development/C7/Fonts/LICENSE-NotoSans.txt . I was quite sure this was originally from Google Fonts, but the current Google Fonts have an OFL license.

After our discussion I am in agreement that the OFL attribution is proper to include in the credits as has been done.

I did additional follow-up by re-downloading the current Noto_Sans.zip from Google fonts, and it comes with an OFL license file. The file lengths and md5 signatures differ from what's in the repo, but I installed an ran [Google's fontdiffenator tool](https://github.com/googlefonts/fontdiffenator) , and there are **no** glyph differences between the current files and the files currently in the repo.

I will replace the Noto_Sans font file in the repo and update the license file. The newer ttf file includes Google copyright info in the header when viewed with `xxd` and the old one doesn't. And I'll do some brief tests and make a PR.

I like being absolutely certain when it comes to licensing!

Edit: Ref: #150

QuintillusCFC:
+1 for the YUMBO.

Is it intentional that the URLs show as underlined and with a hand cursor if you hover over them that makes it look like you can click them, but don't do anything when you click them?  It's no showstopper, but with the hand cursor I kind of thought it might open the link in my default browser or something.  As it is, it just teases me and makes me type it out if I do want to load the link.

Also, cross-ref #150 , the Noto Sans in the credits just says it's copyright Google, All Rights Reserved.  Doesn't mention the Open Font License.  Probably want to update that (whether in this PR or the other) since right now the credits make it look kind of like we just swiped a font from the Googleplex and haven't returned it yet.

+1 for the scrolling as well.

For Carthage we should probably also bold-ize or otherwise indicate the "sections"/components used to visually separate them.  It's a bit wall of texty without them.  Which is fine for now.

QuintillusCFC (approved):
But I see that it'll get exported now and generally looks fine.  Going to approve for now, also knowing that the other one can cover Noto Sans if need be.

WildWeazel:
Bold and url tags were two of the BBcode features I didn't get working. I think we have to handle an event and forward the url to the OS. Bold seems to be because of the default font, and I couldn't get it load Noto.
Jim and I tried to figure out how to apply those licenses, and eventually found that attribution on this page: https://fonts.google.com/attribution

QuintillusCFC:
Hmm... I see it on the link you just mentioned, but it does mention "SIL Open Font License, 1.1" way off to the right of the Noto Sans area.

Definitely okay with having the BBCode features come post-Babylon!

PR #151, opened 2022-03-18
Custom menu background for Babylon

QuintillusCFC:
This adds Babylon's main menu background.

The original image was roughly 5:4, and I trimmed it to roughly 4:3.  Given the source image, it is not feasible to trim it to 16:10 or 16:9.  However, since our launch resolution is 1024x768, it actually looks pretty good at launch.  It will also stretch if you resize it, which IMO is a nice touch even if it does mean the proportions are a bit off.

Actual image is 3840x2789, good for up to 4K.  Size is just under 3 MB, up from a bit over 512K for my prototype (heh) image, which was just over 1200p.  Source image on Wikipedia was 18 MB.


C7/MainMenu.cs
```diff
@@ -61,14 +61,20 @@ private void DisplayTitleScreen()
 		}
 		catch(Exception ex)
 		{
-			GD.Print("Could not set up the main menu: ", ex);
+			GD.Print("Could not set up the main menu", ex);
+			GetNode<Label>("CanvasLayer/Label").Visible = true;
+			GetNode<ColorRect>("CanvasLayer/ColorRect").Visible = true;
```
QuintillusCFC, line 66: I made these hidden by default, as part of the refactor, so they now have to become visible when we can't load the PCX files for things such as button graphics.

C7/MainMenu.cs
```diff
@@ -77,14 +83,14 @@ private void AddButton(string label, int verticalPosition, string actionName)
 		TextureButton newButton = new TextureButton();
 		newButton.TextureNormal = InactiveButton;
 		newButton.TextureHover = HoverButton;
-		newButton.SetPosition(new Vector2(835, verticalPosition));
+		newButton.SetPosition(new Vector2(40, verticalPosition));
```
QuintillusCFC, line 86: I moved these to the left side from the right (which was only chosen because it matched Civ III Vanilla).  This has the advantage of not appearing to jump around when you resize the screen.  It also looks better with Babylon's background (the clouds on the right make it hard to find a color that works well for the font on all buttons).

Long-term, we can anchor it better if we want it on the right for a future background.

Also long-term, the buttons/labels and their spacing don't scale with the screen size, whereas the background (including "C7 Babylon") does.  I don't think it's worth delaying Babylon over figuring out how to get these main menu labels scaling amazingly.

C7/MainMenuTheme.tres
```diff
@@ -0,0 +1,32 @@
+[gd_resource type="Theme" load_steps=6 format=2]
```
QuintillusCFC, line 1: I copied our main theme so I could change the colors on the buttons on the main menu, which shouldn't necessarily follow the main theme.  We probably could go back to the bad old days of font color overrides everywhere, and that would avoid duplicating this... but I'm not sure it's better.

C7/Title_Screen.jpg.import
```diff
@@ -0,0 +1,35 @@
+[remap]
```
QuintillusCFC, line 1: StackOverflow says we should commit these: https://stackoverflow.com/questions/54249305/godot-what-are-import-files-and-should-you-commit-them-to-git

QuintillusCFC:
TODO: Update export-c7.yml so the background gets exported.

I'm going to wait to do this until #147 is merged, because it will conflict horribly, and I'd essentially have to do it again (or make WildWeazel do it again).  I'll also go ahead and create an issue to make that problem not pop up every time we add a resource going forward.

Edit: Done!

WildWeazel:
Code looks good. I want to see it in action later.
When resized, most other scenes will be layered transparently over the game map. We may want to have a background texture to fill in behind these menu screens, perhaps a tileable C7 wallpaper that could be a general purpose fallback image.

WildWeazel (approved):

PR #160, opened 2022-03-24
Serialize Terrain Data

Sean-Brown:
Serialize the terrain key rather than the terrain object in the save file. When loading the save file, look up the terrain object by its key.


Sean-Brown:
I had to turn off automatic code formatting on saves, it'd be nice to have a formatting standard that VS Code could pick up on and automatically format the files to.
Re-wrote the static-map-save file json to include the terrain keys instead of the terrain json objects.
Also when I'd run the project, the IDE would insist on adding a `C7.ini`  file containing one line:
```[audio]
musicVolume = 100
```
but I wasn't sure if that should be included or not so I left it out of the PR.

Sean-Brown:
Also not sure who to add as reviewers, I'll add Quintillus for now

QuintillusCFC (approved):
Looks good, tested BIQ/SAV/JSON, and verified that the references to the terrain are already using the key.  Congratulations on reaching -40,000 net lines of code so quickly!

QuintillusCFC:
Re: the C7.ini, that should be on the .gitignore, but isn't.  It stores user preferences.  I'm a bit surprised we hadn't run across it trying to commit itself already.  If you want to add it to .gitignore before merging this one, go ahead, or we can add it later (but preferably sooner rather than later, or it will wind up committed).

Issue #123, opened 2022-03-05
Upper Left Buttons not Disabled while Main Menu is Open


maxpetul:
While in game, if you click the upper-left main menu button repeatedly without closing the menu, multiple main menus will be created and opened next to one another. Similarly you can click to open the advisor screen while the main menu is open.

WildWeazel:
I see that we're reimplementing Popup which I assume is for the custom background. Godot's Popup is modal by default with multiple display options, and may be better to inherit from than TextureRect.

QuintillusCFC:
Well, the _original_ reason we were "reimplementing" popup was that I didn't realize Godot had a built-in popup when I first decided to add a popup.  But yes, the challenge with moving to the standard one at this point would be the style, and that some of them (such as Build City) are semi-custom in functionality.

I haven't examined the main menu one up close yet.

esbudylin:
The issue seems to be fixed.

PR #161, opened 2022-03-24
Right click menus for units & city production

maxpetul:
Right clicking on a tile with units will now open a menu listing all those units (and only the units, the other menu items from the original game aren't implemented). Shift + right clicking on a city will open a menu listing all available builds in that city. Clicking the menu items works, either to wake & select a unit or to set city production.

The code for the menus is all in RightClickMenu.cs. The interface with the rest of the UI is pretty simple, f.e. opening a tile menu looks something like this: `new RightClickTileMenu(game, tile).Open(mouseLocation)`. The menu then captures all input and closes & deletes itself when appropriate. The way it captures all user input is to override the _Input method on Godot Nodes and call AcceptEvent to eat input events. I'm not sure if that's the best way to do it but it appears to work. It's different from how popups capture input, which is to use an overlay (I think?). I didn't use that approach since it seems more complicated and error prone, for example see issue #123.

The rest of the code is mostly engine changes to support the menu functionality. That includes a ListProductionOptions method for cities, a ChooseProduction message to set city production, and the FortifyUnit message becoming SetFortification with the ability to also wake units. Lastly I added a Describe method to MapUnit that produces the unit description that appears in the menu.


C7/UIElements/RightClickMenu.cs
```diff
@@ -0,0 +1,174 @@
...
+public class RightClickMenu : VBoxContainer
+{
+	protected Game game;
+
+	protected RightClickMenu(Game game) : base()
+	{
+		this.game = game;
+
+		// Set theme for menu node. TODO: This should be made moddable. I noticed in the Godot docs something about loading themes from files
+		// but didn't look into how it works, but that's probably what we'll want to do.
```
QuintillusCFC, line 14: Those would be the .tres files... we have a C7 theme and a main menu theme so far.  I didn't know you could load them programmatically though, I'd only set them per-scene.  The Godot editor is pretty decent for seeing how they look as you work on them.

But I agree, that can be done at some point in the future.  Feel free to write up an issue to research that so we remember it some day.

C7/UIElements/RightClickMenu.cs
```diff
@@ -0,0 +1,174 @@
...
+			CloseAndDelete();
+
+		// Eat all events other than mouse events while the cursor is over the menu. We want the menu to grab all input while it's open but we
+		// must make sure not to block mouse events from reaching its child buttons. (This had me confused for a while since the Godot docs
+		// say that events reach children before their parents, but the catch is that there are three phases of input processing. The "input"
+		// phase, this function, then "gui input", and finally "unhandled input". If a control eats an event during the "input" phase it won't
+		// proceed to the "gui input" phase where buttons actually respond to it.)
+		} else if (! ((@event is InputEventMouse) && mouseOverMenu))
+			this.AcceptEvent();
+	}
```
QuintillusCFC, line 105: Interesting... I'll have to refresh myself on the popup logic when I'm less tired and properly compare them, but I see the logic in the approach taken here.

C7/UIElements/RightClickMenu.cs
```diff
@@ -0,0 +1,174 @@
...
+		using (var gameDataAccess = new UIGameDataAccess()) {
+			MapUnit toSelect = gameDataAccess.gameData.mapUnits.Find(u => u.guid == guid);
+			if (toSelect != null && toSelect.owner == game.controller) {
+				game.setSelectedUnit(toSelect);
+				new MsgSetFortification(toSelect.guid, false).send();
+				ResetItems(toSelect.location, toSelect);
+			}
+		}
+		if (! Input.IsKeyPressed((int)Godot.KeyList.Shift))
+			CloseAndDelete();
```
QuintillusCFC, line 140: Whoa, I never knew that you could keep the menu open and de-fortify multiple units by holding down the Shift key!  I always thought it was either all of them or one at a time.  Probably wouldn't have realized it was possible in C7 either, if it weren't for reading the code...

C7/UIElements/RightClickMenu.cs
```diff
@@ -0,0 +1,174 @@
...
+	}
+}
+
+public class RightClickChooseProductionMenu : RightClickMenu
+{
+	private string cityGUID;
+
+	public ImageTexture GetProducibleIcon(IProducible producible)
```
QuintillusCFC: Minor: This method can be private.

QuintillusCFC (approved):
Looks good, and I discovered new Civ3 functionality in the process!

C7/Game.cs
```diff
@@ -412,7 +412,14 @@ public override void _UnhandledInput(InputEvent @event)
 				using (var gameDataAccess = new UIGameDataAccess()) {
 					var tile = mapView.tileOnScreenAt(gameDataAccess.gameData.map, eventMouseButton.Position);
 					if (tile != null) {
+						bool shiftDown = Input.IsKeyPressed((int)Godot.KeyList.Shift);
+						if (shiftDown && tile.cityAtTile != null && tile.cityAtTile.owner == controller)
```
Sean-Brown: I believe `tile.cityAtTile != null && tile.cityAtTile.owner == controller` can also be written like `tile.cityAtTile?.owner == controller`

maxpetul: It can be though they're not exactly the same. Rewriting it that way, it will pass if cityAtTile and controller are both null. That shouldn't matter right now since I don't think it's possible for controller to be null, but maybe some day it will be like for an observer player. Speaking of some day, we might want to add an empty city object and use that instead of null for cityAtTile. Anyway, in the mean time I'll replace it.

PR #165, opened 2022-03-26
114 - Lower right info box terrain updating

### LINKED ISSUE #114: Terrain type on Lower-Right Info Box Should Update
Right now, the lower-right info box always displays "Grassland" as the terrain type.  It should update dynamically now that we have a more full-fledged terrain system.

Could be Babylon, could be Carthage.  It helps round out the terrain feature but it's also not something I'd delay making a new release over.  Putting under the "World Map" feature since that's the closest thing.

QuintillusCFC:
This updates the lower-right info box to display the correct terrain.  I also added bombard stats (if applicable), and updating it when a unit moves.

Remaining caveat: It only updates after unit movement if the unit was not in combat.  The logical place to have it be updated would be in AnimationTracker.cs, in the endAnimation method, but that method is never called.  I tried to trace through the animation to figure out where the logical place for that method to be called would be, and it looks like the update method, line 105, or maybe 111, might be that area.  If we convert the unit guid ("id") to the unit, we could probably call endAnimation.

But I'm not very confident at all changing the unit animation code, and would thus like a second opinion before I go changing it.

### DIFF
--- a/C7/UIElements/GameStatus/LowerRightInfoBox.cs
+++ b/C7/UIElements/GameStatus/LowerRightInfoBox.cs
@@ -147,9 +147,9 @@ public class LowerRightInfoBox : TextureRect
 		string movementPointsRemaining = NewUnit.movementPointsRemaining > 0 ? "" + NewUnit.movementPointsRemaining : "0";
 		string bombardText = "";
 		if (NewUnit.unitType.bombard > 0)
-        {
+		{
 			bombardText = $"({NewUnit.unitType.bombard})";
-        }
+		}
 		attackDefenseMovement.Text = $"{NewUnit.unitType.attack}{bombardText}.{NewUnit.unitType.defense} {movementPointsRemaining}/{NewUnit.unitType.movement}";
 		attackDefenseMovement.Visible = true;
 	}
--- a/C7/UIElements/GameStatus/LowerRightInfoBox.cs
+++ b/C7/UIElements/GameStatus/LowerRightInfoBox.cs
@@ -138,13 +138,14 @@ public class LowerRightInfoBox : TextureRect
 		
 	}
 
-	public void UpdateUnitInfo(MapUnit NewUnit)
+	public void UpdateUnitInfo(MapUnit NewUnit, TerrainType terrain)
 	{
-		lblUnitSelected.Visible = true;
-		attackDefenseMovement.Visible = true;
+		terrainType.Text = terrain.DisplayName;
 		terrainType.Visible = true;
 		lblUnitSelected.Text = NewUnit.unitType.name;
+		lblUnitSelected.Visible = true;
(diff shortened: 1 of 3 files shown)


maxpetul:
The easiest thing would be to call the update method from Game._Process. That way the info box would always be up to date regardless of what happens with combat, etc., because it would get updated every frame. The only potential downside is that it's heavier in terms of performance, but I highly doubt that would matter.

About endAnimation, my plan for that method was that it would be called to stop repeating animations like workers doing their thing and maybe for cases like animations being interrupted by a unit's death or something. It's not called b/c it's not needed right now. All the animations are one-shots that get ended automatically and the unit layer just doesn't draw units with zero HP so their animation state doesn't matter. Eventually endAnimation will be needed.

C7/UIElements/GameStatus/LowerRightInfoBox.cs
```diff
@@ -138,13 +138,20 @@ private void toggleEndTurnButton()
...
 		terrainType.Visible = true;
 		lblUnitSelected.Text = NewUnit.unitType.name;
-		attackDefenseMovement.Text = NewUnit.unitType.attack + "." + NewUnit.unitType.defense + " " + NewUnit.movementPointsRemaining + "/" + NewUnit.unitType.movement;
+		lblUnitSelected.Visible = true;
+		string movementPointsRemaining = NewUnit.movementPointsRemaining > 0 ? "" + NewUnit.movementPointsRemaining : "0";
+		string bombardText = "";
+		if (NewUnit.unitType.bombard > 0)
+        {
+			bombardText = $"({NewUnit.unitType.bombard})";
+        }
```
maxpetul: These two lines with the braces have spaces instead of tabs.

maxpetul (approved):
Approved since it works and is an improvement. I don't know if you want to also deal with the updating after combat issue here or create another PR for that.

C7/UIElements/GameStatus/LowerRightInfoBox.cs
QuintillusCFC: Good catch, I'm pretty sure Visual Studio must still be inserting spaces.

QuintillusCFC:
> The easiest thing would be to call the update method from Game._Process. That way the info box would always be up to date regardless of what happens with combat, etc., because it would get updated every frame. The only potential downside is that it's heavier in terms of performance, but I highly doubt that would matter.
> 
> About endAnimation, my plan for that method was that it would be called to stop repeating animations like workers doing their thing and maybe for cases like animations being interrupted by a unit's death or something. It's not called b/c it's not needed right now. All the animations are one-shots that get ended automatically and the unit layer just doesn't draw units with zero HP so their animation state doesn't matter. Eventually endAnimation will be needed.

That probably would be the easiest, and for this particular change I don't think it would have a significant performance impact.  What concerns me is the precedent set by tossing re-calculations on Game._Process... if that becomes the default, a whole bunch of small impacts could add up to a quite substantial one.

Maybe stopAnimation would be a slightly better name?  I see the use case, I'd just thought from the name it was intended to be a hook for when the animation ended (in the onStart/onEnd style that many libraries use for hooks).  That can also be for the future though.

PR #167, opened 2022-03-28
Respond to arrow and fn key presses

### LINKED ISSUE #166: Allow units to move on laptop
Respond to key events that let units move as if a numpad were being used. Right now this works in Civ 3 by using the arrow keys and the function key in the following way:

- down -- numpad 2 -- down arrow
- up -- numpad 8 -- up arrow
- left -- numpad 4 -- left arrow
- right -- numpad 6 -- right arrow
- left-down -- numpad 1 -- function-right arrow
- left-up -- numpad 7 -- function-left arrow
- right-up -- numpad 9 -- function-up arrow
- right-down -- numpad 3 -- function-down arrow

This would help me test on my laptop since I don't have access to a numpad, left mouse clicking moves the map, and I'm used to using the function and arrow keys for moving units.

Sean-Brown:
This allows numpad-like movement on a laptop that lacks a numpad. It works the same as Civ 3 currently does.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -463,7 +463,7 @@ public class Game : Node2D
 				}
 			}
 			else if ((eventKey.Scancode >= (int)Godot.KeyList.Kp1) && (eventKey.Scancode <= (int)Godot.KeyList.Kp9))
-			{
+			{ // Move units with the numpad keys
 				if (CurrentlySelectedUnit != MapUnit.NONE)
 				{
 					TileDirection dir;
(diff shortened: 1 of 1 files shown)


C7/Game.cs
```diff
@@ -482,6 +482,25 @@ public override void _UnhandledInput(InputEvent @event)
 					UnitInteractions.moveUnit(CurrentlySelectedUnit.guid, dir);
 				}
 			}
+			else if ((eventKey.Scancode >= (int)Godot.KeyList.Home) && (eventKey.Scancode <= (int)Godot.KeyList.Pagedown))
+			{ // Move units with the arrow and fn keys
+				if (CurrentlySelectedUnit != MapUnit.NONE)
+				{
+					TileDirection dir;
+					switch (eventKey.Scancode) {
```
pcen, line 490: Can we get the direction from a map instead of a switch on the eventKey? It could be nice to have the key mappings less tightly coupled with the unit movement logic in case we want to change them or read them from a file in the future

WildWeazel: Key bindings and configuration could easily be a task of its own, but while we're on the topic it would be good to at least decouple the specific keys to somewhere that we can make configurable later.

pcen: Created https://github.com/C7-Game/Prototype/issues/168 to track this

Sean-Brown: I'll see about implementing that in this PR but if it ends up being too big a chance I'll leave it for that ticket

pcen: Since it's a relatively large feature it might be nice to merge it in a separate PR so it's easier to find feature changes in version control. Either way, feel free to assign yourself to the ticket if you decide to implement it

PR #169, opened 2022-03-29
Various small unit & animation things

### LINKED ISSUE #180: Unit Animation Refactors I

maxpetul:
This PR contains a handful of small changes I wanted to make to units and animations.
1. Move the call to UnitWithAvailableActions to Game.setSelectedUnit so that all newly selected units are guaranteed to have their actions set. This fixes a bug where the action buttons don't always appear for manually selected units, particularly on the first turn.
2. Delete the UnitInteractions methods that do nothing other than send a message. I don't like how it's not clear from their names that all they do is send messages, and calling them is not any shorter than writing out `new MsgWhatever(...).send()`.
3. Use MapUnit.move for AI settler movement so their animations play.
4. Enable the engine to control what happens to animations once they're finished. They can be deleted & forgotten, played again, or held on the last frame. This isn't needed right now but it eventually will be for worker animations.


QuintillusCFC (approved):
Looks good to me, and seems to work as intended.

PR #178, opened 2022-04-01
Structured Logging

### LINKED ISSUE #118: Add a proper logging framework
We have an increasing number of Console.WriteLine and GD.Print statements.  There are two downsides to this:

1. There's no granularity control.  No option to say, "print more stuff in debug mode".
2. There's no ability to send it all to a log file.  Having that ability will likely be useful for bug reports.

This is to figure out which framework to use and to start using it.  Known candidates include Log4Net, NLog, and Serilog.

pcen:
WIP implement https://github.com/C7-Game/Prototype/issues/118 by adding a logging framework

### DIFF
--- a/C7/LogManager.cs
+++ b/C7/LogManager.cs
@@ -1,0 +1,23 @@
+﻿using Serilog;
+using Serilog.Templates;
+using Serilog.Templates.Themes;
+
+public class LogManager
+{
+    static LogManager()
+    {
+		ExpressionTemplate consoleTemplate = new ExpressionTemplate(
+			"{@t:HH:mm:ss.fff} [{@l:u3}]{#if SourceContext is not null} {SourceContext}:{#end} {@m:lj}\n{#if @x is not null}\tException: {@x}{#end}",
+			theme: TemplateTheme.Code);
+
+		Log.Logger = new LoggerConfiguration()
+			.WriteTo.Console(formatter: consoleTemplate)
+			.MinimumLevel.Debug()
+			.CreateLogger();
+	}
+
+	public static ILogger ForContext<T>()
+    {
+		return Log.ForContext<T>();
+    }
+}
--- a/C7/MainMenu.cs
+++ b/C7/MainMenu.cs
@@ -1,11 +1,13 @@
 using Godot;
-using ConvertCiv3Media;
 using System;
+using Serilog;
 
 public class MainMenu : Node2D
 {
+	static ILogger log = LogManager.ForContext<MainMenu>();
+
 	readonly int BUTTON_LABEL_OFFSET = 4;
-	
+
 	ImageTexture InactiveButton;
 	ImageTexture HoverButton;
 	TextureRect MainMenuBackground;
@@ -18,7 +20,8 @@ public class MainMenu : Node2D
 	// Called when the node enters the scene tree for the first time.
 	public override void _Ready()
 	{
-		GD.Print("Hello world!");
+		log.Debug("enter MainMenu._Ready");
+
 		// To pass data between scenes, putting path string in a global singleton and reading it later in createGame
 		Global = GetNode<GlobalSingleton>("/root/GlobalSingleton");
 		Global.ResetLoadGamePath();
@@ -61,7 +64,7 @@ public class MainMenu : Node2D
 		}
 		catch(Exception ex)
 		{
-			GD.Print("Could not set up the main menu", ex);
+			log.Error(ex, "Could not set up the main menu");
 			GetNode<Label>("CanvasLayer/Label").Visible = true;
 			GetNode<ColorRect>("CanvasLayer/ColorRect").Visible = true;
 		}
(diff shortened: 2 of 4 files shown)


C7/LogManager.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Serilog;
+using Serilog.Templates;
+using Serilog.Templates.Themes;
+
+public class LogManager
+{
+    static LogManager()
```
pcen, line 7: for any runtime initialisation that must be done before some feature is used, AFAIK we don't have an "entry point" for C7, so I think using a static factory class to ensure the initialisation happens first is the best way to solve this

pcen:
I think the `LogManager` and configuration should belong to `C7Engine`. Along those lines, I think it would also be nice to avoid `using Serilog;` in all our files so we aren't tied to a specific framework. I feel like this should be easy to do since we only need to use the `Serilog.ILogger` interface, but I'm not sure how to/if it's possible to alias/re-export this in C#

WildWeazel:
I'll get to this as soon as I can get running again... my dev environment is currently FUBAR

C7/LogManager.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+		ExpressionTemplate consoleTemplate = new ExpressionTemplate(
+			"{@t:HH:mm:ss.fff} [{@l:u3}]{#if SourceContext is not null} {SourceContext}:{#end} {@m:lj}\n{#if @x is not null}\tException: {@x}{#end}",
```
QuintillusCFC, line 10: Can we put a sample statement above the format, or an explanation of the more complex elements ([{@1:u3}] and {@m:lj}, IMO) to make it easier to read what the format is trying to do?  Log formats are rarely intuitive and this one isn't too bad, but it's still handy to have the reference right there.

C7/MainMenu.cs
```diff
@@ -1,11 +1,13 @@
 using Godot;
-using ConvertCiv3Media;
 using System;
+using Serilog;
 
 public class MainMenu : Node2D
 {
+	static ILogger log = LogManager.ForContext<MainMenu>();
```
QuintillusCFC, line 7: Question: Is it possible to selectively enable logs?

For reference, I see it prints out:

```
12:37:32.429 [DBG] MainMenuMusicPlayer: setting volume to 100, which is 0 decibel offset
```

MainMenuMusicPlayer being the ForContext of that log.

I know some libraries, such as Log4J, have a hierarchical structure, where a fully-qualified class name is considered, such as com.civfanatics.xplatformeditor.map.rightSidePanel .  This then allows turning on debug messages based on partially-qualified class name, such as "com.civfanatics.xplatformeditor.map" to turn on debugging statements for the map, but not for the whole application.  I've only sparingly used that in practice, but it could be useful for debugging areas such as the graphics or AI, especially for hard-to-replicate events.

I see that we aren't using namespaces in the sample classes where it's enabled... so maybe the answer is that if we were it would be supported.

QuintillusCFC (commented):
See question.  One other question, do log statements look correct in the Godot command prompt for you, or do the newlines fail and right arrow characters get inserted?  It doesn't really work there for me on Windows (8.1; although it does work just fine in the "Output" area of Visual Studio).

I don't think that's a blocker, but it's probably worth documenting where to find them if someone is launching from the Godot editor and the output is less than stellar.

QuintillusCFC:
Adding @TinaFemea as a reviewer, as our resident .NET logging expert.

C7/MainMenu.cs
TinaFemea: I totally agree with this.  Having some way to show only the debug messages you're interested in is crucial.

TinaFemea:
> Along those lines, I think it would also be nice to avoid using Serilog; in all our files so we aren't tied to a specific framework. I feel like this should be easy to do since we only need to use the Serilog.ILogger interface, but I'm not sure how to/if it's possible to alias/re-export this in C#

I think you can do this buy just using the .net ILogger and telling it that Serilog is what you want to use as a logger.  I have not done this, but see:

https://stackoverflow.com/questions/61544047/use-serilog-with-microsoft-extensions-logging-ilogger
and https://stackoverflow.com/questions/61544047/use-serilog-with-microsoft-extensions-logging-ilogger, which is linked that response.

C7/MainMenu.cs
pcen: You can configure Serilog to filter by by namespace [like this](https://stackoverflow.com/a/38785632). With `ForContext`, the fully qualified class name is added to the log, but I believe it is just omitted from console output for brevity

WildWeazel:
Maybe I'm missing something obvious but do I need to DO anything to enable Serilog? I get console output and log files with our regular print statements but I don't see any of your formatted examples.

pcen:
> Maybe I'm missing something obvious but do I need to DO anything to enable Serilog? I get console output and log files with our regular print statements but I don't see any of your formatted examples.

I've only run on an intel mac, and I don't have to do anything to enable Serilog. Don't know if this makes a difference, but I always run it by invoking `Godot` on the project from the command line as opposed to using visual studio or Godot editor

WildWeazel:
Finally got back to this- and figured out how to run Godot scenes from the command line. I see the Serilog and regular Godot logs in the console, which makes sense given that your config has "WriteTo.Console". I can also configure Serilog to print to a file. I don't know how to combine Godot and Serilog output, and it doesn't look like many people are using them together. We may not need Godot's own Print but it would be good to capture output in the editor, and we definitely wouldn't want them going to different files. I'm not sure if that means configuring Serilog to pipe to Godot, or vice versa, maybe creating a [Godot sink](https://github.com/serilog/serilog/wiki/Provided-Sinks), or if we should keep looking elsewhere.

WildWeazel:
Sure enough, the suggestion was to create a sink to pipe logs to GD.Print. That's a pretty simple block of code, the catch being that the logging configuration then indirectly depends on Godot. I don't think that will be a problem. The core code is not a standalone executable, so doesn't need to configure its own logging, just expose the interface. The Godot and unit test projects can each then configure logs as needed.

WildWeazel:
going to recreate this on a local branch for further editing

PR #179, opened 2022-04-02
Add zip compression for save files

### LINKED ISSUE #33: Add optional gzip compression to C7SaveFormat.Save() and .Load()
Not necessarily an Aztec 0.0 target

pcen:
adds compression for https://github.com/C7-Game/Prototype/issues/33

zip results on `c7-static-map-save.json`
unzipped: 2047679
zipped: 35159

### DIFF
--- a/C7GameData/SaveFormat.cs
+++ b/C7GameData/SaveFormat.cs
@@ -1,114 +1,117 @@
 namespace C7GameData
-
 /*
-    The save format is intended to be serialized to JSON upon saving
-    and deserialized from JSON upon loading.
+	The save format is intended to be serialized to JSON upon saving
+	and deserialized from JSON upon loading.
 
-    The names are capitalized per C#, but I intend to use JsonSerializer
-    settings to use camel case instead, unless there is reason not to.
-
+	The names are capitalized per C#, but I intend to use JsonSerializer
+	settings to use camel case instead, unless there is reason not to.
 */
 {
-    using System.IO;
-    using System.IO.Compression;
-    using System.Text.Json;
+	using System.IO;
+	using System.IO.Compression;
+	using System.Text.Json;
 
-    public enum SaveFileFormat
-    {
-        Plain,
-        Zip,
-    }
+	public enum SaveFileFormat
+	{
+		Plain,
+		Zip,
+	}
 
-    public class C7SaveFormat
-    {
-        public string Version = "v0.0early-prototype";
+	public class C7SaveFormat
+	{
+		public string Version = "v0.0early-prototype";
 
-        // Rules is intended to be the analog to a BIC/X/Q
-        public C7RulesFormat Rules;
+		// Rules is intended to be the analog to a BIC/X/Q
+		public C7RulesFormat Rules;
 
-        // This naming is probably bad form, but it makes sense to me to name it as such here
-        public GameData GameData;
+		// This naming is probably bad form, but it makes sense to me to name it as such here
+		public GameData GameData;
 
-        public C7SaveFormat()
-        {
-	        GameData = new GameData();
-        }
+		public C7SaveFormat()
+		{
+			GameData = new GameData();
+		}
 
-        public C7SaveFormat(GameData gameData, C7RulesFormat rules = null)
-        {
-            GameData = gameData;
-            Rules = rules;
-        }
+		public C7SaveFormat(GameData gameData, C7RulesFormat rules = null)
+		{
+			GameData = gameData;
+			Rules = rules;
+		}
 
-        public static C7SaveFormat Load(string path, SaveFileFormat format = SaveFileFormat.Plain)
-        {
-            System.Console.WriteLine("loading...");
-            C7SaveFormat save = null;
-            if (format == SaveFileFormat.Plain)
-            {
-                string json = File.ReadAllText(path);
-                save = JsonSerializer.Deserialize<C7SaveFormat>(json, JsonOptions);
-            }
-            else
-            {
-                using (var fileStream = new FileStream(path, FileMode.Open))
-                using (var archive = new ZipArchive(fileStream, ZipArchiveMode.Read))
-                {
-                    ZipArchiveEntry entry = archive.GetEntry("save");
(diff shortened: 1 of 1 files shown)


pcen:
tested loading and saving locally, will add unit tests once we have a legitimate unit testing setup

C7GameData/SaveFormat.cs
```diff
@@ -10,27 +10,59 @@ namespace C7GameData
...
+            else
+            {
+                using (var fileStream = new FileStream(path, FileMode.Open))
+                using (var archive = new ZipArchive(fileStream, ZipArchiveMode.Read))
+                {
+                    ZipArchiveEntry entry = archive.GetEntry("save");
+                    using (Stream stream = entry.Open())
+                    {
+                        var task = JsonSerializer.DeserializeAsync<C7SaveFormat>(stream, JsonOptions).AsTask();
+                        save = task.GetAwaiter().GetResult();
```
QuintillusCFC: Is there any practical benefit to doing this async?  As I read it, the thread that spins off the async task on line 60 is just going to immediately pause on line 61 to wait for the result.  So I don't think there's any benefit to it being async?

(Disclaimer: Not a C# expert yet.  So maybe there is?)

C7GameData/SaveFormat.cs
```diff
@@ -44,11 +76,30 @@ public static C7SaveFormat Load(string path)
...
+            {
+                using (var zipStream = new MemoryStream())
+                {
+                    var archive = new ZipArchive(zipStream, ZipArchiveMode.Create);
+                    ZipArchiveEntry entry = archive.CreateEntry("save");
+                    using (Stream stream = entry.Open())
+                    {
+                        stream.Write(json, 0, json.Length);
+                    }
+                    archive.Dispose();
```
QuintillusCFC: It's _probably_ okay since you've tested it locally, but it looks suspicious that we're disposing of the ZipArchive before we write the data to disk.  Most likely it's probably in the MemoryStream already so it works... but if there isn't any benefit to doing this before writing to disk, I'd rather have code that doesn't look suspicious.

C7GameData/SaveFormat.cs
```diff
@@ -10,27 +10,59 @@ namespace C7GameData
...
+            C7SaveFormat save = null;
+            if (format == SaveFileFormat.Plain)
+            {
+                string json = File.ReadAllText(path);
+                save = JsonSerializer.Deserialize<C7SaveFormat>(json, JsonOptions);
+            }
+            else
+            {
+                using (var fileStream = new FileStream(path, FileMode.Open))
+                using (var archive = new ZipArchive(fileStream, ZipArchiveMode.Read))
```
QuintillusCFC: Does it cause a problem to inline the FileStream?  I don't know if that's idiomatic in C# like it is in Java, but given the amount of nested using statements it seems like it would read better.

QuintillusCFC (commented):
A few small suggestions/questions.  More broadly, there isn't a way to test this in game without rigging it up manually somehow, right?  Just want to make sure I'm not missing a way to easily try it out.

pcen:
> A few small suggestions/questions. More broadly, there isn't a way to test this in game without rigging it up manually somehow, right? Just want to make sure I'm not missing a way to easily try it out.

Yes to test I had to created a sandbox command line project

C7GameData/SaveFormat.cs
pcen: You're right in that there is no performance benefit since 61 blocks until the result is available. However, since the zip archive only gives a stream, this seemed better than reading from the stream into a buffer in memory and then doing the deserialisation on the buffer (and maybe this uses slightly less memory for really large saves? I'm also new to C#)

C7GameData/SaveFormat.cs
pcen: It's unintuitive but we need to dispose of the archive in order for the contents of the stream to be valid to write to file: https://stackoverflow.com/questions/12347775/ziparchive-creates-invalid-zip-file#12350106

C7GameData/SaveFormat.cs
pcen: I'm not sure if `FileStream.Dipose` gets called if it's inlined, but I'll double check

C7GameData/SaveFormat.cs
QuintillusCFC: I was thinking just the following:

```
var task = JsonSerializer.Deserialize<C7SaveFormat>(stream, JsonOptions);
```

They seem to be equivalent behind the scenes; from the parameter name I suspect it's a stream in both cases.

Good to know you're also new to C#!  For some reason I thought you were a C# developer; maybe we should start a list of who our C# experts are.

C7GameData/SaveFormat.cs
QuintillusCFC: Interesting.  Well, in that case, we'll have to keep it!  Maybe we should add a link to that answer in a comment above the line for the benefit of anyone investigating some future error who comes across the line; I agree it's unintuitive.

C7GameData/SaveFormat.cs
pcen: The interface is pretty bad, but in order for GetResult to block until the job is actually finished, the `AsTask()` is required, if my understanding of [this documentation](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#valid-consumption-patterns-for-valuetasks) is correct

edit I missread your comment - I didn't think it was possible to deserialise on the stream so I must have missed that, but it makes sense

C7GameData/SaveFormat.cs
pcen: looks like it might be idiomatic, seeing as `ZipArchive` supports this: [docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.ziparchive.dispose?view=net-6.0#:~:text=Unless%20you%20construct%20the%20object,available%20for%20subsequent%20write%20operations.)

QuintillusCFC (approved):
Looks good to me, thanks for making the changes... and for tabifying the file while you were at it.

PR #182, opened 2022-04-02
Make Engine Tests Runnable From CLI

pcen:
On mac / unix, running xunit tests from the command line via `dotnet test` fails because `Microsoft.VisualStudio.TestPlatform.ObjectModel` cannot be found. [This issue](https://github.com/microsoft/vstest/issues/2469#issuecomment-666998302) explains the issue, and the solution which is to add the necessary package as a reference explicitly.

Fixing this makes local testing possible without Visual Studio on mac and linux, and will help with setting up CI testing on GitHub.

### DIFF
--- a/EngineTests/EngineTests.csproj
+++ b/EngineTests/EngineTests.csproj
@@ -1,25 +1,29 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-    <PropertyGroup>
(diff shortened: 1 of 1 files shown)


QuintillusCFC (approved):
Makes sense to me, and it still works on Windows too.

PR #183, opened 2022-04-04
Move units with go to button

### LINKED ISSUE #164: Go To Button Should Let You Send Units Places
Virote_Considon noticed that NumLock is still needed to move units.  The core developers have probably become used to NumLock, but that is something that we shouldn't keep as a requirement forever.

We already have some buttons working (hold, wait, fortify, disband, build city), which can be used as references.

The unique thing about Go To is that the next click on the map should send the unit to that location.

We already have a BFSLandAlgorithm for figuring out the shortest path between two tiles on land; it is not going to be the forever algorithm, so use the PathingAlgorithm interface.  But it can be used to assign a path for units.

Stretch goal (split into a seperate PR if not easy) is that units moving multiple tiles will automatically do so at the appropriate place in the human's turn.

Ideally, it should also include a cursor to indicate which tile is selected, but in the interest of keeping the PR small, I suggest leaving that for a secondary PR.  Doubly so because the cursor should show how many turns it will take to get there, which might not be bad from a calculation standpoint with the PathingAlgorithm, but adds another graphical twist.

pcen:
Implements using the go to button to move units as per https://github.com/C7-Game/Prototype/issues/164. This PR does not visually indicate which tile the cursor is hovering over, or draw the path on the map.

There are some issues with the current implementation namely
- it's possible to click on destinations more than one turn away, and upon arriving the unit will attack if a unit is on the destination tile
- units always get to use their last movement point to move onto difficult terrain, and their movement points reset each round. I'm pretty sure this is not how civ 3 works for multi-turn moves
- when a unit continues moving along their path after the first turn their path is assigned, the engine does not notify the UI to centre the UI on the unit
- there is no way to cancel a units current path once it has been created (need to double check how this works in civ 3 but I think just making the unit active again should cancel their current path)
- `Game.inUnitGoToMode` is the boolean flag indicating if the player's most recent action was clicking the unit go to button / pressing G. this flag needs to be untoggled in a bunch of situations (the most obvious being the player presses esc to cancel selecting a go to tile) otherwise you will probably accidentally move a unit. I haven't checked all the possible situations in which civ 3 cancels go to, so I'm confident some are missing

I'm not sure where the best place to move units along their current path at the start of the player's turn is, and the best way to notify the UI to centre on these units, so I think we should save handling the edge cases and above issues for a future PR to improve on this one.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -251,6 +251,11 @@ public class Game : Node2D
 	{
 		unit = UnitInteractions.UnitWithAvailableActions(unit);
 
+		if ((unit.path?.PathLength() ?? -1) > 0) {
+			GD.Print("cancelling path for " + unit);
+			unit.path = TilePath.NONE;
+		}
+
 		this.CurrentlySelectedUnit = unit;
 		this.KeepCSUWhenFortified = unit.isFortified; // If fortified, make sure the autoselector doesn't immediately skip past the unit
 
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -30,7 +30,7 @@ public class Game : Node2D
 
 	// CurrentlySelectedUnit is a reference directly into the game state so be careful of race conditions. TODO: Consider storing a GUID instead.
 	public MapUnit CurrentlySelectedUnit = MapUnit.NONE;	//The selected unit.  May be changed by clicking on a unit or the next unit being auto-selected after orders are given for the current one.
-	private bool isUnitGoTo = false;
+	private bool inUnitGoToMode = false;
 
 	// Normally if the currently selected unit (CSU) becomes fortified, we advance to the next autoselected unit. If this flag is set, we won't do
 	// that. This is useful so that the unit autoselector can be prevented from interfering with the player selecting fortified units.
@@ -381,15 +381,14 @@ public class Game : Node2D
 				GetTree().SetInputAsHandled();
 				if(eventMouseButton.IsPressed())
 				{
-					if (isUnitGoTo) {
-						isUnitGoTo = false;
+					if (inUnitGoToMode) {
+						setGoToMode(false);
 						using (var gameDataAccess = new UIGameDataAccess()) {
 							var tile = mapView.tileOnScreenAt(gameDataAccess.gameData.map, eventMouseButton.Position);
 							if (tile != null) {
-								UnitInteractions.beginUnitMoveTo(CurrentlySelectedUnit.guid, tile);
+								UnitInteractions.setUnitPath(CurrentlySelectedUnit.guid, tile);
 							}
 						}
-						GD.Print("UNIT SHOULD GO HERE");
 					}
 					else
 					{
@@ -423,8 +422,9 @@ public class Game : Node2D
 				GetTree().SetInputAsHandled();
 				AdjustZoomSlider(-1, GetViewport().GetMousePosition());
 			}
-			else if ((eventMouseButton.ButtonIndex == (int)ButtonList.Right) && (! eventMouseButton.IsPressed()))
+			else if ((eventMouseButton.ButtonIndex == (int)ButtonList.Right) && (!eventMouseButton.IsPressed()))
 			{
+				setGoToMode(false);
 				using (var gameDataAccess = new UIGameDataAccess()) {
 					var tile = mapView.tileOnScreenAt(gameDataAccess.gameData.map, eventMouseButton.Position);
 					if (tile != null) {
@@ -521,13 +521,15 @@ public class Game : Node2D
 			}
 			else if (eventKey.Scancode == (int)Godot.KeyList.G && eventKey.Control)
 			{
-				mapView.gridLayer.visible = ! mapView.gridLayer.visible;
+				mapView.gridLayer.visible = !mapView.gridLayer.visible;
 			}
 			else if (eventKey.Scancode == (int)Godot.KeyList.Escape)
 			{
-				GD.Print("Got request for escape/quit");
-				PopupOverlay popupOverlay = GetNode<PopupOverlay>(PopupOverlay.NodePath);
-				popupOverlay.ShowPopup(new EscapeQuitPopup(), PopupOverlay.PopupCategory.Info);
+				if (!inUnitGoToMode) {
+					GD.Print("Got request for escape/quit");
+					PopupOverlay popupOverlay = GetNode<PopupOverlay>(PopupOverlay.NodePath);
+					popupOverlay.ShowPopup(new EscapeQuitPopup(), PopupOverlay.PopupCategory.Info);
+				}
 			}
 			else if (eventKey.Scancode == (int)Godot.KeyList.Z)
 			{
@@ -542,6 +544,11 @@ public class Game : Node2D
 					slider.Value = 0.5f;
 				}
 			}
+
+			// always turn off go to mode unless G key is pressed
+			// do this after processing esc key
+			setGoToMode(eventKey.Scancode == (int)Godot.KeyList.G);
+
 		}
 		else if (@event is InputEventMagnifyGesture magnifyGesture)
 		{
@@ -563,6 +570,11 @@ public class Game : Node2D
 		this.setSelectedUnit(UnitInteractions.getNextSelectedUnit(gameData));
 	}
 
+	private void setGoToMode(bool isOn)
+	{
+		inUnitGoToMode = isOn;
+	}
+
 	///This is our global handler for unit buttons being pressed.  Both the mouse clicks and
 	///the keyboard shortcuts should wind up here.
 	///Eventually, we should quite possibly put this somewhere other than Game.cs, or at
@@ -570,6 +582,10 @@ public class Game : Node2D
 	///more things going on before figuring out what the 'right' thing is, though.
 	private void UnitButtonPressed(string buttonName)
 	{
+		// this will detoggle goTo when clicking unit buttons
+		// other than goTo
+		setGoToMode(buttonName == "goTo");
+
 		GD.Print("The " + buttonName + " button was pressed");
 		switch (buttonName) {
 			case "hold":
@@ -602,7 +618,6 @@ public class Game : Node2D
 				break;
 
 			case "goTo":
-				isUnitGoTo = true;
 				break;
 
 			default:
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -386,7 +386,7 @@ public class Game : Node2D
 						using (var gameDataAccess = new UIGameDataAccess()) {
 							var tile = mapView.tileOnScreenAt(gameDataAccess.gameData.map, eventMouseButton.Position);
 							if (tile != null) {
-								UnitInteractions.setUnitPath(CurrentlySelectedUnit.guid, tile);
+								new MsgSetUnitPath(CurrentlySelectedUnit.guid, tile).send();
 							}
 						}
 					}
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -30,6 +30,7 @@ public class Game : Node2D
 
 	// CurrentlySelectedUnit is a reference directly into the game state so be careful of race conditions. TODO: Consider storing a GUID instead.
 	public MapUnit CurrentlySelectedUnit = MapUnit.NONE;	//The selected unit.  May be changed by clicking on a unit or the next unit being auto-selected after orders are given for the current one.
+	private bool isUnitGoTo = false;
 
 	// Normally if the currently selected unit (CSU) becomes fortified, we advance to the next autoselected unit. If this flag is set, we won't do
 	// that. This is useful so that the unit autoselector can be prevented from interfering with the player selecting fortified units.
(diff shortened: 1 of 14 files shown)


C7/Game.cs
```diff
@@ -547,43 +568,60 @@ private void GetNextAutoselectedUnit(GameData gameData)
 		this.setSelectedUnit(UnitInteractions.getNextSelectedUnit(gameData));
 	}
 
+	private void setGoToMode(bool isOn)
```
pcen, line 578: in the future we will need to update UI here too

C7Engine/AI/Pathing/BFSLandAlgorithm.cs
```diff
@@ -18,23 +18,31 @@ public class BFSLandAlgorithm : PathingAlgorithm
...
 				distances[tile] = 1;
 				predecessors[tile] = start;
 				tilesToVisit.Enqueue(tile);
 			}
-			
+
 			//The core BFS algorithm.
 			while (tilesToVisit.Count > 0) {
 				Tile current = tilesToVisit.Dequeue();
+				if (current == destination) {
```
pcen, line 42: @QuintillusCFC tagging you here because we might want to make this change to BFS independent of this PR since it would be nice to get in ASAP for other code using pathing

QuintillusCFC: What particular change does line 42 make?  I see how 21-23 covers the case where a player orders a unit to go to the tile it's already on, which wasn't something the AI will do.

It looks like line 42 is intended as a shortcut for the case where the current tile and the destination are the same - but not the start tile and destination?  Thinking through the code, is it correct that this case is going to be triggered only when the destination is the immediate neighbor of the start tile?  Otherwise, it would hit lines 51-53 (which we may be able to remove now?).

I'm too tired to look at the full PR tonight; I might also double-check if I missed something when transcribing this from my Algorithms textbook when less tired (would lines 42-44 make 51-53 no longer required?).

QuintillusCFC: Also for context, this is a temporary pathing algorithm chosen due to its simplicity, and the only other place it is used is the settler AI (well, and my just-opened PR will use it for the explorer AI).  So fixing bugs in the short term is good, but we shouldn't aim to highly optimize it.  Hence why I want to make sure I understand the goal.

QuintillusCFC:
> * units always get to use their last movement point to move onto difficult terrain, and their movement points reset each round. I'm pretty sure this is not how civ 3 works for multi-turn moves

The first quoted item _is_ how units behave in Civ3, as well as Civ4.  I know it's different in Civ6, can't remember if it's that way or not in Civ5.

It is sort of an exploit to get "free" movement like that, but for compatibility we need to support it.  It's also sort of weird in Civ6 how units can start the turn with fractional movement points.  A hazard of a turn-based game with fractional movement.

Once we implement more advance pathing algorithms, this quirk will be a "fun" complexifier, which will mean we can't just use off-the-shelf algorithms, but will have to modify them.

> * there is no way to cancel a units current path once it has been created (need to double check how this works in civ 3 but I think just making the unit active again should cancel their current path)

You are correct about cancelling unit orders by activating them in Civ3.  By comparison, Civ4 (and I believe all later iterations) do not automatically cancel orders when you activate (click on) a unit; instead there's a cancel button you can select after you activate them, but you can just click on them if you want to see where they're going and leave the order intact.  Personally, I like the Civ4 model here as it's handy if a save game has been sitting a while between sessions, and would like that as an option... but for compatibility the Civ3 behavior should be the first goal.

C7Engine/AI/Pathing/BFSLandAlgorithm.cs
pcen: without the change on line 42, pathing fails when the start and destination are adjacent tiles, since all of the adjacent tiles are put into `tilesToVisit`, but we only compare destination to the neighbours of each `tile` in `tileToVisit` when `!distances.Keys.Contains(tile)`, which means that we never actually compare start to neighbouring tiles because they are already in `distances` (I think other tiles are missed too, but I only noticed on neighbouring tiles)

pcen:
@QuintillusCFC unrelated to this PR specifically, but I think fixing the indentation of every file my change touches makes it unnecessarily difficult to review, so it might be nice to set up linting and fix all the indentation in one go, because the dif is confusing to look at imo (at least in the github ui)

C7Engine/EntryPoints/MessageToEngine.cs
```diff
@@ -1,149 +1,164 @@
...
-	public override void process()
+	public class MsgSetUnitPath : MessageToEngine
 	{
-		MapUnit unit = EngineStorage.gameData.mapUnits.Find(u => u.guid == unitGUID);
-		if (unit != null)
-			unit.move(dir);
-	}
-}
+		private string unitGUID;
+		private Tile dest;
```
maxpetul: Instead of storing a Tile reference in this message, you should store the tile's index or coordinates. My plan is that these messages will be passed back and forth between the host and clients in multiplayer, so they can't depend on any engine instance. That's why I've been using unit GUIDs instead of Unit object references.

Not really related to this PR, but eventually we should implement a common way for all game objects to be identified. Then we could use that to automatically serialize game object references in these messages (maybe, not sure it's worth the effort). We still need to work out how that identification will work. I don't like GUIDs for reasons I've mentioned before on the forums.

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -69,28 +69,33 @@ public static void EndTurn()
...
+				mapUnit.movementPointsRemaining = mapUnit.unitType.movement;
+			}
+			//Clear all wait queue, so if a player ended the turn without handling all waited units, they are selected
+			//at the same place in the order.  Confirmed this is what Civ3 does.
+			UnitInteractions.ClearWaitQueue();
+			gameData.turn++;
 
-	    new MsgStartTurn().send();
-        }
+			foreach (MapUnit mapUnit in gameData.mapUnits) {
```
QuintillusCFC: Because this is after the movement reset on lines 78-80, if you give a unit a go to order, it effectively gets double movement for the first turn of its path.

I think this should probably be at the start of the "EndTurn" method.  It should probably also use the gameData.players collection (which has a _units_ list of MapUnits), to only apply this to human players.  It works out to be the same right now, but if the AI starts giving units go to orders, that distinction will be useful.

(For full reference, in Civ3, units with go to orders are actually moved _before_ the turn ends.  All units without a go to order, and which need orders, are cycled through first.  Then units with Go To orders move.  Then, if any units received a "wait" order on their initial cycle, they are activated again.  At this point I don't think we need to perfectly emulate that, and there are times when I don't even like that methodology, particularly the go to happening before the "waited" units.  But at some point we should give that as an option at least)

QuintillusCFC (changes requested):
One comment about when go to orders are processed, which gives double-movement on the first turn of a go to order.

Overall I like how this looks.  It got me thinking about how it would probably make sense to combine the go to order with AI equivalents (e.g. settlers effectively get a "go to" order when they decide to settler cities.  But that only really makes sense to combine once both this PR and my recent AI refactors are merged.  Longer term, though, I am thinking a common "go to" AI used for both AI and human units probably makes sense (the explorer AI will also be an example of one used by both AI and human players, when the latter gives the Explore (Automated) order).

QuintillusCFC:
> @QuintillusCFC unrelated to this PR specifically, but I think fixing the indentation of every file my change touches makes it unnecessarily difficult to review, so it might be nice to set up linting and fix all the indentation in one go, because the dif is confusing to look at imo (at least in the github ui)

I agree, it turns out this review was about 1/4 of the size GitHub initially indicated, once I turned off whitespace changes.  I still cloned it locally so I could use an IDE to help reason about which methods called which others ones (and to try it out)... but the default GitHub view with whitespace changes didn't make things easier.

C7Engine/EntryPoints/TurnHandling.cs
pcen: yeah good catch, not sure how I didn't notice any extra movement when testing this. And I agree this could go in an AI. I need to play around with it in code, but I think most behaviour can and probably should be modelled that way since it will prevent pieces of engine code from getting coupled together

maxpetul (changes requested):

C7Engine/EntryPoints/MessageToEngine.cs
pcen: Makes sense thank you for pointing this out!

I haven't read forum discussions on GUIDs, but I am definitely in favour of not using GUIDs. I'm a fan of auto-incrementing integer IDs. They could also be scoped by an accompanying class ID to simplify generating unique IDs for each object. Do we have a thread specifically discussing this? Whatever we use it's probably something worth figuring out early on

maxpetul: I agree that auto-incrementing integer IDs, i.e. serial numbers, are the way to go. There's no thread about this on the forums, it was discussed briefly in the architecture thread starting with my post here: https://forums.civfanatics.com/threads/dev-architecture-thread.674254/#post-16218946. We ought to start such a thread to get this settled once and for all. The way I see it, the hard part about identifying objects is how those IDs will work with mods. In particular, mods should be able to define their own types of game objects, which will be given serial numbers and a class ID, and players should be able to add or remove mods from a saved game and have it still work, within reason. It will require some care to make sure we can still reliably identify objects as the ground moves beneath us.

pcen: I'll start a [discussion](https://github.com/C7-Game/Prototype/discussions) about it so we can also try out GitHub discussions. I'd like to merge this PR without adding Tile GUIDs, and refactor the GoTo functionality to use whatever we settle on, since AFAIK tiles don't have GUIDs currently.

pcen:
> One comment about when go to orders are processed, which gives double-movement on the first turn of a go to order.
> 
> Overall I like how this looks. It got me thinking about how it would probably make sense to combine the go to order with AI equivalents (e.g. settlers effectively get a "go to" order when they decide to settler cities. But that only really makes sense to combine once both this PR and my recent AI refactors are merged. Longer term, though, I am thinking a common "go to" AI used for both AI and human units probably makes sense (the explorer AI will also be an example of one used by both AI and human players, when the latter gives the Explore (Automated) order).

I'm pretty sure the functionality matches civ 3 now. Unfortunately, I still can't run civ 3, so I can't actually validate this myself. I'm still generally confused about what the "game loop" is (my understanding is _Process generates events that the engine processes), and when we enter `EndTurn`, this signifies the player is done interacting with units/cities etc. for that turn, and now the engine may act on the current game state to finish the player's turn and then process AI. But I'm not sure if just shoeing everything into `EndTurn` makes sense or not long term.

C7Engine/EntryPoints/MessageToEngine.cs
maxpetul: Tiles don't have GUIDs, that's why I suggested using the tile index or coordinates. But it's okay if you want to merge this PR without removing the object references since they will only be a problem for multiplayer, which is still a ways off.

C7Engine/EntryPoints/UnitInteractions.cs
```diff
@@ -1,98 +1,98 @@
...
+		{
+			foreach (Player player in gameData.players) {
+				//TODO: Should pass in a player GUID instead of checking for human
+				//This current limits us to one human player, although it's better
+				//than the old limit of one non-barbarian player.
+				if (player.isHuman) {
+					foreach (MapUnit unit in player.units) {
+						if (unit.movementPointsRemaining > 0 && !unit.IsBusy()) {
+							if (!waitQueue.Contains(unit)) {
+								return UnitWithAvailableActions(unit);
```
QuintillusCFC: Any particular reason we're switching this back?  Flintlock changed it last week at https://github.com/C7-Game/Prototype/pull/169/files#diff-3309f3270d6b4945af6fac63b05a6230c15e3167db67577293ed3dbfe72f1f84R22, so that Game.getSelectedUnit grabs the actions, fixing a bug described on that PR.  I don't think it necessarily harms anything to add it here, but I don't know what the benefit would be.

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -4,62 +4,72 @@
...
-                    Console.WriteLine("New barbarian added at " + tile);
-                }
-                else if (tile.NeighborsWater() && result < 10) {
-                    MapUnit newUnit = new MapUnit();
-                    newUnit.location = tile;
-                    newUnit.owner = gameData.players[0];    //todo: make this reliably point to the barbs
-                    newUnit.unitType = gameData.unitPrototypes["Galley"];
-                    newUnit.hitPointsRemaining = 3;
-                    newUnit.isFortified = true; //todo: hack for unit selection
+			foreach (Player player in gameData.players) {
```
QuintillusCFC: Two small things, one, it should probably be only human players.  But I'm also thinking maybe this should be in an appropriately named helper method.  My first thought looking at it after two days was, "why isn't this in PlayerAI.PlayTurn", before remembering it only applies to humans.

QuintillusCFC (commented):
Two comments, I think one reversion might have been introduced by rebase issues.

Looking forward to when this is merged, I got spoiled by the go to button when I tried it out.

C7Engine/EntryPoints/UnitInteractions.cs
pcen: Nope that's a bad rebase on my part 😦 I'll double check the diff to make sure nothing else is changed

C7Engine/EntryPoints/TurnHandling.cs
pcen: makes sense, I'll add a helper to get human players and iterate on those instead

QuintillusCFC (approved):

PR #185, opened 2022-04-05
104 - Basic AI Tile Visibility

### LINKED ISSUE #104: Tile visibility - AI keeps track of which tiles it knows about
The AI should keep track of which tiles it knows about.  This will enable several features:

 - Explorer AI will only explore places the AI doesn't know about
 - Settler AI will only settle where the AI has explored... unless the AI has done a poor job exploring and they just go in a random direction and hope for the best.
 - Defensive AI may eventually be able to defend chokepoints/resources only once they are discovered

It's worth noting that while we have some decent indications that the Civ AI cheats in this regard, and simulates exploration, that doesn't mean we also have to make the AI cheat.  And even if it does cheat, it will be easier to simulate exploration with this data.

QuintillusCFC:
Changes:

- AI stores info about which tiles it knows (does not yet import this from Civ3)
- Explorer AI now seeks to explore new tiles, rather than moving randomly

This doesn't accomplish all of the goals of #104 , in addition to not importing Civ3 data, the settler AI doesn't use tile knowledge, and the AI is a little too persistent about discovering tiles that have opposing cities on them.  But at 200 lines net, I want to at least have a draft PR in case adding more pushes it too high, and will likely split those out.


C7Engine/AI/ExplorerAI.cs
```diff
@@ -0,0 +1,131 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using C7GameData;
+using C7GameData.AIData;
+using C7Engine.Pathing;
+
+namespace C7Engine
+{
+	public class ExplorerAI
```
pcen, line 10: It could be nice to have an interface for both AI and for AI data. Partially because it would be nice to treat AI polymorphically, but also because the current naming convention is slightly confusing for someone who isn't familiar with the code ie. we have `ExplorerAIData : UnitAI`, but I don't really expect implementations of `UnitAI` to contain data

QuintillusCFC: Yeah, I was thinking about that when I fell asleep last night and forgot about it when I woke up.  UnitAI is definitely confusing, I will rename that.  I should be able to get the AI to be polymorphic as well.

QuintillusCFC: I wound up spinning this (other than the UnitAI data rename) off into https://github.com/C7-Game/Prototype/pull/203, since it's a semi-decent chunk of changes and isn't really related to tile visibility/the initial explorer AI.  It is a nice improvement though, glad you mentioned it.

C7GameData/AIData/TileKnowledge.cs
```diff
@@ -0,0 +1,34 @@
...
+
+		public bool isTileKnown(Tile t) {
+			return knownTiles.Contains(t);
+		}
+
+		/**
+		 * Returns a copy of the list of known tiles.
+		 * This prevents external modifications.
+		 **/
+		public List<Tile> AllKnownTiles()
```
pcen, line 25: is there a particular reason for returning the copy as a list?

pcen (approved):
Since this is in conjunction with https://github.com/C7-Game/Prototype/pull/203, I'm just reviewing the AI logic, which looks good to me. I like how there's a comment for all the "temporary" logic, it will make it easy to iterate over in the future

C7GameData/AIData/TileKnowledge.cs
QuintillusCFC: Potentially excessive caution around the data being modified outside of the TileKnowledge API.

The gist of the thinking was that there are places where it's handy for the AI to have a list of all known tiles - for example, deciding where to send a settler.  But it should never be possible to remove items from that list, and if you send the whole list out, it is.

In Java, I would have done this via the [Collections.unmodifiableList](https://docs.oracle.com/javase/1.5.0/docs/api/java/util/Collections.html#unmodifiableList(java.util.List)) method, which returns a read-only view of the list.  This is lighter weight than making a whole copy, but I'm not aware of an equivalent in C#.

pcen: Yeah it makes senes to to return a copy, I was just curious if there's a reason why the "real" known tiles is in a map, but the AllKnownTiles returns a copy as a list, as opposed to a copy as a map.

QuintillusCFC: Oh, right.

The internal collection is a Set so that if a tile is added more than once via _AddTilesToKnown_, it still only winds up there once.  My intention with _AddTilesToKnown_ is that by passing it the tile a unit is on, it can automatically handle logic around how many neighboring tiles should be added, including based on terrain height (hills, mountains, etc.).  A set seemed like the simplest way to avoid adding duplicates.

For consumption purposes, a list just seemed like the default collection.  A set would work just as well, though.

I'm guessing you meant "set" instead of "map" in both cases?

QuintillusCFC: _Long_ term, we'll also need more methods to deal with tiles revealed by air missions... but that can probably just be a new method similar to AddTilesToKnown, but that doesn't take into account ground terrain, and has a wider radius.  I think it will be compatible, but am not too concerned about it now.

PR #199, opened 2022-04-05
Add .editorconfig

### LINKED ISSUE #163: Automatic code formatting plugins for VSCode + Rider
As mentioned by @Sean-Brown , and previously at CFC, we should have code formatting happen automagically, especially as the project and the potential for inconsistency grows.

Ideally, this will be something that can be used across IDEs.  I don't know if that's feasible in the .NET world though.  If you are a .NET person and know a good solution, you can take this issue and run with it!

Also of note is that we haven't settled on many code styles yet, except for using tabs for indentation rather than spaces.  There are definite inconsistencies, so a plugin that fixes everything in the project at once will result in a huge change set.  It might be better to have one that applies when files are saved.  But taking this issue might also mean herding some cats to agree on currently diverging standards.

pcen:
adds a `.editorconfig`

configures:
- new line rules based on the [google C# style guide](https://google.github.io/styleguide/csharp-style.html#whitespace-rules)
  - rational: most familiar for developers coming from Java / C++ and shortens files
- indentation is tabs
  - rational: this has already been picked as the standard moving forward
- use `var` when type is clear (ie. a new statement)
  - rational: somewhat arbitrary but some code already uses var, and this way types are still explicit when you can't see the type on the same line

does not configure:
- naming conventions
  - why not: the code is inconsistent and renaming a bunch of variables right now is more likely to break stuff

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,10 +29,10 @@ csharp_new_line_before_members_in_anonymous_types = true
 csharp_new_line_between_query_expression_clauses = true
 
 # indentation
-# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0049
+# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#indentation-options
 csharp_indent_case_contents = true
 csharp_indent_switch_labels = true
-csharp_indent_labels = one_less_than_current
+csharp_indent_labels = flush_left
 csharp_indent_block_contents = true
 csharp_indent_braces = false
 csharp_indent_case_contents_when_block = true
@@ -41,7 +41,7 @@ csharp_indent_case_contents_when_block = true
 # https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008
 # https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0049
 csharp_style_var_for_built_in_types = false:warning
-csharp_style_var_when_type_is_apparent = true:warning
+csharp_style_var_when_type_is_apparent = false:warning
 csharp_style_var_elsewhere = false:warning
 dotnet_style_predefined_type_for_locals_parameters_members = true:warning
 dotnet_style_predefined_type_for_member_access = true:warning
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,71 @@
+root = true
+
+# Defaults
+
+[*]
+charset = utf-8
+insert_final_newline = true
+indent_style = tab
+indent_size = 4
+trim_trailing_whitespace = true
+end_of_line = lf
+
+
+# C#
+
+[*.cs]
+
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:warning
+dotnet_sort_system_directives_first = true
+
+# new lines
+# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#new-line-options
+csharp_new_line_before_open_brace = none
+csharp_new_line_before_else = false
+csharp_new_line_before_catch = false
+csharp_new_line_before_finally = false
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# indentation
+# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0049
+csharp_indent_case_contents = true
+csharp_indent_switch_labels = true
+csharp_indent_labels = one_less_than_current
+csharp_indent_block_contents = true
+csharp_indent_braces = false
+csharp_indent_case_contents_when_block = true
+
+# type rules
+# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008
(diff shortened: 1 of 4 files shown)


Sean-Brown (approved):
Nice PR, looks thorough. My personal preference is newlines before open brace, before else, before catch/finally, but not a big deal.

QuintillusCFC:
This will definitely be helpful, especially since some IDEs are opinionated, and don't always agree with each other - e.g. Visual Studio has taken it upon itself to format files based on its default preferences.

The reason some code uses var, and others don't, is that we don't have agreement on that, so some developers have been using var and others have been avoiding it.

I prefer to always use the explicit type, for clarity.  But of the three var-related rules, the one I really care about is _csharp_style_var_elsewhere_ being false, which it is.  

I suppose for the other two var rules, the reason I prefer the specific type is that if I'm not sure what type a variable is and I control+click on it to see where it's declared, I tend to naturally move my eyes to the left side where I expect the type to be.  If it's not there, it's another step to see if it's present on the right.  If the type _is_ there, it's not horrible, but if it's a method return and it's var (_csharp_style_var_elsewhere_ being true), it's even more work to figure out what type it is.  The effort to figure out the type when debugging/reading others' code - or my own previous code - is not worth the savings of typing "var" instead of the real type, IMO.

Hence, I'd say I have a preference for always using the explicit type (the opposite of what is configured), and a strong preference for the explicit type with _csharp_style_var_elsewhere_ (which is the way it is configured).

Still, looking at the 221 instances where there is a "var", a fair amount of them would be changed to clearer types with this change, so I approve of making things better.

I prefer the braces to be on the same line as the try/else/if/etc., but at one point I preferred the opposite.  I could survive either way on that one.

.editorconfig
```diff
@@ -0,0 +1,71 @@
...
+csharp_new_line_before_open_brace = none
+csharp_new_line_before_else = false
+csharp_new_line_before_catch = false
+csharp_new_line_before_finally = false
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
+# indentation
+# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0049
```
QuintillusCFC: This should link to https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#indentation-options .  0049 is "Use language keywords instead of framework type names for type references"

QuintillusCFC:
_Minor_ preferences, after looking through the rules:

 - csharp_new_line_before_else/catch/finally: Slight preference for _false_.  Don't have a great reason for it though, probably just what was the default in Java back in the day.
 - csharp_indent_switch_labels: Preference for _flush_left_.  I've only ever seen it done that way, whether it's in C or BASIC, and it [seems to be standard](https://softwareengineering.stackexchange.com/questions/151806/indenting-labels-in-c) for at least C and FORTRAN.  We probably won't have a lot of these though so it's quite minor.

Looked through all of them and all of the rest I either approve, or for a few are indifferent to.

QuintillusCFC (approved):

pcen:
> _Minor_ preferences, after looking through the rules:
> 
> * csharp_new_line_before_else/catch/finally: Slight preference for _false_.  Don't have a great reason for it though, probably just what was the default in Java back in the day.
> * csharp_indent_switch_labels: Preference for _flush_left_.  I've only ever seen it done that way, whether it's in C or BASIC, and it [seems to be standard](https://softwareengineering.stackexchange.com/questions/151806/indenting-labels-in-c) for at least C and FORTRAN.  We probably won't have a lot of these though so it's quite minor.
> 
> Looked through all of them and all of the rest I either approve, or for a few are indifferent to.

The link you shared is for goto labels, so I assume you're referring csharp_indent_labels as opposed to switch labels?

.editorconfig
pcen: Thanks!

QuintillusCFC:
> > _Minor_ preferences, after looking through the rules:
> > 
> > * csharp_new_line_before_else/catch/finally: Slight preference for _false_.  Don't have a great reason for it though, probably just what was the default in Java back in the day.
> > * csharp_indent_switch_labels: Preference for _flush_left_.  I've only ever seen it done that way, whether it's in C or BASIC, and it [seems to be standard](https://softwareengineering.stackexchange.com/questions/151806/indenting-labels-in-c) for at least C and FORTRAN.  We probably won't have a lot of these though so it's quite minor.
> > 
> > Looked through all of them and all of the rest I either approve, or for a few are indifferent to.
> 
> The link you shared is for goto labels, so I assume you're referring csharp_indent_labels as opposed to switch labels?

You are correct.  Not sure how the "switch" got in there, must have been a neuron firing in the wrong direction.

QuintillusCFC:
Is there any reason not to merge this now?  I'm vaguely aware of Rider's defaults not entirely lining up with the .editorconfig preferences, and Visual Studio's defaults also don't match up fully (but in a different way), so every time I add code there are likely minor differences from the .editorconfig style being added.

pcen:
> Is there any reason not to merge this now? I'm vaguely aware of Rider's defaults not entirely lining up with the .editorconfig preferences, and Visual Studio's defaults also don't match up fully (but in a different way), so every time I add code there are likely minor differences from the .editorconfig style being added.

I was waiting in case anyone had any strong opinions to voice about style, but it doesn't seem like it so I think we can merge. I'm only using VS Code, so I'm not sure what those differences are, but we can refine the .editorconfig over time. I think the main one is probably variable / method naming conventions. Currently VS Code complains when private methods / members aren't capitalised, but I think in visual studio they're supposed to not be.

PR #200, opened 2022-04-06
Work-around to loading images via res://SomePath.jpg not working in release mode

### LINKED ISSUE #158: Load images as images instead of image files, so release builds of Godot work
Best explained by a screenshot of command.com:

image

The relevant code is in the SetMainMenuBackground of MainMenu.cs.

This can only be tested by making a release build with GitHub Actions (although that does not imply making an actual release).

QuintillusCFC:
Unfortunately, I couldn't find a good way to keep that working in debug mode; most didn't work, and the rest will be deprecated with Godot 4.0 and added complexity.

So instead, I added a utility method that will use the current method in debug mode, and use a release-appropriate method.  That latter method was discovered in a comment in the Godot Mono source code; unfortunately I closed the file before noting which one it was.

Also, commit a change to the VSStub.csproj project that enables building it in release mode (see answer #2 at https://stackoverflow.com/questions/15134384/the-outputpath-property-is-not-set-for-project).  This might fix the issues that WildWeazel and others have had with VSStub.

### DIFF
--- a/C7/MainMenu.cs
+++ b/C7/MainMenu.cs
@@ -69,10 +69,7 @@ public class MainMenu : Node2D
 
 	private void SetMainMenuBackground()
 	{
-		ImageTexture TitleScreenTexture = Util.LoadTextureFromPCX("Art/title.pcx");
-		Image backgroundImage = new Image();
-		backgroundImage.Load("res://Title_Screen.jpg");
-		TitleScreenTexture.CreateFromImage(backgroundImage);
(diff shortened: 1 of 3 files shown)


QuintillusCFC:
@WildWeazel Curious if the VSStub change fixes the issues you were seeing with it.  If your system was trying to build in Release mode, there's a decent chances it does.  If not... I don't have another guess yet.

Sean-Brown (approved):
lgtm

PR #203, opened 2022-04-06
104.5 Unit AI Refactor

QuintillusCFC:
Refactor the unit AI to use polymorphic interfaces.  This cleans up the code, and is a step towards being able to more easily plug in future unit AIs (although the interface should not be considered 100% final forever, it just happens to have been sufficient for what I've added so far).


pcen (approved):
looks good!

PR #204, opened 2022-04-06
Save File I/O Tests

### LINKED ISSUE #154: Add regression tests for being able to open the main types of files (SAV, BIQ), or at least to fail acceptably
Although we're still getting C#/Godot testing sorted, we do have at least one test.  So, it's time to start prioritizing the highest-priority stuff.

One of those is making sure we don't regress on being able to open the big categories of files:

 - SAV files
 - BIQ files
 - Subtypes of those, mainly:
   - BIQ files with/without custom rules
   - BIQ files with/without custom player data
   - BIQ files with/without random maps
   - SAVs based on BIQs in those "without" categories
 - The default C7 static-json file

Note that some of these aren't fully supported yet, e.g. BIQs without custom player rules.  In those cases, we should fail gracefully, rather than crash.

The exact form of these tests could vary, although I am thinking an integration test is more appropriate than a unit test in this case (we should also have more unit tests, but that's another story).  I'll create a separate issue for integrating these tests into GitHub workflows.

### LINKED ISSUE #155: Integrate regression tests into GitHub workflows/actions
See #154 for the base card.

Once we have some regression tests, they will be much more useful if they're integrated to run automatically instead of just sitting there not being run.

My inclination is that they should be run for every pull request, and quite possibly for commits to the Development branch (i.e. merges).

pcen:
Adds integration tests for `C7Engine.SaveManager`


pcen:
currently I can't run civ 3, so if someone could add a sav and biq format that would be great

WildWeazel:
I guess this got overlooked. What do you need, just a random SAV and BIQ added to the project?

pcen:
Yep, any random SAV and BIQ is fine (or a way to run civ 3 on a MacBook)

WildWeazel:
@QuintillusCFC do you have a "canonical" test file of either format, or suggestions for making one?

QuintillusCFC:
The closest I have are the BIQ files listed at https://hg.sr.ht/~adj/civ3_cross_platform_editor/browse/src/test/java/com/civfanatics/civ3/xplatformeditor?rev=tip .  The "Custom All.biq" through "Custom Rules and Plyr Data.biq" should cover all the standard expected things for all variants of parts of the BIQ being enabled and not enabled.

Many of the rest (the scenarios) are there primarily for regression testing across a wider range of input scenarios.  Some, e.g. the Marla Singer's World Map 1.29f.bic, and Euro.bic, provide coverage for older versions of the format, which I don't believe we have any intention of supporting in C7 (it would require customizations for importing them).

I don't have an equivalent for SAVs.  For a variety of sample inputs, one good option would be the GOTM Results at http://gotm.civfanatics.net/results/index.php (make sure you get a Conquests one, the regular GOTM is PTW).  I have a submission from COTM 157 (http://gotm.civfanatics.net/download.php?file=Quintillus_COTM157_C3C122_01.SAV) that ought to work well enough.

QuintillusCFC:
As a side note, the reason that I don't consider my collection "canonical" is that I haven't been thorough about making sure that all the sections of the BIQ that _may_ contain items (but are not required to) actually contain elements.  In practice I believe that most do, but there may be a couple edge cases which don't contain items by default lurking that aren't covered by my automated tests and their sample files.

WildWeazel:
Sounds good for now at least!

WildWeazel:
oops, accidentally opened this for review. Is anyone still working on it? It's failing CI tests.

QuintillusCFC:
Other than merging in master to resolve conflicts and seeing it failed, I haven't been, and the rumored return of pcen doesn't appear to have happened yet.  I did see it failed the tests, but didn't have the energy to debug it at the time.

pcen:
we can close this draft, I've made much more progress towards a viable save format / system here https://github.com/C7-Game/Prototype/pull/413 and I don't think it's worth investing in github actions for the godot 3.x branch when ultimately we will want a godot 4 docker image

PR #212, opened 2022-04-08
Right Click Menu Wake / Fortify All Options

### LINKED ISSUE #188: Right-click menu: Wake/Fortify all
An enhancement to the right-click menu on tiles: being able to wake up or fortify all units at once.  Should only be able to do this on tile with multiple units.

pcen:
Adds Fortify All and Wake All options when right clicking on a unit stack:
Screen Shot 2022-04-07 at 11 08 37 PM

### DIFF
--- a/C7/UIElements/RightClickMenu.cs
+++ b/C7/UIElements/RightClickMenu.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using Godot;
 using C7GameData;
 using C7Engine;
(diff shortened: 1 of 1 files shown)


C7/UIElements/RightClickMenu.cs
```diff
@@ -100,45 +101,94 @@ public override void _Input(InputEvent @event)
...
+			bool isFortified = isUnitFortified(unit);
+			fortifiedCount += isFortified ? 1 : 0;
+			string action = getUnitAction(unit, isFortified);
+
 			AddItem($"{action} {unit.Describe()}").Connect("pressed", this, "SelectUnit", new Godot.Collections.Array() {unit.guid});
 		}
+		int unfortifiedCount = tile.unitsOnTile.Count - fortifiedCount;
+
+		if (fortifiedCount > 0) {
+			AddItem($"Wake All ({fortifiedCount} units)").Connect("pressed", this, "ForAll", new Godot.Collections.Array() {tile.xCoordinate, tile.yCoordinate, false});
```
pcen, line 149: instead of using tile.XCoordinate and tile.YCoordinate pair, it might be nice to use 1-dimensional coordinates for tile ID

QuintillusCFC: Having gone the 1D route in parts of my scenario editor, I prefer the 2D approach.  The downside is the conversion (calls to tileAt).  But it's easier for humans to reason about the 2D grid, even with it being isometric.  If we need to debug something, X/Y makes it much easier to map up to what's visually happening on the map.  I never want to have to say, "the 2347th tile from the start".  Much easier to say, "The tile at 47, 31".

That said, I believe you could pass the Tile object instead of the x, y coordinates here.  It's all on the front-end, so there's no message about a tile to the engine (only a message about a unit).  It appears that this method converts the Tile to coordinates, and ForAll then converts it back to a Tile on line 173... so passing the Tile would save a conversion both ways.

pcen: I actually tried using a tile initially, but in order for Godot to pass params to the connected function, the parameters need to be a serializable type, which apparently tile is not, so it seems easier to pass the coordinates than to figure out how to make Tile serializable given it would only saves 1 parameter. Makes sense about x/y being more human friendly though 👍

QuintillusCFC: Oh yeah, I ran into that back in November, forgot about the serializeable requirement for Godot events.  There's a work-around in the form of ParameterWrapper.cs, in the top level of the C7 project, that we use in a few places.

Although given that we only really need the coordinates here, I don't feel compelled to ParameterWrap the Tile.

C7/UIElements/RightClickMenu.cs
```diff
@@ -100,45 +101,94 @@ public override void _Input(InputEvent @event)
...
 		RemoveAll();
+
+		bool isUnitFortified(MapUnit unit) {
+			if (uiUpdatedUnitStates is null || !uiUpdatedUnitStates.ContainsKey(unit.guid)) {
+				return unit.isFortified;
+			}
+			return uiUpdatedUnitStates[unit.guid];
+		}
+
+		string getUnitAction(MapUnit unit, bool isFortified) {
```
QuintillusCFC: I'd kind of favor making this and isUnitFortified private methods that aren't embedded in another private method.  That would keep the focus of this method at a higher level.

QuintillusCFC (approved):
One stylistic comment, but the logic looks sound and it works in practice.

C7/UIElements/RightClickMenu.cs
pcen: sounds good, I'll change that

QuintillusCFC:
So, I missed on this (and have now confirmed in Civ3) that wake/fortify all should only apply if there are at least 2 applicable units on the tile.  Which makes sense, it's redundant if there's only one.  Not a major issue, but noting it before I forget it.

pcen:
> So, I missed on this (and have now confirmed in Civ3) that wake/fortify all should only apply if there are at least 2 applicable units on the tile. Which makes sense, it's redundant if there's only one. Not a major issue, but noting it before I forget it.

Nice catch. I'll definitely forget so I'll make an issue to fix this

Issue #209, opened 2022-04-07
Methodology for Debugging and Tracing


pcen:
While developing I am encountering runtime exceptions such as:

```
*** Processing turn 15 ***
Moving barbarian at [15, 57] (Plains on Plains) to [15, 59] (Plains on Plains)
Moving barbarian at [50, 64] (Grassland on Grassland) to [49, 65] (Forest on Grassland)
Moving barbarian at [16, 22] (Grassland on Grassland) to [14, 22] (Plains on Plains)
Moving barbarian at [63, 35] (Forest on Grassland) to [63, 37] (Hills on Grassland)
Moving barbarian at [13, 45] (Grassland on Grassland) to [14, 46] (Forest on Grassland)
Moving barbarian at [70, 54] (Forest on Plains) to [71, 53] (Forest on Plains)
Moving barbarian at [13, 45] (Grassland on Grassland) to [15, 45] (Forest on Grassland)
Moving barbarian at [16, 30] (Forest on Grassland) to [16, 32] (Grassland on Grassland)
Moving barbarian at [16, 50] (Forest on Grassland) to [17, 51] (Grassland on Grassland)
Moving barbarian at [17, 51] (Grassland on Grassland) to [15, 51] (Grassland on Grassland)
-> Begin Athens turn
Set random exploration AI for Chariot with 2 movement points and 3 hit points, guid = 95b35a93-d959-4009-bb25-aee601cb9d7c
-> Begin Philadelphia turn
Set random exploration AI for Chariot with 2 movement points and 3 hit points, guid = df5afcca-10c2-4284-9ded-bb781231f64b
-> Begin Amsterdam turn
Set random exploration AI for Chariot with 2 movement points and 3 hit points, guid = 3419d188-6427-430d-8643-8413f17ffe63
libc++abi: terminating with uncaught exception of type std::__1::system_error: recursive_mutex lock failed: Invalid argument

=================================================================
	Native Crash Reporting
=================================================================
Got a abrt while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================

=================================================================
	Native stacktrace:
=================================================================
	0x103109bb5 - /Applications/Godot.app/Contents/MacOS//Godot :
	0x1030fd77e - /Applications/Godot.app/Contents/MacOS//Godot :
	0x1031091af - /Applications/Godot.app/Contents/MacOS//Godot :
[...]

Issue #171, opened 2022-03-29
Flashing and illegible city UI


WildWeazel:
At certain zoom levels, and especially after zooming, the city info boxes flash white. It's probably best to hide them past a certain zoom level anyway, or devise a scheme to hide or simplify different things like the strategic view in Civ5, because they become illegible smears.

See also #172

QuintillusCFC:
I've also noticed this happen late game (200+ turns into the game), even at 100% zoom.  It could be related to the number of things being displayed, which is artificially high late game with ICS and units all over the place, but I also suspect my implementation of the city UI is highly sub-optimal, as it was back in Aztec days when I was still pretty new to Godot.  Thanks for writing it up!

QuintillusCFC:
I took a stab at figuring this out tonight; see the 171-Debugging-Attempt branch.

Long story short, I didn't succeed.  I removed progressively more code, but this line:

```
looseView.DrawTextureRectRegion(label, labelDestination, allOfTheLabel);
```

Stubbornly causes the flashing white rectangle even as the label becomes as simple as a red rectangle, and removing everything else doesn't help.  Unfortunately, that line is kind of important as `label` contains almost all the graphics for the city label.  The only thing it doesn't contain is the text (and the city graphics themselves).

Logic would seem to dictate that a small percentage of the time, the `labelImage` from which `label` is created doesn't get any graphics drawn to it for some reason, and is thus just a white rectangle.  But why?  When?  Is one of the methods we are calling asynchronous, and not returning in time occasionally?  Nothing is jumping out as the cause...

QuintillusCFC:
The flashes seem to be correlated with these Godot errors that have long bamboozled us:

```
ERROR: Condition "!id_map.has(p_rid.get_data())" is true. Returned: nullptr
```

Indeed, removing the line mentioned above seems to remove them.  But why?

WildWeazel:
Interesting that they are connected. I wonder if creating even a very simple scene instead of drawing it programmatically would have any effect.

QuintillusCFC:
Tried out Godot 3.4.5 RC1, the problem persists there.  So if it's an engine thing, it isn't fixed yet.  Seems to be a common error that people run into and don't know how to fix...

QuintillusCFC:
> Interesting that they are connected. I wonder if creating even a very simple scene instead of drawing it programmatically would have any effect.

[...]

PR #217, opened 2022-04-08
Remove Loading Resources in City Drawing

pcen:
Removes loading fonts and images from resource loader from CityLayer.drawObject

Far from a proper re-factor of CityLayer.drawObject, but at least it has private font and Image instances instead of relying on Godot's resource cache.

Potentially addressing some of the crashes discussed in https://github.com/C7-Game/Prototype/issues/209

### DIFF
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -286,7 +286,7 @@ public class ForestLayer : LooseLayer {
 		pinePlainsTexture        = Util.LoadTextureFromPCX("Art/Terrain/plains forests.pcx"   , 0, 704, 768, 176);
 		pineTundraTexture        = Util.LoadTextureFromPCX("Art/Terrain/tundra forests.pcx"   , 0, 704, 768, 176);
 	}
-	
+
 	public override void drawObject(LooseView looseView, GameData gameData, Tile tile, Vector2 tileCenter) {
 		if (tile.overlayTerrainType.Key == "jungle") {
 			//Randomly, but predictably, choose a large jungle graphic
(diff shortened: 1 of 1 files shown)


QuintillusCFC (approved):
This also seems to reduce https://github.com/C7-Game/Prototype/issues/171, though not completely eliminate it.  Playing through 100+ turns with and without it, it takes more cities to become an issue with these changes.

pcen:
> This also seems to reduce #171, though not completely eliminate it. Playing through 100+ turns with and without it, it takes more cities to become an issue with these changes.

I'm pretty sure that cities themselves, or some other class should contain all of the Image instances required to draw the cities, and only re-recalculate width, height, offsets, etc. when something about the city changes (ie. population grows, production changes, etc.) as re-creating those may cause some stuttering. Just a guess, but long term I think we should still refactor that code some more.

QuintillusCFC:
> > This also seems to reduce #171, though not completely eliminate it. Playing through 100+ turns with and without it, it takes more cities to become an issue with these changes.
> 
> I'm pretty sure that cities themselves, or some other class should contain all of the Image instances required to draw the cities, and only re-recalculate width, height, offsets, etc. when something about the city changes (ie. population grows, production changes, etc.) as re-creating those may cause some stuttering. Just a guess, but long term I think we should still refactor that code some more.

Yes and no (or "jein", as I like to say).  The cities can't contain that information, at least directly, as it is front-end Godot, and the cities are back-end data storage.  But I do think it should be possible to cache some of that on the front end to reduce the amount of re-calculation, and agree that a more thorough refactor would be good.

One potential time to do that would be when we make the change so that cities that aren't the capital don't display the capital star.  That isn't on the schedule yet.  It's small enough that I wouldn't object if someone did want to run with it... but if that doesn't happen it's likely to be a Dutch item.

Issue #191, opened 2022-04-05
Dijkstra's Algorithm - Consider Whether Tiles have Roads


QuintillusCFC:
Depends on #100 .

Road tiles will only cost 1/3 of a movement point (by default; this should be configurable).  This affects their weight in terms of unit pathing.

This could also include that rivers _don't_ get that bonus (without a future bridge-building tech), but that can be spun off into a separate issue if this doesn't turn out to be easy.

Kright:
We use integers for movement costs, but road requires fractional parts (like 1/3). In addition, there are railroads. For example, in 4th civ it has movement cost 1/10.  
What we should do with this? Use floats and be careful because of possible rounding errors?

QuintillusCFC:
That is a great question and one where we haven't decided on an answer.  As you say, floats could be used, but we run the risk of three movements on road resulting in a unit having only used, e.g. 0.9999982 of its movement points, and thus being able to move again when it shouldn't (not sure if 1/3 will actually be under or over the real value, but you get the point).

Two ideas that I've thought of:

 - Use floats, but add a very small amount to each movement cost to each move that isn't free (e.g. 1/1000th of a point or less to everything that isn't like railroads in Civ3) to account for this.  In the UI, we'll only display to a certain precision, but ensuring that 1/3 costs is 0.3333334 instead of 0.3333332 can avoid errors.
 - Use fractions, numerator and denominator.  Have our own math system for movement.  A Knight starts with 2/1 movement, if it moves on a road that costs 1/3 of a movement, the math system will expand 2/1 to 6/3, and then subtract 1/3 so it now has 5/3 movement left.  If it then moves on a railroad that costs 1/10 of a movement, the 5/3 becomes 50/30, and then we subtract to get 47/30 movement points left.  In the UI, it can still be displayed as decimal, though we could print out the fractions for debugging.

I suspect the floats-but-add-a-bit would be faster and may be simpler.  But the part of me that remembers when math was my favorite subject likes the fraction idea.

There may be other good ideas I haven't though of too.

Kright:
I think it's better to use floats and treat them as equal when abs(a - b) < eps. Eps is something small, 0.0001 for example. I'm going to make struct wrapping float and properly implement =, >, < operators for it. It will be less error-prone.

QuintillusCFC:
That also sounds like a good idea.  I've done something similar with timestamps in the past.

PR #219, opened 2022-04-08
Add Dijkstra's Algorithm for Pathing

### LINKED ISSUE #100: Add Dijkstra's Algorithm as Pathing Option to consider Tile Weights
The next step in unit pathing is to add Dijkstra's Algorithm, which will consider the weights of edges.

For now, the weights are simply going to be based on the tile movement cost, although later on that will includes roads, bridges or lack thereof, etc.

It may be helpful to add mouse go-to navigation first in order to test this, as the AI is only sending out one-movement-point units so far.

pcen:
With this implementation, the weight of each edge is calculated with the newly added `Tile.MovementCost` function:
```
public int MovementCost() {
	return overlayTerrainType.movementCost;
}
```
This method probably doesn't belong in C7GameData, and in the future Dijkstra's will need to consider other factors such as units blocking the path, cities, and [roads](https://github.com/C7-Game/Prototype/issues/191). Furthermore, this implementation does not consider that units can always use their remaining movement point to move onto any tile, so sometimes units will take a suboptimal path.

The implementation also returns the path as soon as the destination tile has been visited, however in the future when multiple units are either pathing from the same starting tile, or one unit is attempting to discover a subpath to the same destination, it may be better to compute the shortest path to all tiles once and re-use the result for multiple units.

### DIFF
--- a/C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
+++ b/C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
@@ -56,6 +56,11 @@ namespace C7Engine.Pathing
 					}
 				}
 				closest = getNextClosest();
+				// TODO: this is fastest if recomputing Dijkstra's for each unit
+				// and ignoring that units may path from the same starting tile
+				if (closest.Key == destination) {
+					break;
+				}
 			}
 
 			return ConstructPath(destination, predecessors);
--- a/C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
+++ b/C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
@@ -13,11 +13,33 @@ namespace C7Engine.Pathing
 	 * Notes: - Edge weight is defined by the movement cost of the 2nd node:
 	 *          w(u, v) = movement_cost(v)
 	 */
-	public class DijkstrasLandAlgorithm : PathingAlgorithm
-	{
-		//N.B. This should really be static, but we can't put a static method on interfaces, so it isn't.
-		public TilePath PathFrom(Tile start, Tile destination)
-		{
+	public class DijkstrasLandAlgorithm : PathingAlgorithm {
+
+		// Returns the next closest vertex in the graph, and updates the visited set to include
+		// the returned vertex
+		private KeyValuePair<Tile, int> getNextClosest(Dictionary<Tile, int> dist, HashSet<Tile> visited) {
+			// TODO: pick a different data structure to avoid O(n) search
+			KeyValuePair<Tile, int> next = new KeyValuePair<Tile, int>(Tile.NONE, int.MaxValue);
+			foreach (KeyValuePair<Tile, int> pair in dist) {
+				if (pair.Value < next.Value && !visited.Contains(pair.Key)) {
+					next = pair;
+				}
+			}
+			visited.Add(next.Key);
+			return next;
+		}
+
+		// Updates the shortest distance to a vertex. Returns true if the distance parameter
+		// is shorter than the shortest known distance and dist is updated, and false otherwise.
+		private bool updateShortestDistance(Tile tile, int distance, Dictionary<Tile, int> dist) {
+			if (!dist.ContainsKey(tile) || dist[tile] > distance) {
+				dist[tile] = distance;
+				return true;
+			}
+			return false;
+		}
+
+		public override TilePath PathFrom(Tile start, Tile destination) {
 			// shortest distance from start to each tile on the continent
 			Dictionary<Tile, int> dist = new Dictionary<Tile, int>();
 			Dictionary<Tile, Tile> predecessors = new Dictionary<Tile, Tile>();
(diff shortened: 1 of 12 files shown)


pcen:
I realized that in order to write proper integration tests for this (which we definitely should), it will require a lot of boilerplate or a refactor / improvement of the current `GameMap` to make it easier to programatically generate maps. The latter should probably be done in order to make testing all forms of pathing (and other game scenarios potentially) easier, so it may be good to make an issue and PR for a better C# interface for scripting basic maps and scenarios, but I'd like to get people's thoughts on this.

QuintillusCFC:
In short, I agree.  Being able to write tests around maps more easily would be a very good thing.  Whether that's directly programmatically, or just being able to point to a JSON file that defines a super-simple "map" for the purposes of testing.

It would help with AI testing as well.  While the "what should the AI do?" part is somewhat subjective, the "how should the AI try to do that?" involves steps that would lend themselves well to tests.

pcen:
I created https://github.com/C7-Game/Prototype/issues/220 to address testability since this will require a decent amount of refactoring and new code.

QuintillusCFC:
Sorry about not getting to this yet; I thought it was like 700 lines but it turns out about 550 of them are whitespace.  So it's a lot closer to the length I would have expected.  Examining now...

C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
```diff
@@ -0,0 +1,87 @@
...
+				if (closest.Key == destination) {
+					break;
+				}
+			}
+
+			return ConstructPath(destination, predecessors);
+		}
+
+		// Should not be public
+		public static TilePath ConstructPath(Tile destination, Dictionary<Tile, Tile> predecessors)
```
QuintillusCFC: This appears to be identical to the version in BFSPathingAlgorithm, other than improving the syntax a bit.  We should probably either make PathingAlgorithm an abstract class with this as a shared method*, or make it a static method on the base PathingAlgorithm.

Longer term, we might have to revisit that if we introduce pathing algorithms that don't use this method, so I'm not 100% sold on my own suggestion - but anticipating the future is never foolproof, so optimizing for what makes sense now is generally the right thing to do.

* = not 100% sure that works the same in C# as it does in Java

C7GameData/Tile.cs
```diff
@@ -51,6 +51,14 @@ public Tile()
 			unitsOnTile = new List<MapUnit>();
 		}
 
+		// TODO: this should be either an extension in C7Engine, or otherwise
+		// calculated somewhere else, but it's not obvious to someone unfamiliar
+		// with the save format that it's the overaly terrain that has actual
+		// movement cost
```
QuintillusCFC, line 58: The base/overlay terrain distinction has been a source of confusion in Civ3 since time immemorial.  My general rule nowadays is always look at the overlay; the base is just for drawing and for if the tile gets chopped down.  But I think this method is a good way to make things clearer.

I also believe this is the appropriate place for it, and it should not be an extension method.

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -14,8 +14,7 @@ public static void EndTurn()
 
 			foreach (Player player in gameData.GetHumanPlayers()) {
 				foreach (MapUnit unit in player.units) {
-					Console.WriteLine($"{unit}, path length: {unit.path?.PathLength() ?? 0}");
-					if (unit.path?.PathLength() > 0) {
+					if ((unit.path?.PathLength() ?? -1) > 0) {
```
QuintillusCFC: This is probably modified due to the possibility of the Queue named `path` on line 8 of TilePath being null?  Which can probably only happen with TilePath.NONE?

If so, probably the right solution is to fix the defaults in TilePath - setting `path` to an empty (but not null) queue, and the `Tile destination` to Tile.NONE.  Then we shouldn't have to worry about null at all (and that also allows a bit of clean-up of null cases in TilePath).

(And yes, it looks like I didn't think of that when adding TilePath.  Oops)

C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
```diff
@@ -0,0 +1,87 @@
...
+		public TilePath PathFrom(Tile start, Tile destination)
+		{
+			// shortest distance from start to each tile on the continent
+			Dictionary<Tile, int> dist = new Dictionary<Tile, int>();
+			Dictionary<Tile, Tile> predecessors = new Dictionary<Tile, Tile>();
+			HashSet<Tile> visited = new HashSet<Tile>();
+
+			dist[start] = 0;
+
+			bool updateShortestDistance(Tile tile, int distance) {
```
QuintillusCFC: I try not to notpick style in code reviews (especially before we had the .editorconfig)... but why not make this a private helper method instead of an line one?  It would interrupt the flow of reading the method significantly less as a helper method than being inline.

QuintillusCFC: Ah, perhaps because it modifies the dist Dictionary?  I'd still favor kicking it out into a separate method.

C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
```diff
@@ -0,0 +1,87 @@
...
+					return true;
+				} else if (dist[tile] > distance) {
+					dist[tile] = distance;
+					return true;
+				}
+				return false;
+			}
+
+			// TODO not O(n) search
+			KeyValuePair<Tile, int> getNextClosest() {
```
QuintillusCFC: See above about putting in a method.  Although it would take params, once both of these are moved, the top-level method would read easier - what follows this method essentially _is_ Dijkstra's, these are just the sub-components.

For now, I wouldn't worry about the O(n).  My textbook has descriptions of various options for these.  But I always anticipated that this would be an iterative process of making things more optimal.

C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
```diff
@@ -0,0 +1,87 @@
...
+			while (closest.Key != Tile.NONE) {
+				foreach (Tile tile in closest.Key.GetLandNeighbors()) {
+					if (!visited.Contains(tile) && updateShortestDistance(tile, closest.Value + tile.MovementCost())) {
+						predecessors[tile] = closest.Key;
+					}
+				}
+				closest = getNextClosest();
+				// TODO: this is fastest if recomputing Dijkstra's for each unit
+				// and ignoring that units may path from the same starting tile
+				if (closest.Key == destination) {
```
QuintillusCFC, line 60: So, yeah.  This is one of the great questions - should we return early as you have here, or not?

In order to be efficient, we probably do want to have some sort of caching.  But as you point out, it's also not efficient to calculate everything if the situations - edge lengths, in math terms - are going to change before the next unit requests a path, which could happen if a worker builds a road, for example.

Long term, I think caching results will be helpful.  But we'll also have to deal with cache invalidation, one of the three hard problems in computer science.  This is another area where I expect a lot of iteration as we seek to get the optimal results.

For now though, this is perfectly fine.  We need Dijkstra's to work for roads in particular, not to be optimal on a Huge map with 7000 units.

QuintillusCFC (commented):
How'd I wind up with 7 comments?  It fundamentally looks correct though.  I think they're all style related, or general commentary on your notes.

QuintillusCFC (approved):
Okay, yeah, reviewed my comments, no bugs spotted, just some style things.  Giving approval.  Hopefully the merge conflicts aren't too bad, TurnHandling had a bunch of them for me but you don't have many changes there so it shouldn't be as bad.

C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
pcen: The main reason was because I didn't have to type out as many parameters, and I have no objection to changing it to a helper if it improves readability. The other reason is because since it modifies `dist`, I prefer a nested method since `dist` can only refer to the `dist` in the outer scope and my IDE is aware of this, whereas with a helper method you have to figure out that you're passing in a mutable/output parameter which is actual `dist` in the main dijsktra's function

C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
pcen: I was doing some reading about caching in graph traversal and found a couple of interesting papers on how/when to cache. I only skimmed through them, but it's possible that those concepts may be applicable in our case as well so I'll try and find them to share. I agree we will probably need caching, which will definitely introduce complexity over time

C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
QuintillusCFC: Hmm... you're right, dist would be a mutable parameter (although I believe it wouldn't technically be a C# `out` parameter; you could modify the collection with it being passed as a "regular" parameter).

I think it comes down to style and background.  I've mostly worked in Java/Kotlin/TypeScript, and given the length of these inline functions, they'd usually be kicked out.  A lambda that is a couple lines, or a short anonymous inner function, might stick around, but if the inner function rivals the length of the parent one, it should probably be its own named function; I'd expect my colleagues to request I move one of this length.  But it's also not a hard-and-fast boundary, and I'm sure I have some anonymous inner functions of humungous length in my editor from back in the day.

I would ask what your favorite IDE is here, but I think instead I'll create a poll on GitHub Discussions to try that out.

pcen:
@QuintillusCFC I've address all the feedback you left so it should be fine to merge now

PR #221, opened 2022-04-09
City production based on tile yield

### LINKED ISSUE #141: City Production Rate Based on Tile Yield
Instead of being at a hard-coded fixed rate, city production (food and materials) should be based on the tile yield.

This applies for both human and AI players.

QuintillusCFC:
Resolves #141 .

City production (food, production, and commerce although the latter is not yet used) will now be based on actual tile yields rather than hard-coded.

This has a downside - the AI will often be spawned in an area where it won't generate enough food to grow enough to spawn a second settler.  But it is, ultimately, almost necessary to have in order to have other features such as irrigation and resource bonuses be effective.  Long term, we should have some guardrails around AI spawning anyway.

As part of this, I added a test project for C7GameData.  My vote would be to include tests with the main project (C7GameData, in this case), rather than as a separate parallel project, but since .NET seems to encourage separate projects, I've done that for now.  I don't understand the benefit of it though; the Java model of "tests live with the code they test" makes more sense to me.  Still, it's another unit test, which is not a bad thing.


pcen (approved):

QuintillusCFC:
Pretty sure I got all the rebases fixed up.  Going to test it on my desktop just in case; it looks like we jumped to .NET 6 since I last did something on my laptop, so the laptop is not working with C7 right now.

QuintillusCFC:
Closing in favor of #232 , which is the same except excluding [391a1a4](https://github.com/C7-Game/Prototype/pull/221/commits/391a1a4a28889bb99ad0ea67f4d04793035b1f18), which was supposed to be on another branch, and resolving the merge conflicts that proved intractable by starting over from Development.

PR #225, opened 2022-04-09
Fix Wake / Fortify All

### LINKED ISSUE #222: Wake / Fortify All should only appear applicable to 2+ units

pcen:
only display if the action is applicable to more than one unit in the stack

### DIFF
--- a/C7/UIElements/RightClickMenu.cs
+++ b/C7/UIElements/RightClickMenu.cs
@@ -135,20 +135,21 @@ public class RightClickTileMenu : RightClickMenu
 		RemoveAll();
 
 		int fortifiedCount = 0;
-		foreach (MapUnit unit in tile.unitsOnTile) {
+		List<MapUnit> units = tile.unitsOnTile.FindAll(unit => unit.owner.guid == game.controller.guid);
 
+		foreach (MapUnit unit in units) {
 			bool isFortified = isUnitFortified(unit, uiUpdatedUnitStates);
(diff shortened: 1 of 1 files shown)


QuintillusCFC (commented):
This looks good - but I just found another edge case.  Fortify/wake all should only appear if it's your units on the tile.  I just right clicked on a tile with a Roman worker to get the debugging print of tile yields on the console, and noticed that I could fortify it.  The AI seems to ignore that, and continue on anyway, but it still plays the fortify animation, so we ought to add a check for that.

Ideally, that would probably be that the unit belongs to the player who is the Controller in Game.cs; the Controller role is designed to be the local human player.  In practice, pre-multiplayer, a player.isHuman check would be equivalent for now.

pcen:
> This looks good - but I just found another edge case. Fortify/wake all should only appear if it's your units on the tile. I just right clicked on a tile with a Roman worker to get the debugging print of tile yields on the console, and noticed that I could fortify it. The AI seems to ignore that, and continue on anyway, but it still plays the fortify animation, so we ought to add a check for that.
> 
> Ideally, that would probably be that the unit belongs to the player who is the Controller in Game.cs; the Controller role is designed to be the local human player. In practice, pre-multiplayer, a player.isHuman check would be equivalent for now.

Makes sense. In order to properly support multiplayer, the engine should also compare the sender of messages to the owner of the unit(s) to make sure it's a legal operation.

PR #227, opened 2022-04-09
Refactor turn handling

maxpetul:
Refactors turn handling to create something like a game loop and removes the old obsolete turn handling from the UI. The game loop is now in TurnHandling.AdvanceTurn, which contains most of the code that used to be in EndTurn. AdvanceTurn gets called when the game is started and when the player is done moving their units. (A proper game loop wouldn't need to get called constantly but this is good enough for now). I also factored out OnBeginTurn, a method that gets called at the start of each turn before the movement phase.


C7Engine/EntryPoints/MessageToEngine.cs
```diff
@@ -159,8 +159,17 @@ public override void process()
 	public class MsgEndTurn : MessageToEngine {
 		public override void process()
```
pcen, line 160: Would it make sense to have process call out to an OnEndTurn method in the turn handler? Organisationally it might be nice to have the begin / end / advance logic all in the same place

maxpetul: I don't think it would make sense to simply move that code into a method called OnEndTurn. For one thing it applies only to the UI controlling player, unlike OnBeginTurn, so they wouldn't be symmetric. For another thing, that code doesn't actually end the turn from the perspective of the engine, it only ends the movement phase for the UI controller. I plan to refactor AdvanceTurn again at some point using "yield return" to create a state machine instead of requiring the engine to call it repeatedly. Once that's done it will be easy to insert OnBegin/EndMovementPhase methods that are called for each player. Then that code in MsgEndTurn.process could be (and should be) moved into OnEndMovementPhase.

pcen: Ok, makes sense that it's at the controller level. Sounds good to me

pcen (approved):

PR #229, opened 2022-04-10
Refactor GUID

pcen:
Change all current uses of guid strings to be System.Guid instead.
- System.Guid is passed by value, so comparing Guid structs doesn't require any dereferences, while comparing string requires 2

### DIFF
--- a/C7GameData/Player.cs
+++ b/C7GameData/Player.cs
@@ -2,71 +2,64 @@ using System.Collections.Generic;
 
 namespace C7GameData
 {
-using System;
-
-public class Player
-{
-	public string guid { get; set; }
-	public int color { get; set; }
-	public bool isBarbarians = false;
-	//TODO: Refactor front-end so it sends player GUID with requests.
-	//We should allow multiple humans, this is a temporary measure.
-	public bool isHuman = false;
-	public bool hasPlayedThisTurn = false;
-
-	public Civilization civilization;
-	private int cityNameIndex = 0;
-	
-	public List<MapUnit> units = new List<MapUnit>();
-	public List<City> cities = new List<City>();
-	public TileKnowledge tileKnowledge = new TileKnowledge();
-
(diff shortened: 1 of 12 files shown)


maxpetul (approved):
Looks good to me.

C7GameData/Player.cs
```diff
@@ -2,71 +2,64 @@
...
-
-public class Player
-{
-	public string guid { get; set; }
-	public int color { get; set; }
-	public bool isBarbarians = false;
-	//TODO: Refactor front-end so it sends player GUID with requests.
-	//We should allow multiple humans, this is a temporary measure.
-	public bool isHuman = false;
-	public bool hasPlayedThisTurn = false;
```
QuintillusCFC, line 15: This variable shouldn't have been removed; code in TurnHandling.cs depends on it.  I'm pretty sure Development doesn't compile currently due to its removal :(.

It looks like the guid (line 9) also got deleted, but is still in use.  Although there also is a line in GameData.cs where it tries to access the guid as a Guid instead of a string, which means that it doesn't work when I revert line 9.

Most likely due to a bad rebase or merge, but it is a +1 for having an automatic compile check on PRs.

QuintillusCFC: It looks like EngineStorage.cs is using the player guid as a string, and GameData.cs is using it as a Guid.  It looks like the new "unique" class is what introduced a Guid?

pcen: hmm, iirc I rebased without conflict from my working branch and assumed it was fine. I'll revert for now and fix this later

QuintillusCFC: That is what an automatic PR build would catch though... if we assumed our code wasn't working, we wouldn't think it was ready to merge :).

pcen: haha yes I agree, we should definitely compile / run tests before merging 😄

PR #232, opened 2022-04-11
City Production Based on Tile Yield

### LINKED ISSUE #141: City Production Rate Based on Tile Yield

QuintillusCFC:
Resolves #141 .

This is basically the same as #221 , but the merge conflicts there were too complex to resolve, so instead I started over from Development, cherry-picking each commit at a time, modifying the code as needed, and testing at each stage to make sure it worked.

### DIFF
--- a/C7Engine/AI/CityTileAssignmentAI.cs
+++ b/C7Engine/AI/CityTileAssignmentAI.cs
@@ -25,11 +25,13 @@ namespace C7Engine.AI
 			double maxScore = 0;
 			Tile preferredTile = Tile.NONE;
 			foreach (Tile t in cityCenter.neighbors.Values) {
-				double score = CalculateTileYieldScore(t, targetTileFoodAmount);
-				Console.WriteLine($"Tile {t} scored {score}");
-				if (score > maxScore) {
-					maxScore = score;
-					preferredTile = t;
+				if (t.personWorkingTile == null) {
+					double score = CalculateTileYieldScore(t, targetTileFoodAmount);
+					Console.WriteLine($"Tile {t} scored {score}");
+					if (score > maxScore) {
+						maxScore = score;
+						preferredTile = t;
+					}
 				}
 			}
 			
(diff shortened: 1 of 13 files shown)


C7Engine/AI/CityTileAssignmentAI.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using C7GameData;
+
+namespace C7Engine.AI
+{
+	public class CityTileAssignmentAI
+	{
+		public static int DESIRED_FOOD_SURPLUS_PER_TURN = 2;
```
pcen: on style, naming isn't in the `.editorconfig` yet, but pretty much all C# style guides suggest PascalCase for constant values (I assume it's intended to be constant given how it's named)

pcen (approved):

C7Engine/AI/CityTileAssignmentAI.cs
QuintillusCFC: Yes and no on whether it's intended to be a constant.  Most of these are what I'd call "semi-constant", in that it is intended to be mod configurable, but not to be updated.  In Java, it might be a _final_ variable that would be initialized at runtime.

I'll admit that I don't understand why C# recommends PascalCase for... seemingly everything other than local variables.  Java, C, C++, and JavaScript all use SCREAMING_SNAKE_CASE for constants, and it seems to be recognized as such.  C# recommends PascalCase for both method-level variables and constants.  I guess the implication is the C# designers didn't think whether a variable was a constant was as significant as their predecessors did.

I'll go ahead and change this one since it arguably shouldn't be as much of a constant anyway (as cities grow, it should arguably change).  But I think it's a good example of an area where we could have a discussion of whether we want to go with what's most popular in C#, or what's most popular in the rest of the OO world.  IMO, the C/C++/Java/JS style more clearly conveys the meaning in this case.

PR #236, opened 2022-04-14
Experience levels

### LINKED ISSUE #193: Experience levels + promotions in combat
Units will still start out with 3 hitpoints, but can promote in combat, gaining a hitpoint.  IIRC, the odds are 1/8 when winning an attack, 1/16 when winning in defense - but these should be stored in variables so they can later be configured by mods.

The experience level concept should also be implemented, so instead of just "gaining a hitpoint", the unit gains an experience level (as well as the immediate +1 hitpoint from the promotion), if it is not already at the highest experience level.  Add default experience levels in the JSON that will be read in to the game.

Don't worry about great leaders, we have a lot of other things to consider before getting to that point (and even in Civ, that behavior is configurable).

maxpetul:
Closes #193
Experience levels now work how you'd expect: each time a unit gets a kill it gets a chance to promote and its max HP is now determined by its exp level instead of hard coded to 3. I set the promotion odds to the numbers Quintillus mentioned in the issue. They feel right to me but we should still make sure at some point. Under the hood, experience levels are read from the JSON save or BIQ. To make that work, I followed the example of terrain types. I put experience levels in the C7RulesFormat object instead of the game data object since some comments indicated that's what it's there for. The rules object wasn't used for anything before this so I had to do some work wiring it up, e.g. including it in EngineStorage and passing it to the dummy game creation process.

### DIFF
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -30,6 +30,21 @@ public static class MapUnitExtensions {
 		unit.isFortified = false;
 	}
 
+	public static void RollToPromote(this MapUnit unit, bool wasAttacking, bool waitForAnimation)
+	{
+		C7RulesFormat rules = EngineStorage.rules;
+		double promotionOdds = wasAttacking ? rules.promotionChanceAfterAttacking : rules.promotionChanceAfterDefending;
+		if (EngineStorage.gameData.rng.NextDouble() < promotionOdds) {
+			ExperienceLevel nextLevel = rules.GetExperienceLevelAfter(unit.experienceLevel);
+			if (nextLevel != null) {
+				unit.experienceLevelKey = nextLevel.key;
+				unit.experienceLevel = nextLevel;
+				unit.hitPointsRemaining++;
+				unit.animate(MapUnit.AnimatedAction.VICTORY, waitForAnimation);
+			}
+		}
+	}
+
 	public static bool fight(this MapUnit unit, MapUnit defender)
 	{
 		// Rotate defender to face its attacker. We'll restore the original facing direction at the end of the battle.
@@ -54,8 +69,12 @@ public static class MapUnitExtensions {
 				unit.hitPointsRemaining -= 1;
 		}
 
+		MapUnit loser = (defender.hitPointsRemaining <= 0) ? defender : unit,
+			winner = (defender == loser) ? unit : defender;
+
+		winner.RollToPromote(winner == unit, false);
+
 		// Play death animation
-		MapUnit loser = (defender.hitPointsRemaining <= 0) ? defender : unit;
 		loser.animate(MapUnit.AnimatedAction.DEATH, true);
 		loser.disband();
 
@@ -86,6 +105,7 @@ public static class MapUnitExtensions {
 			new MsgStartEffectAnimation(tile, AnimatedEffect.Miss, null, AnimationEnding.Stop).send();
 
 		if (target.hitPointsRemaining <= 0) {
+			unit.RollToPromote(true, false);
 			target.animate(MapUnit.AnimatedAction.DEATH, true);
 			target.disband();
 		}
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -72,7 +72,7 @@ public static class MapUnitExtensions {
 		MapUnit loser = (defender.hitPointsRemaining <= 0) ? defender : unit,
 			winner = (defender == loser) ? unit : defender;
 
-		winner.RollToPromote(winner == unit, false);
+		winner.RollToPromote(winner != defender, false);
 
 		// Play death animation
 		loser.animate(MapUnit.AnimatedAction.DEATH, true);
--- a/C7GameData/RulesFormat.cs
+++ b/C7GameData/RulesFormat.cs
@@ -16,8 +16,21 @@ namespace C7GameData
 		[JsonIgnore]
 		public ExperienceLevel defaultExperienceLevel;
 
+		// TODO: Make sure these numbers are correct
+		public double promotionChanceAfterDefending = 1.0/16;
+		public double promotionChanceAfterAttacking = 1.0/8;
+
 		public C7RulesFormat() {
 			Version = "v0.0early-prototype";
+		}
+
+		public ExperienceLevel GetExperienceLevelAfter(ExperienceLevel experienceLevel)
+		{
+			int n = experienceLevels.IndexOf(experienceLevel);
+			if (n + 1 < experienceLevels.Count)
+				return experienceLevels[n + 1];
+			else
+				return null;
 		}
 	}
 }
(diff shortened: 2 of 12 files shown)


C7GameData/RulesFormat.cs
```diff
@@ -5,11 +5,32 @@ namespace C7GameData
...
-            Version = "v0.0early-prototype";
-        }
-    }
+	using System.Collections.Generic;
+	using System.Text.Json.Serialization;
+
+	public class C7RulesFormat {
+		public string Version { get; set; }
+
+		public List<ExperienceLevel> experienceLevels = new List<ExperienceLevel>();
```
QuintillusCFC, line 14: Calling out @JimOfLeisure 's opinion on this, as he added this file back in November, and would know more about its intended use than I do.  The commit comment doesn't clarify it enough.

My first thought is that this variable should be in GameData.cs, like all of our other game data.

Of course that leaves the question of, "what is C7RulesFormat"?  Previously, it had just been a placeholder of a version string.  Maybe that's intentional?  Maybe it is meant to correspond to the RULE tab from Civ3 (which we haven't really added yet).

Overall though, I don't like that experienceLevels is over in left field here, away from everything else in GameData.  It's not obvious and there's not a clear reason for it.

I see the comment about comments indicating that's what this is here for; are there comments for that beyond the one at the start of this file?  Unless I'm missing something, I'd rather follow precedent than one comment.

QuintillusCFC: Perhaps also relevant, IMO there shouldn't be anything wrong with adding "new" items to our C7 data classes, including those linked from GameData, where appropriate.  promotionChanceAfterDefending/promotionChanceAfterAttacking being good examples of items that aren't moddable in Civ but should be in C7, and should be written out to the JSON save file.

They probably should be in some sort of "Rules" file, but I'd rather have the rules be part of the game data than separte from it.

C7GameData/SaveFormat.cs
```diff
@@ -32,9 +33,10 @@ public class C7SaveFormat
 		public C7SaveFormat()
 		{
 			GameData = new GameData();
+			Rules = new C7RulesFormat();
```
QuintillusCFC, line 36: Ah, now I see the comment on line 27 about rules being an equivalent to BIC/X/Q.  That is probably one of the ones you were referring to.

I'm kind of thinking that separation is an unnecessary holdover from Civ.  Civ3 embeds a BIQ within a SAV, and that leads to weird situations such as separate TILE and CITY objects, with different structures, in the embedded BIQ and in the "save" part of the SAV.  This was added in the pre-Aztec days and hasn't been used, so I wouldn't take it as gospel.

C7Engine/MapUnitExtensions.cs
```diff
@@ -54,8 +69,12 @@ public static bool fight(this MapUnit unit, MapUnit defender)
 				unit.hitPointsRemaining -= 1;
 		}
 
+		MapUnit loser = (defender.hitPointsRemaining <= 0) ? defender : unit,
+			winner = (defender == loser) ? unit : defender;
+
+		winner.RollToPromote(winner == unit, false);
```
QuintillusCFC: Shouldn't the first parameter be whether the unit was the attacker?  Looking at the parameter names for RollToPromote, I would expect it to be something like, `unit != defender`.

QuintillusCFC (changes requested):
See question about RulesFormat.cs; IMO that paradigm is likely a carryover from Civ3 that we shouldn't carry over.  Also, I think there's a parameter mismatch at one place (see comment).

C7GameData/RulesFormat.cs
maxpetul: I'll admit I haven't been paying much attention to the work you guys have been doing on saving to JSON. When I saw those comments on C7RulesFormat I just assumed they were recent and relevant. I figured the point of separating out the rules was to allow C7 save files to omit them and instead refer to a default rules file, like in the original game. If the plan is that all saves will have their rules embedded then I don't see any reason to keep the rules separate from the rest of the game data, and agree they should be merged.

Another reason to combine them is that the distinction between game data and rules is unclear in general and we might draw the line in the wrong place. In particular I'm thinking of space 4X games where it's common to feature a ship designer, meaning unit types would need to be part of the game data instead of the immutable rules.

C7Engine/MapUnitExtensions.cs
maxpetul: It's correct as written, but I can see how it would be misleading. `unit` is always the attacker. I replaced that expression with `winner != defender` for clarity. When I was originally writing this code I briefly considered renaming `unit` to `attacker` but decided not to for consistency with the other extension methods, and laziness.

C7GameData/RulesFormat.cs
QuintillusCFC: I agree that there is a not a clear-cut distinction between game data and rules.

IMO (and @WildWeazel can correct me if his understanding is different), it would be preferable to embed all the rules within the JSON, at least as much as feasible.  "As feasible" meaning that there will likely still have to be call-outs to mod files - custom graphics, AI routines, etc.  But by embedding as much of the rules as we can, we maximize the self-contained aspect of the save file.

C7Engine/MapUnitExtensions.cs
QuintillusCFC: Oh, yeah, I see it now, looking at it again.  Thanks for changing it though, it does make it more obvious.

C7GameData/RulesFormat.cs
maxpetul: I created an issue for C7RulesFormat here: #247. I'm going to merge it into C7GameData unless someone swoops in with a good reason to keep it around. But I'd like to first merge this and my other PRs for Combat Part II, then worry about C7RulesFormat.

JimOfLeisure:
The original JSON 'save format' was just a start. I thought it made sense to maintain a separate BIC analog versus game data analog because there isn't really a reason to copy per-mod data in every single save game file. I'm not saying it should or should have mirrored the Civ3 BIC, but clearly there is a bunch of data that doesn't change between games.

But a lot has happened since November, and *I* never did anything with C7RulesFormat that I can recall aside from 'reserving' it it a nascent 'save format'.

QuintillusCFC (approved):
Looks like we're in general alignment.  To Jim's point, eventually (once modding is farther along), I wouldn't be surprised if there is some data (scripts, images, sounds, etc.) that we don't want to copy per save.  But right now, 99% of the uncompressed save is the tile info, which compresses by 90%; simplicity seems more important than space when we're in the same range as Civ3 saves.  Approving/resolving my comments.

QuintillusCFC:
Removing "Carthage" tag as #193 is already included there, and was thus being double-counted for progress towards the next preview.

I _think_ the best way to handle this going forward is to use the "Closes #XYZ" reference to associate the PR with the issues, but to leave the "Carthage" tag off the PR itself (unless it's a PR that doesn't have a corresponding issue as in a "fix this blocker that needs done ASAP" PR).  That keeps the high-level projects board (https://github.com/C7-Game/Prototype/projects/3) focused on the issues/fixes, and allows estimating how far along we are towards milestones semi-accurately.

WildWeazel:
I concur with that

PR #244, opened 2022-04-18
Defensive bonuses

### LINKED ISSUE #195: Terrain defensive bonuses
Units should receive defensive bonuses when standing on terrain that provides such a bonus.

This will also involve importing that data from BIQ/SAV/JSON, so that it's dynamic based on the terrain type.

### LINKED ISSUE #196: Fortified units + units across river get defensive bonus
Units that are fortified should get a bonus, based on the to-be-imported "fortifications defensive bonus" field in the RULE section (see the [Apolyton docs](https://apolyton.net/forum/civilization-series/civilization-iii/122361-civilization-iii-bix-biq-file-format) for reference).

Since that's probably fairly small, while we're at it, let's add the "river defensive bonus" as well.

These would be good candidates for some unit tests if the infrastructure for that is already set up, especially the river case and variants thereof (e.g. diagonal attacks across rivers should still cause the defensive bonus).

maxpetul:
Closes #195, Closes #196
Adds defensive bonuses to the game. Bonuses are applied for fortification, terrain, river crossing, and city garrison. The city garrison bonus is a placeholder. It still needs to scale for city level, right now all cities have the level 2 bonus (+50% under standard rules).
The branch for this PR extends off of my experience levels branch. I did that so I could easily use the C7RulesFormat class to hold the defensive bonuses. Now it's looking like the rules object will be merged into the game data object. I'd like to get these PRs merged first, though, and settle that issue later. It wouldn't be difficult to move the defensive bonuses and experience levels over to the game data object.
As for how it's implemented: Instead of using `unit.unitType.defense` to get a unit's defense strength, you'd now use `unit.DefenseStrengthVersus(...)` which takes an opponent unit, a flag for bombard vs regular attack, and an optional attack direction. `DefenseStrengthVersus` computes the effective defense strength by summing a list of `StrengthBonuses` returned by `ListStrengthBonusesVersus`. This sounds like overkill but it's for a good purpose. `StrengthBonuses` include descriptions so they can be displayed in the interface for the player's erudition, like in Civ 4 when you hover over a possible attack. Also I added a little combat log that shows the bonuses when combat begins, this is useful for debugging. Lastly `ListStrengthBonusesVersus` is a convenient point to mod in additional strength bonuses, for example a bonus for spearmen versus mounted units or even Civ 4 style promotions.
I wrote that `HasRiverCrossing` function before I knew that the info it returns could be imported from the Civ 3 data. Quintillus already set that up for his river commerce PR so once that's merged I'll modify the river crossing code to use that instead.

### DIFF
--- a/C7/c7-static-map-save.json
+++ b/C7/c7-static-map-save.json
@@ -27,7 +27,28 @@
 	      "retreatChance": 0.66
 	  }
       ],
-      "defaultExperienceLevelKey": "regular"
+      "defaultExperienceLevelKey": "regular",
+
+      "fortificationBonus": {
+	  "description": "Fortified",
+	  "amount": 0.25
+      },
+      "riverCrossingBonus": {
+	  "description": "Behind river",
+	  "amount": 0.25
+      },
+      "cityLevel1DefenseBonus": {
+	  "description": "Town",
+	  "amount": 0.00
+      },
+      "cityLevel2DefenseBonus": {
+	  "description": "City",
+	  "amount": 0.50
+      },
+      "cityLevel3DefenseBonus": {
+	  "description": "Metropolis",
+	  "amount": 1.00
+      }
   },
   "gameData": {
     "turn": 0,
--- a/C7/c7-static-map-save.json
+++ b/C7/c7-static-map-save.json
@@ -6,25 +6,29 @@
 	      "key": "conscript",
 	      "displayName": "Conscript",
 	      "baseHitPoints": 2,
-	      "retreatChance": 0.34
+	      "retreatChance": 0.34,
+	      "promotionChance": 0.50
 	  },
 	  {
 	      "key": "regular",
 	      "displayName": "Regular",
 	      "baseHitPoints": 3,
-	      "retreatChance": 0.50
+	      "retreatChance": 0.50,
+	      "promotionChance": 0.25
 	  },
 	  {
 	      "key": "veteran",
 	      "displayName": "Veteran",
 	      "baseHitPoints": 4,
-	      "retreatChance": 0.58
+	      "retreatChance": 0.58,
+	      "promotionChance": 0.125
 	  },
 	  {
 	      "key": "elite",
 	      "displayName": "Elite",
 	      "baseHitPoints": 5,
-	      "retreatChance": 0.66
+	      "retreatChance": 0.66,
+	      "promotionChance": 0.0
 	  }
       ],
       "defaultExperienceLevelKey": "regular",
--- a/C7/c7-static-map-save.json
+++ b/C7/c7-static-map-save.json
@@ -67273,7 +67273,11 @@
         "baseFoodProduction": 0,
         "baseShieldProduction": 1,
         "baseCommerceProduction": 0,
-        "movementCost": 1
+          "movementCost": 1,
+	  "defenseBonus": {
+	      "description": "Desert",
+	      "amount": 0.10
+	  }
       },
       {
         "key": "plains",
@@ -67281,7 +67285,11 @@
         "baseFoodProduction": 1,
         "baseShieldProduction": 1,
         "baseCommerceProduction": 0,
-        "movementCost": 1
+          "movementCost": 1,
+	  "defenseBonus": {
+	      "description": "Plains",
+	      "amount": 0.10
+	  }
       },
       {
         "key": "grassland",
@@ -67289,7 +67297,11 @@
         "baseFoodProduction": 2,
         "baseShieldProduction": 0,
         "baseCommerceProduction": 0,
-        "movementCost": 1
+          "movementCost": 1,
+	  "defenseBonus": {
+	      "description": "Grassland",
+	      "amount": 0.10
+	  }
       },
       {
         "key": "tundra",
@@ -67297,7 +67309,11 @@
         "baseFoodProduction": 1,
         "baseShieldProduction": 0,
         "baseCommerceProduction": 0,
-        "movementCost": 1
+          "movementCost": 1,
+	  "defenseBonus": {
+	      "description": "Tundra",
+	      "amount": 0.10
+	  }
       },
       {
         "key": "flood plain",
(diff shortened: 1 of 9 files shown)


C7/c7-static-map-save.json
```diff
@@ -67224,111 +67273,167 @@
         "baseFoodProduction": 0,
         "baseShieldProduction": 1,
         "baseCommerceProduction": 0,
-        "movementCost": 1
+          "movementCost": 1,
+	  "defenseBonus": {
+	      "description": "Desert",
```
WildWeazel, line 67278: Why does defenseBonus have its own description instead of using displayName?

maxpetul: That was the easiest way to do it. As an alternative we could save only the defense bonus amounts and fill in the descriptions automatically at load time. We'll need to do another pass over descriptions in any case to make them translatable. They might become text keys and/or be auto-genenerated. For now I think it's fine to just store them as plain strings.

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -69,6 +69,8 @@ internal static void AdvanceTurn()
 						newUnit.location = tile;
```
WildWeazel: just a note for later, this whole section is a prime candidate for refactoring into some factory methods as mentioned in #238, maybe something like GetNextRandomUnit(rng) which calls UnitPrototype.GetInstance()

C7Engine/MapUnitExtensions.cs
```diff
@@ -30,19 +31,95 @@ public static void wake(this MapUnit unit)
 		unit.isFortified = false;
 	}
 
+	public static IEnumerable<StrengthBonus> ListStrengthBonusesVersus(this MapUnit unit, MapUnit opponent, bool attacking, bool bombard, TileDirection? attackDirection)
```
WildWeazel, line 34: I'm not really familiar with yield in C#. What's the purpose of it here over just building a list? A comment block explaining would be nice.

C7Engine/MapUnitExtensions.cs
```diff
@@ -67,15 +148,18 @@ public static bool fight(this MapUnit unit, MapUnit defender)
...
 			return; // Do nothing if we don't have a unit to attack. TODO: Attack city or tile improv if possible
 
 		var unitOriginalOrientation = unit.facingDirection;
 		unit.facingDirection = unit.location.directionTo(tile);
 
-		int defenderStrength = target.unitType.defense;
-		double attackerOdds = defenderStrength > 0 ? (double)unit.unitType.bombard / (unit.unitType.bombard + defenderStrength) : 1.0;
+		double bombardStrength  = unit  .AttackStrengthVersus (target, true, unit.facingDirection);
+		double defenderStrength = target.DefenseStrengthVersus(unit  , true, unit.facingDirection);
+		double attackerOdds = bombardStrength / (bombardStrength + defenderStrength);
```
WildWeazel, line 165: seems like we would also want BombardStrengthVersus and BombardDefenseVersus since those may have their own modifiers later on

C7GameData/RulesFormat.cs
```diff
@@ -5,11 +5,55 @@ namespace C7GameData
...
+	public class C7RulesFormat
+	{
+		public string Version { get; set; }
+
+		public List<ExperienceLevel> experienceLevels = new List<ExperienceLevel>();
+		public string defaultExperienceLevelKey;
+		[JsonIgnore]
+		public ExperienceLevel defaultExperienceLevel;
+
+		// TODO: Make sure these numbers are correct
```
WildWeazel: in fact, make 'em configurable per level

C7Engine/TileExtensions.cs
```diff
@@ -0,0 +1,18 @@
+namespace C7Engine
+{
+	using C7GameData;
+
+	public static class TileExtensions {
```
WildWeazel, line 5: I'd like to hear your thoughts on extension methods in #238, since my only experience with them is as a way to extend third party classes. What do they provide being in the same namespace as the data classes?

WildWeazel (approved):
Overall this is a good approach- functional encapsulation and separating Civ3 from C7 rules. I had a few comments but I'm ok with merging it.

C7Engine/MapUnitExtensions.cs
maxpetul: I'm no C# expert myself. Before C7, the only significant thing I've done with C# was mod Rimworld, and that's where I learned that pattern of a method yield-returning an `IEnumerable`. I like this pattern because it omits the list itself and focuses on its contents, which is what we actually care about. In general it's best to minimize mutable state. Doing things this way is also possibly more efficient since no list needs to be allocated. Though whether that's actually a net win for performance depends on how efficiently the compiler handles yield-returns. I read somewhere that this pattern gets compiled to a state machine, in other words it's not just building a list under the hood.

C7Engine/MapUnitExtensions.cs
maxpetul: As it is now, the Attack/DefenseStrengthVersus functions do double duty, as they also compute strengths for bombarding depending on that boolean flag. Since then I have also added a DefensiveBombardStrengthVersus method. We could split these up into six methods, but actually I was thinking of doing the opposite and combining them all into one StrengthVersus method that would take an enum specifying the combat type (Attack, Defense, Bombard, BombardDefense, DefensiveBombard, or DefensiveBombardDefense). Otherwise I think we'll end up with a lot of code duplication since the strength bonuses will be similar in many cases.

C7GameData/RulesFormat.cs
maxpetul: Sure, actually we have no choice. I looked around for the exact promotion chances and found this post from someone at Firaxis:
https://forums.civfanatics.com/threads/promotion-probabilities.26048/#post-352717
In the original game promotion chance varies by experience level.

C7Engine/MapUnitExtensions.cs
WildWeazel: That makes sense. I've been in python land too long to even think about mutability.

QuintillusCFC:
Removing this PR (and #236) from the milestone as the issues closed were already in the milestone, so it was resulting in our percent-done being higher than it really is (which is something that's going to be presented as progress in our next preview).

PR #245, opened 2022-04-18
240 Unit Prototype Refactor

### LINKED ISSUE #240: Unit attributes should be stored and used
Unit attributes (canBuildCity, pillage, etc.) should be stored on the unit prototypes, _and_ should be used.

The motivation right now is I want to start enhancing the AI's build decisions, and eventually unit order decisions.  However, right now we still have a lot of hard-coded decisions based on unit names - at least 15, via a quick search.  I don't want to keep propagating that indefinitely, so it's time to start adding these attributes.

This also ties into #175 ; some actions (e.g. Pillage) aren't implemented yet, so should be hidden.  I'll likely do that here, but there will still be other areas on that card (e.g. main menu, in-game menu) where non-implemented things are visible.

QuintillusCFC:
Resolves #240 .  Should be reasonable size minus the JSON (how do we teach GitHub not to count its lines of code?  Tortoise seems to have already picked up on that locally).

This adds unit prototypes to both our JSON, and our import from Civ3.  They will no longer be hard-coded.

I've structured them to have three configurable/expandable properties:

 - Category.  Land/air/sea for now.  Could be extended in the future.
 - Actions.  These are, well, _actions_ the user can take with a unit.  Will typically (always?) correspond to a button.
 - Attributes.  These are properties of the unit.  King unit.  Generates leaders.  Starts golden age.  Etc.

There will likely be more eventually, e.g. AvailableTo, IgnoresMovementCost, CanMoveOnTerrain.  But they can be added as we implement them.

Sample JSON prototype:

```
      "Warrior": {
        "name": "Warrior",
        "shieldCost": 10,
        "populationCost": 0,
        "attack": 1,
        "defense": 1,
        "bombard": 0,
        "movement": 1,
        "iconIndex": 6,
        "categories": [
          "Land"
        ],
        "actions": [
          "hold",
          "wait",
          "fortify",
          "disband",
          "goTo"
        ],
        "attributes": []
      },
```

Of note, UnitWithAvailableActions in UnitInteractions is something I'm favoring refactoring as well, now that actions are primarily stored on the prototype.  We still need to return them, but it seems it would make sense to return a list of actions by itself.  However, to avoid scope creep on the refactor, and because the goal of the refactor was to allow back-end improvements (e.g. AI making decisions based on prototypes), I decided to leave that for later.


pcen (approved):
lgtm

WildWeazel:
I'm thinking give it a bit more structure. Not only does that group logically related attributes but it makes it easier to grab a whole set that may contain other keys later:
`"cost": {
  "production": 10,
  "population": 0
},
'movement": {
  "speed": 1,
  "ignoreCost": []
},
"bombard": {
  "strength": 0,
  "range": 0,
  "rate": 0
}`
etc.

WildWeazel:
Follow-up thought: we've tossed around the idea of direct JSON serialization, which would suggest the format should follow the class structure. In that case this becomes more of an architectural question...

QuintillusCFC:
I thought I'd replied to your comment on 4/25, but must have only thought about it.  My impression is that Ozymandias has been pushing for flatter structures to make it easier to add new properties.  That isn't necessarily incompatible with nesting, however, so long as we make it easy to grab nested properties with keys.  e.g. unitPrototype.get("bombard.strength") could return the nested strength attribute on the bombard group.

The main advantage I see in that is for self-documentation/readability.  Some of the Civ3 properties have similar names to each other, especially on unit prototypes, and I find myself having to look up which one is which occasionally even now.  Admittedly I don't really follow the goal of "makes it easier to grab a whole set that may contain other keys later".  When would that be beneficial?

For now, since there's an approval, and this is required for future Carthage goals (notably around the AI decision-making), I'm going to go ahead and merge that.  No opposition to a discussion/issue/etc. for potential future nesting.

PR #246, opened 2022-04-18
Tests to .NET 4.7.2/C# 8.0.

QuintillusCFC:
Fixes the issue WildWeazel was having running them locally.  Don't know why it worked for me with .NET 6.0, probably having more frameworks installed.

Looks like pcen already hotfixed most of this, so it's actually just the C# 8.0 change.

### DIFF
--- a/C7GameDataTests/C7GameDataTests.csproj
+++ b/C7GameDataTests/C7GameDataTests.csproj
@@ -1,9 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFramework>net6.0</TargetFramework>
+        <TargetFramework>net472</TargetFramework>
         <Nullable>enable</Nullable>
-
+		<LangVersion>8.0</LangVersion>
         <IsPackable>false</IsPackable>
     </PropertyGroup>
 
--- a/C7GameDataTests/C7GameDataTests.csproj
+++ b/C7GameDataTests/C7GameDataTests.csproj
@@ -3,7 +3,7 @@
     <PropertyGroup>
         <TargetFramework>net472</TargetFramework>
         <Nullable>enable</Nullable>
-		    <LangVersion>8.0</LangVersion>
+        <LangVersion>8.0</LangVersion>
         <IsPackable>false</IsPackable>
     </PropertyGroup>
 
--- a/C7GameDataTests/C7GameDataTests.csproj
+++ b/C7GameDataTests/C7GameDataTests.csproj
@@ -2,7 +2,6 @@
 
     <PropertyGroup>
         <TargetFramework>net472</TargetFramework>
-        <Nullable>enable</Nullable>
         <LangVersion>8.0</LangVersion>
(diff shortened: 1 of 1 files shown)


pcen:
Is there a reason for enabling nullable if we aren't using nullable reference type annotations?

QuintillusCFC:
I just removed that in the most recent commit.  Not that I know of.

WildWeazel:
If we're not using nullable do we still need 8.0? I think not, but by default I'd rather we use it everywhere.

pcen:
I don't think every csproj in the solution specifies the language version, so we may want to add c# 8 to those project files too

WildWeazel:
This builds for me but I propose adding to all projects. I don't know any reason not to use the latest version that Godot supports.

WildWeazel:
We had some chat related to this on discord and... yeeesh. I'm not sure what the takeaway is. Learn how Godot Mono works?

WildWeazel:
But seriously, we need to figure out which .NET implementation and/or target we're going with. I'm not sure if it's possible to build, run, and unit test with the same one on all platforms. Let's discuss it in #249

WildWeazel:
The plot, unfortunately, thickens. As explained in #249 **.NET 6 seems to be the right answer.** I'm not sure what configuration I had when I first said this branch builds for me, but now I'm having problems. I _can_ build, but only by either using MSBuild and Mono, _or_ by removing VSStub from the solution. I don't really know how VSStub works, but it's using a TargetFrameworkVersion of v4.7.2 instead of a TargetFramework of net472. That tag apparently only works with the Windows .Net Framework, which of course I can't install. Finally, if I change all projects to TargetFramework net472 as directed (and then add a suddenly necessary reference to System.IO.Compression), it insists that VSStub is somehow targeting .Net Framework v4.0 and can't reference C7 that targets net472. That makes no sense, and I'm really at my wit's end with this. I'm using VS Code so I don't actually need VSStub anyway. I tried to look for conditional or configurable project inclusion, but the answer seems to be maintaining multiple .sln files for different configurations. I guess that's doable, but can anyone else weigh in on what's going on? @QuintillusCFC @pcen and @JimOfLeisure for good measure

QuintillusCFC:
See also my comment on #249 .  tl;dr: I propose we deprecate Visual Studio 2019 support, and with that VSStub.  We can debug in VSCode now, and Rider also works well for those who want a more full-fledged IDE.

Removing VSStub should _probably_ be separate from this step PR, though.

WildWeazel:
I don't really have a stake in that since I won't be using VS2019 anyway, but that's what I did locally just to get past the errors. If it's not breaking anyone else's setup I support standardizing on VS Code with Rider as an option for those with a license. I _don't_ have IDE support currently after all my mucking around, so between that and the new LTS release I think once this is settled I'll start over in another VM with 22.04 and notes from last time, and upload a credential-less copy of that somewhere for people to use.

As far as this issue goes, it seems like the solution should be to use net472 and 8.0 in all projects.

WildWeazel:
Will follow up in #256

PR #250, opened 2022-05-05
Combat Part Two

### LINKED ISSUE #194: Units should heal when resting
Units should heal when they spend a turn not doing anything.

For now, since we don't have borders, that should apply anywhere on the map for land units, and in port for sea units.

Optionally, teach the AI to heal injured units too.  Or, spin that into a separate issue and just implement the core mechanic, allowing the human player (and city-guarding, non-moving AI units) to make use of it.

### LINKED ISSUE #197: Units with Defensive Bombardment should make use of that bombardment
Units that have a defensive bombardment should make use of it.

Right now, that's the catapult, and it already has the bombard attribute.  So this amounts to ensuring it gets to fire when another unit on the same tile is attacked.

Odds are computed by comparing the bombard strength to the attacker strength.  So if a unit with 1 attack point attacks a stack with a 4 defensive bombard strength unit, there is an 80% chance it will cause a hitpoint of damage.

Preferably, add a "lethal land bombard" flag (default to false) for units, so that it won't destroy the last hitpoint.  Even better, add a "min hitpoint remaining after bombard" attribute that is configurable (default to 1 for bombard units), so mods can set it differently based on different units.

### LINKED ISSUE #198: Retreat probability - units should have a "retreat" flag, probability initially based on experience levels
Units should have a retreat flag, which (for legacy compatibility) should default to yes for units with 2+ movement points.

Retreat probability for units with that flag should be based on experience levels, which should be imported from BIQ/SAV/JSON.

There must be a valid tile to retreat to if the defender is retreating.  I believe this means a tile that is not neighboring the unit that attacked, nor any other unit belonging to that civ.  But there are edge cases I'm not familiar with, so finding a reference for this at CFC would be a good idea.

maxpetul:
This PR finishes off "Combat Part Two", the second pass over the combat mechanics for Carthage. Major changes:
- Units heal at the start of each turn (closes #194). This is done in the new `MapUnit.OnBeginTurn` method, which also resets movement and flags.
- Units can now perform defensive bombard (closes #197). The implementation is pretty straight forward, it's mostly just a block of code in `MapUnit.fight` that produces the behavior.
- Units can now retreat from combat (closes #198). This required an overhaul of `MapUnit.fight` since it multiplies the possible combat outcomes.
- Added `CombatResult` and `CombatRole` enums for better abstraction in combat code. E.g: `MapUnit.fight` returns a CombatResult instead of a win/loss boolean. There's a new method `UnitPrototype.BaseStrength` that returns base strength for each combat role. The various `StrengthVersus` methods have been merged into one that's parameterized by combat role.

Programming this stuff made me realize there are many edge cases in Civ 3 where I'm not sure of the rules. For example, is it possible for a unit to retreat while defending a city? If a unit attacks with 1 HP, can it get killed by defensive bombard? In both cases, I think the answer is "no" since I can't recall ever seeing those things happen, but I'm not sure about that.


C7Engine/MapUnitExtensions.cs
```diff
@@ -102,53 +97,111 @@ public static void RollToPromote(this MapUnit unit, MapUnit opponent, bool waitF
...
+		// Defensive bombard
+		MapUnit defensiveBombarder = MapUnit.NONE;
+		double defensiveBombarderStrength = 0.0;
+		foreach (MapUnit candidate in defender.location.unitsOnTile.Where(u => u != defender && ! u.hasExhaustedDefensiveBombard)) {
+			double strength = candidate.StrengthVersus(attacker, CombatRole.DefensiveBombard, attacker.facingDirection.reversed());
+			if (strength > defensiveBombarderStrength) {
+				defensiveBombarder = candidate;
+				defensiveBombarderStrength = strength;
+			}
+		}
```
WildWeazel: eventually, it's probably worth converting all these search & compare cases to something more LINQ based (map/reduce, sort by, take first, and whatever else) and put them in utility functions

WildWeazel: eventually, it's probably worth converting all these search & compare cases to something more LINQ based (map/reduce, sort by, take first, and whatever else) and put them in utility functions

C7GameData/CombatResult.cs
```diff
@@ -0,0 +1,22 @@
+namespace C7GameData
+{
+	public enum CombatResult {
```
WildWeazel, line 3: Is it worth differentiating here between killed & captured, or whether the attacker advances? This seems adequate for now.

WildWeazel, line 3: Is it worth differentiating here between killed & captured, or whether the attacker advances? This seems adequate for now.

C7Engine/MapUnitExtensions.cs
```diff
@@ -183,6 +236,30 @@ public static void bombard(this MapUnit unit, Tile tile)
 		unit.facingDirection = unitOriginalOrientation;
 	}
 
+	public static int HealRateAt(this MapUnit unit, Tile location)
+	{
+		C7RulesFormat rules = EngineStorage.rules;
+		City city = location.cityAtTile;
+		bool inFriendlyCity = (city != null) && (city != City.NONE) && unit.owner.IsAtPeaceWith(city.owner);
+		return inFriendlyCity ? rules.healRateInCity : rules.healRateInField;
+		// TODO: Consider friendly/neutral/enemy territory once that's implemented, barracks, the Red Cross, and rules for naval units (can they
```
WildWeazel, line 245: heal rate in friendly/neutral/hostile field should be separately configurable

WildWeazel: heal rate in friendly/neutral/hostile field should be separately configurable

C7GameData/MapUnit.cs
```diff
@@ -31,6 +31,7 @@ public class MapUnit
 	}
 	public bool isFortified {get; set;}
 	//sentry, etc. will come later.  For now, let's just have a couple things so we can cycle through units that aren't fortified.
+	public bool hasExhaustedDefensiveBombard;
```
WildWeazel: don't assume 1 DB per turn :)

WildWeazel: don't assume 1 DB per turn :)

C7GameData/RulesFormat.cs
```diff
@@ -43,6 +43,9 @@ public class C7RulesFormat
 		public StrengthBonus cityLevel2DefenseBonus;
 		public StrengthBonus cityLevel3DefenseBonus;
 
+		public int healRateInField;
+		public int healRateInCity;
```
WildWeazel, line 49: really, there should be options for friendly, neutral, and hostile field, possibly in transport, and I suppose any other conditions (eg fortress) would be modifiers

WildWeazel, line 49: really, there should be options for friendly, neutral, and hostile field, possibly in transport, and I suppose any other conditions (eg fortress) would be modifiers

C7GameData/CombatResult.cs
maxpetul: I hadn't thought about capturing units, but we'll probably want to add that as a possible result. Differentiating whether or not the attacker advances at this point would be awkward since CombatResults are produced by `MapUnit.fight`, which isn't responsible for determining whether or not the attacker can advance. That's done in `MapUnit.move` after it checks the result. If we never end up using CombatResult for anything other than the return value from `MapUnit.fight`, it should be renamed to FightResult for clarity.

C7GameData/RulesFormat.cs
maxpetul: Do fortresses affect heal rate in the original game?

QuintillusCFC: They don't.  I agree that _long-term_, we should allow more flexibility.  However, this is "Combat Part II", so my opinion is we should write up what WildWeazel wrote as an "enhancement" issue, but not try to add it in here.  A big question will be which enhancements such as this do we prioritize, but they aren't all necessary for compatibility.

QuintillusCFC:
Taking a look at this PR, I'd been leaving it for @WildWeazel since he already did a pass, but that's been almost a month now.

```
Programming this stuff made me realize there are many edge cases in Civ 3 where I'm not sure of the rules. For example, is it possible for a unit to retreat while defending a city? If a unit attacks with 1 HP, can it get killed by defensive bombard? In both cases, I think the answer is "no" since I can't recall ever seeing those things happen, but I'm not sure about that.
```

I also think the answer is "no", 100% sure on not being able to retreat while defending a city, 98.1% sure on defensive bombard.  I also don't recall seeing either of them.

C7Engine/MapUnitExtensions.cs
QuintillusCFC: This appears to be resolved to me.

C7Engine/MapUnitExtensions.cs
```diff
@@ -183,6 +236,30 @@ public static void bombard(this MapUnit unit, Tile tile)
...
 	}
 
+	public static int HealRateAt(this MapUnit unit, Tile location)
+	{
+		C7RulesFormat rules = EngineStorage.rules;
+		City city = location.cityAtTile;
+		bool inFriendlyCity = (city != null) && (city != City.NONE) && unit.owner.IsAtPeaceWith(city.owner);
+		return inFriendlyCity ? rules.healRateInCity : rules.healRateInNeutralField;
+		// TODO: Consider friendly/neutral/enemy territory once that's implemented, barracks, the Red Cross, and rules for naval units (can they
+		// heal outside of port? I don't think so, but I might be getting confused with another civ game).
```
QuintillusCFC: Not in Civ3.  I think in Civ6 it's possible to heal outside of but near ports.

C7Engine/MapUnitExtensions.cs
```diff
@@ -102,53 +97,111 @@ public static void RollToPromote(this MapUnit unit, MapUnit opponent, bool waitF
...
+		double defensiveBombarderStrength = 0.0;
+		foreach (MapUnit candidate in defender.location.unitsOnTile.Where(u => u != defender && u.defensiveBombardsRemaining > 0)) {
+			double strength = candidate.StrengthVersus(attacker, CombatRole.DefensiveBombard, attacker.facingDirection.reversed());
+			if (strength > defensiveBombarderStrength) {
+				defensiveBombarder = candidate;
+				defensiveBombarderStrength = strength;
+			}
+		}
+		if (defensiveBombarder != MapUnit.NONE &&
+		    attacker.hitPointsRemaining > 1) { // TODO: Can defensive bombard kill? What does the original game do in this case?
```
QuintillusCFC: I tested this just now, and it's a good thing I did because what I had thought was the case was wrong.  Even if the lethal land bombardment flag is on (e.g. Hwacha), defensive bombard will not kill a unit.  I set up a test where all units have 1 HP, gave myself warriors next to a stack of Korean warriors and Hwachas, and the Hwachas didn't even fire.

C7Engine/MapUnitExtensions.cs
```diff
@@ -102,53 +97,111 @@ public static void RollToPromote(this MapUnit unit, MapUnit opponent, bool waitF
...
+			double dADB = attacker.StrengthVersus(defensiveBombarder, CombatRole.DefensiveBombardDefense, defensiveBombarder.facingDirection);
+			if (EngineStorage.gameData.rng.NextDouble() < defensiveBombarderStrength / (defensiveBombarderStrength + dADB))
+				attacker.hitPointsRemaining -= 1;
+
+			defensiveBombarder.defensiveBombardsRemaining -= 1;
+			defensiveBombarder.facingDirection = dBOriginalDirection;
+		}
+
+		// TODO: Check these rules are accurate and that we're not missing any.
+		bool defenderEligibleToRetreat = defender.hitPointsRemaining > 1 && ! defender.location.HasCity;
```
QuintillusCFC, line 160: There's at least one other case, which is that there has to be a valid tile to retreat to.  At one level, this means the unit can't be up against the sea, but it _also_ means the unit hasn't been fully flanked.  I'd have to check to be 100% sure of the rules, but my understanding has been that the tiles "behind" the unit after it is rotated to face its attack must not be occupied.  But all of them not be occupied, just one, depends on ZOC?  That's where it's tricky.

I propose writing up an issue for the check-for-a-valid-tile-to-retreat-to and referencing that issue in the TODO, since it could be tricky and probably belongs in Combat Part III.

QuintillusCFC: Ah, I see that this is at least somewhat covered on line 169.  We probably do still need to have a test to document the original game's rules on the full extent of this.

C7Engine/MapUnitExtensions.cs
```diff
@@ -102,53 +97,111 @@ public static void RollToPromote(this MapUnit unit, MapUnit opponent, bool waitF
...
+			dead.animate(MapUnit.AnimatedAction.DEATH, true);
+			dead.disband();
+		}
 
-		if (defender != loser)
+		if (result.DefenderWon())
 			defender.facingDirection = defenderOriginalDirection;
 
-		return unit != loser;
+		return result;
```
QuintillusCFC, line 208: Probably for the next set of improvements, but at over 100 lines of code and now covering several (albeit related) mechanics, I'm thinking it's about time to add helper methods.  E.g. defensive bombard could probably be broken out.  I found as I added more code to my editor that it's easier to find the relevant section for making updates if methods are split out at least somewhat, especially when using an IDE that gives an outline of the available methods.

C7Engine/MapUnitExtensions.cs
QuintillusCFC: Okay, it's at least mostly covered.  Only question remaining is whether a unit adjacent to both the unit and the tile it wants to retreat to prevents that retreat.  Well, and I guess if it can retreat at a "diagonal" if the tile behind it is not of a valid terrain type, but the tile behind-and-to-the-side is of a valid terrain type.

C7GameData/CombatResult.cs
QuintillusCFC: I think captured should be part of a future "Combat Part III".  It's an advanced rule, and IMO not as important as more basic things like adding technologies and buildings and map visibility.  Eventually, though, it probably should be a possible result.

QuintillusCFC (approved):
Left a few comments, but they are all things that can be done as part of Part 3, or saying, I agree with not having this (e.g. capturing units) yet.  Feel free to make changes based on them if you want, but only if you want.

I like how combat is turning out to be a pretty readable module.  There are some nice helper methods such as the BaseStrength one that, combined with the enums, help out quite a bit in that regard.

QuintillusCFC:
Oh and also thanks for leaving a good overview comment.  That really helped with orienting myself at the start.

C7Engine/MapUnitExtensions.cs
maxpetul: I opened issue #274 for the matter of perfecting defender retreat behavior, and updated that comment & a couple of others in light of your remarks about the combat rules. I hadn't even considered zone of control until you mentioned it, that's one of those little rules I always forget about until I get hit by it.

PR #252, opened 2022-05-15
Disable animations with shift

maxpetul:
This relatively small PR lets players disable animations by holding down the shift key. That way it's possible to click through the turns much more quickly. Also I noticed that the game becomes sluggish after 20 or so turns, and I was concerned that animations were responsible, so I implemented this to test if disabling them would improve turn times. It does not.


QuintillusCFC (approved):
I see how that works, with listening for the event key no longer being pressed.

Looks good to me and I'd also been thinking being able to skip animations would be nice.

PR #255, opened 2022-05-17
118 add a proper logging framework

### LINKED ISSUE #118: Add a proper logging framework

WildWeazel:
This is #178 by @pcen plus an additional commit by me to demo a simple Godot sink to route the logs to the editor output (this also prints to console when running from the command line, so I disabled the other sink). I'm not familiar with Serilog so this is just a quick proof of concept based on their sink tutorial and the Console sink. The comments on #178 may still be relevant.


C7/LogManager.cs
```diff
@@ -0,0 +1,68 @@
...
+using Serilog.Core;
+using Serilog.Events;
+using Serilog.Configuration;
+using Serilog.Formatting;
+using System;
+using System.IO;
+
+public class LogManager
+{
+	static LogManager()
```
WildWeazel: @pcen commented: "for any runtime initialisation that must be done before some feature is used, AFAIK we don't have an "entry point" for C7, so I think using a static factory class to ensure the initialisation happens first is the best way to solve this"

WildWeazel: I agree, at least for now. As mentioned in the other review if we're using a Godot sink that means the UI layer will have to be responsible for configuring the logger.

C7/LogManager.cs
```diff
@@ -0,0 +1,68 @@
...
+	{
+		ExpressionTemplate consoleTemplate = new ExpressionTemplate(
+			"{@t:HH:mm:ss.fff} [{@l:u3}]{#if SourceContext is not null} {SourceContext}:{#end} {@m:lj}\n{#if @x is not null}\tException: {@x}{#end}",
```
WildWeazel, line 21: @QuintillusCFC commented: "Can we put a sample statement above the format, or an explanation of the more complex elements ([{@1:u3}] and {@m:lj}, IMO) to make it easier to read what the format is trying to do? Log formats are rarely intuitive and this one isn't too bad, but it's still handy to have the reference right there."

C7/MainMenu.cs
```diff
@@ -1,10 +1,11 @@
 using Godot;
-using ConvertCiv3Media;
-using C7GameData;
 using System;
+using Serilog;
 
 public class MainMenu : Node2D
 {
+	static ILogger log = LogManager.ForContext<MainMenu>();
```
WildWeazel: @QuintillusCFC commented:
"Question: Is it possible to selectively enable logs?

For reference, I see it prints out:

12:37:32.429 [DBG] MainMenuMusicPlayer: setting volume to 100, which is 0 decibel offset

MainMenuMusicPlayer being the ForContext of that log.

I know some libraries, such as Log4J, have a hierarchical structure, where a fully-qualified class name is considered, such as com.civfanatics.xplatformeditor.map.rightSidePanel . This then allows turning on debug messages based on partially-qualified class name, such as "com.civfanatics.xplatformeditor.map" to turn on debugging statements for the map, but not for the whole application. I've only sparingly used that in practice, but it could be useful for debugging areas such as the graphics or AI, especially for hard-to-replicate events.

I see that we aren't using namespaces in the sample classes where it's enabled... so maybe the answer is that if we were it would be supported."

@TinaFemea replied:
"I totally agree with this. Having some way to show only the debug messages you're interested in is crucial."

@pcen replied:
"You can configure Serilog to filter by by namespace [like this](https://stackoverflow.com/a/38785632). With `ForContext`, the fully qualified class name is added to the log, but I believe it is just omitted from console output for brevity"

WildWeazel:
Added another commit to address a couple of comments and make it more Godot-oriented. This is ready for review/approval.

pcen (approved):
Thanks for adding the Godot sink and changing LogManager to be a Godot node, both look like nice changes. I guess next will be to change all our print statements to log calls 😄

WildWeazel:
#258 :)

PR #271, opened 2022-06-07
266 - Strategic AI Priorities

### LINKED ISSUE #266: Strategic AI Priorities - Part 1
As I started to work on #137 , I realized that I'd started adding the structure for not-exactly what that is (AI deciding what to build in cities dynamically), but rather the structure for a strategic high-level AI.

That isn't necessarily bad; they are both things I'd been thinking about how to add.  And the strategic AI _should_ play a role in what the AI builds in cities.  I wound up starting on it first because in my thoughts about how these should work, the strategic layer plays an almost critical role in deciding what to build.  There are local concerns, to be sure - don't build a Settler if the pop requirement will never be met, for example.  But a lot of the decision comes down to what is necessary to meet strategic objectives.  Do we need more cities?  More military units?  Better exploration?  I could make the city production choices a little better than the placeholder hard-coded selection without a strategic layer.  But it will be more robust, and soon more effective, if the strategic layer is looped in first.

Thus, this is:

 - Include the base structure for strategic AI priorities, including the general structure for how to choose between priorities.
 - Include at least two strategic AI priorities.  Expansion will be one of them; there will be another simply to provide an example/secondary priority.
 - Add a blurb on the Wiki about this.  While nothing is final at this stage, I think this will be one of the first areas where the AI has some long-term vision (versus mostly being an experimental placeholder).  Pathing is the other area that's kind of at that stage.

This should lead into #137 nicely.

Longer-term, I also envision strategic priorities being a pluggable component, e.g. a mod could add one.  We'll see how this all works (in particular, city selection and about 42 million other places eventually having to be able to query the strategic priorities to see if they should affect decisions being made).  But if we can get things aligned to make that happen, I think it opens up a fair amount of moddability when it comes to AI behavior, with only a subset of it needing to be modified to get those effects.

QuintillusCFC:
This is the first pass of strategic AI priorities.

As written, the AI will choose 2-4 strategic priorities every 15-24 turns, out of a possible 3 types that have been implemented (in practice, this means it will never choose 4).  For details on the concept, see the extensive comments included in this branch, or [this Wiki page](https://github.com/C7-Game/Prototype/wiki/Design:-AI).

The design is meant to be extensible, with a common interface for StrategicPriority implementations, and PriorityAggregator standing in as a class that in the future would find all registered StrategicPriority implementations and make them available to StrategicPriorityArbitrator, which uses the common interface to allow AI players to select their priorities.

Future work will also be required to persist strategic priorities in saves.  See my "Discussions" post about Newtonsoft as a JSON API that will get us that feature with no per-implementation work.  I think that is the best bang-for-our-buck given that writing custom converters would tax our most in-demand resource (coding) a fair amount, and the only significant tradeoff is that Newtonsoft is not as fast as Microsoft's built-in version.

Finally, note that right now all the priorities are used for is printing out to the console information about which one is chosen.  I plan to start using them in my next change set, about the AI choosing what to build in its cities in a more dynamic, informed fashion.

Closes #266


QuintillusCFC:
It's been 16 days, so I resolved the conflicts with Development, did a self-review, and am going to merge.  I have too much stuff in progress and want to rebase my latest changes, which will be easier if Development is current.

PR #272, opened 2022-06-09
137 - Dynamic City Production

### LINKED ISSUE #137: AI Decides What To Build Based On Game Circumstances (Part I)
Instead of building things in a fixed order, the AI should consider game circumstances.

This will of course be something that is revisited a lot as new features are added.  Notably, we don't have buildings yet, and only have a handful of units.  That said, some possibilities:

 - AI starts building settlers if the city will grow in time, even if it's size 1 or 2
 - AI stops building settlers when it's out of land
 - AI tries to consider its needs when building units.  Does it need an explorer unit?  Maybe build a fast Chariot.  Does it need defenders?  Cheap Warriors are a better option.  Does it already have a bunch of Galleys?  Maybe it's time to stop building them.

Given how few options there are (and how there isn't an option for "stronger defender" or "attack unit" or "boost the economy"), I don't expect this to be very impressive right away.  But it should be a slight improvement, and set the stage for the future.

Preferably, componentize this so the AI is extensible.  For example, if the AI determines it needs defenders, it should call a module to figure out what type of unit to build as a defender.  Right now that might be simple, but it sets the stage for expanding it later on.  For example, if it needs a defender for a mountain tile, that module can be enhanced in the future so it doesn't build a unit that can't travel on mountains.

QuintillusCFC:
This continues buildout of AI modules, this time on city production.  Closes #137 .

Previously, city production was based on a hard-coded, very simple, non-extensible tree.  Now, it will look at all possible units that can be built, and for each unit consider, in order:

 - A base weight calculation.  Better core stats (attack, defense, movement, etc.) improve this; higher costs decrease it.
 - Additions/subtractions based on the strategic priorities.  For example, if expansion is a priority, units that can build cities get a buff here.  This is intended to make sure units that are not attractive in general (0/0/1 settler for 30 shields) will not have a zero/negative weight if the strategy dictates building them.
- A zeroing out of the score for "invalid" units, e.g. sea units in non-sea cities, settlers when the population is too small (and will not grow in time).
- A multiplicative adjuster based on strategy.  This is intended to make the strategy's preferred units much more likely to be chosen.

At this point, each unit prototype has a calculated weight, which is then adjusted by a quadratic weighting system.  Finally, based on these weights, a unit is selected to be built.

While I am sure this will evolve, this should put in place a general framework to which new units can be added and automatically built, and new strategies can be added which impact how the AI considers what to build.

Longer-term (likely post Carthage), there will also be other factors considered, e.g. build often/never weighting, military strategy (does this AI prefer cavalry?  artillery?), how much the terrain needs to be improved.  In theory, these can be consulted during the loop structure being added here.  We'll also have to add in building support.


QuintillusCFC:
Opened as a draft as there's still a couple places where I want to refactor/deduplicate code.  But now it's visible for those curious; I might also switch gears and see about building out logging to more layers as the lack of granular logging control was a bit of a pain point while working on this.

QuintillusCFC:
Took care of the todo items for this.  Since there haven't been any reviewers for two weeks, I'm going to go ahead and merge.

PR #275, opened 2022-06-12
Merge C7RulesFormat into C7GameData

### LINKED ISSUE #247: Fate of C7RulesFormat
Quintillus and I were discussing this here: https://github.com/C7-Game/Prototype/pull/236#discussion_r851700459. I think we're in agreement that it would be best to do away with C7RulesFormat by merging it into C7GameData. If anyone has a good reason to keep C7RulesFormat around, please share it here. Otherwise, this issue will be for refactoring it away.

maxpetul:
Resolves #247
This PR is pretty simple even though it changes several hundred lines. All it does is move the contents of C7RulesFormat into C7GameData, updates several dozen references to point to the new locations, and cleans up the now-empty C7RulesFormat class & objects.


WildWeazel (approved):
I want to get away from relying on static access to data and use dependency injection instead, but that's a larger issue. This seems like a step in the right direction.

Issue #267, opened 2022-05-31
Add Serilog to Projects other than C7


QuintillusCFC:
#118 did the hard work for getting a logging framework.  But it's only in the C7 project.  We should have it everywhere.  I'm particularly guilty of adding a lot of Console.WriteLine statements in C7Engine, but this applies to all of them.

This might be a good first issue, as it should be possible to emulate the existing changes in C7.  Though I'm not 100% sure there won't be any complicating issues.

WildWeazel:
This raises a question that I was going to ask for the file reorg: do we need LogManager.ForContext<T>? I think not, as it's just an alias and Serilog.Log is already static. But I ask because other projects need to not use LogManager, because it depends on Godot.

QuintillusCFC:
> This raises a question that I was going to ask for the file reorg: do we need LogManager.ForContext? I think not, as it's just an alias and Serilog.Log is already static. But I ask because other projects need to not use LogManager, because it depends on Godot.

So, looking at #255 and #178 , I think the answer is "it's complicated".  I'm not the .NET expert, but my understanding is the benefit of Log.ForContext (in Serilog) is that it includes the "context", e.g. the type logging the message, so logs can be selectively enabled/disabled by class (e.g. only enable debug messages for components related to the audio if we're debugging an audio bug).  So we do want the Serilog ForContext.

But the LogManager appears to be a helper class that integrates the logs with the Godot sink.  You are right that it can't be used in other projects.

There were already some oddities about logs from projects outside C7, namely that if you run from Godot, they don't show up in the log within the Godot engine window (they do in the command prompt window that starts alongside Godot, however... also this may be completely different on platforms other than Windows).  I think they do show up properly in Rider, but the EngineTests project is currently broken, so I can't compile to test that.  Not sure about VSCode.  But my conclusion is that there may be extra hoops to jump through to get non-Godot-project logs showing up in Godot's console output.

But that's already a hazard; I would put "enable Serilog and its benefits of logging levels and ForContext filtering" above "make the logs show up in the Godot output window", which is merely a convenience.  That's a nice thing to have eventually, but at least for me not a significant pain point, just a minor nuisance.

tl;dr: We should use Serilog's ForContext in the other projects, but they obviously can't be tied to Godot.

QuintillusCFC:
[...]

Issue #259, opened 2022-05-22
File-based logging config


WildWeazel:
Allow the user to configure logging from a file without rebuilding. Suggest using Serilog.Settings.Configuration: https://github.com/serilog/serilog-settings-configuration
Ideally you would be able to specify:
Whether to create log files, where, and max size
Namespace filters
Minimum level (per namespace?)

PR #276, opened 2022-06-13
#267 - Add Serilog to C7Engine

QuintillusCFC:
And replace all Console logs in that project with it.

Used what I thought was the appropriate level.  We can shift them as time goes; I expect we'll generally be demoting logs from info to debug (or verbose) as more is added.

Also set the default minimum level to "info".  We should be selectively enabling debug for certain areas, which I plan to test next.

Opening this PR even though it's only the Engine part of propagating Serilog, since the sooner we get it to more modules the fewer Console logs we will have to convert.  I might have more modules added soon, but if not at least it'll be in one.

### DIFF
--- a/C7/LogManager.cs
+++ b/C7/LogManager.cs
@@ -13,9 +13,18 @@ public class LogManager : Node
 		ExpressionTemplate consoleTemplate = new ExpressionTemplate(
 			"{@t:HH:mm:ss.fff} [{@l:u3}]{#if SourceContext is not null} {SourceContext}:{#end} {@m:lj}{#if @x is not null}\tException: {@x}{#end}");
 
+		// You can filter this several ways with the expression in Filter.ByIncludingOnly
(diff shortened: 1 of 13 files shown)


WildWeazel (approved):

WildWeazel:
Regarding your newest commit, I think that would be covered by #259. @TinaFemea I know you haven't used Serilog, but generally the approach would be to create your logging sinks (outputs) to log everything and then configure your filters at runtime to capture what you want, right?

TinaFemea:
Yep!  Log everything, then filter out what you don't want.

QuintillusCFC:
Yeah, for that last commit I had a feeling there was a better way to do it; the way I found felt a bit clumsy.  Ironically I don't think I've ever actually set up namespace-based log filtering before.  At work we usually had something like Splunk to filter the logs (and if not, we just had lots of logs to wade through), and with projects like my editor I've just tweaked things manually when needed.

I also hadn't seen #259 yet.  Glad we have a card for it.  Going to go ahead and merge this one, and probably swing back around to that later if someone else hasn't first.  This was kind of a side quest from my other tasks, which will benefit from it, so I want to wrap them up next.

PR #280, opened 2022-06-23
Log to a file by default

### LINKED ISSUE #278: Figure out why Serilog logging is slow
As part of investigating #273, I determined that the logging I was adding to get a better idea of what was taking so long was actually slowing things down, in some cases significantly more than what I was investigating in the first place.  This investigation has now gone deep enough that it warrants being split off into its own issue.

This is good to know sooner rather than later, as it seems to be much slower than Log4J in the Java world, and if we are going to switch to e.g. Log4Net or another .NET solution, it's easier to do so early.

QuintillusCFC:
This is much quicker than writing to the Godot sink, which in turn writes to the console.

This allows investigating of performance hot spots without logging of the performance taking up more overhead than the actual performance hot spot.

See the extensive notes on Closes #278 .  When investigating why turn times were slow, the logging I added made things another order of magnitude slower.  These changes at least significantly decrease that overhead.

### DIFF
--- a/C7/C7.csproj
+++ b/C7/C7.csproj
@@ -16,6 +16,7 @@
   <ItemGroup>
     <PackageReference Include="Serilog" Version="2.10.0" />
     <PackageReference Include="Serilog.Sinks.Console" Version="4.0.1" />
+    <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
     <PackageReference Include="Serilog.Expressions" Version="3.3.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


WildWeazel:
I'm not ready to close #278 (even though I just repeated the magic words) but this a step forward for now. Another reminder that #259 should eventually supersede this.

WildWeazel (approved):

QuintillusCFC:
Merging in.  For some reason I can't clear the link to #278 on the sidebar, so I'll move it back manually.

QuintillusCFC:
I also agree that it's a step forward, not a solve everything forever.

QuintillusCFC:
Adding "Carthage" tag since the main issue has been split into a part 1 and part 2, linked to the same issue since it has the notes.

PR #281, opened 2022-06-23
273 - Fix slow explorer pathing

### LINKED ISSUE #273: Figure out why turns take longer as the game goes on, and fix it
Flintlock mentioned semi-recently that after about turn 20 things slowed down, and it sure does.  I've started investigating this, and found that galleys are always slow, and they're slow due to trying to find a path to their exploration destination.  Probably related to this, they can't move currently.  But sometimes other units are slow, too.  Log sample:

```
-> Begin Rome turn
Unit processing time for Chariot: 0 ms
Pre found new path time = 0 ms
Found new path time = 633 ms
Moved to new path time = 633 ms
Unit processing time for Galley: 633 ms
-> Begin Babylon turn
Unit processing time for Chariot: 0 ms
Set NEAR_CITIES exploration AI for Catapult with 1 movement points and 3 hit points, guid = 7130dd5f-c053-468f
-8257-2241ac68574b
Unit processing time for Catapult: 245 ms
Unit processing time for Warrior: 0 ms
Unit processing time for Chariot: 0 ms
Unit processing time for Warrior: 0 ms
Unit processing time for Chariot: 0 ms
Unit processing time for Warrior: 230 ms
Unit processing time for Chariot: 0 ms

```

I've added some logging on the SlowTurnsInvestigation branch to produce this, using the Stopwatch class for the timings.  200+ ms for one unit is obviously too slow.

This is tested by fortifying all my units at game start, and then pressing down both the Enter and Shift keys at the same time.  Enter keeps turns rolling, Shift skips all the animations.  So the only thing it's doing is processing the mechanics/AI.

The catapult happens to be bombarding a city in this log, so it's possible that some of the slowness is when combat occurs (even if animations are shift-skipped).  But the galleys are definitely pathfinding-related or pathfinding-adjacent.

(This is also an area where nifty logs might be useful.  For example, if we detect a unit taking longer than, say, 100ms, we should increase logging so we can investigate why (although this depends on having useful logging in the first place.  But at least a warning would be nice regardless).  Although we'd have to stop the timer for combat animations)

QuintillusCFC:
Closes #273 

There are two main optimizations here:

 - If we're pathing to water, don't even bother.  At some point we will want to bother, but since the algorithm doesn't support water yet, this prevents it from wasting time.
 - More importantly, the ExplorerAI now optimizes its calls to pathing.  It sorts all tiles it wants to query by crow-flies distances first, which is fast.  Then it starts calling them, but if it has already a tile whose path is as short or shorter than the crow-flies distances for the tile it's evaluating, then it skips it because it's impossible for the new one to be quicker.  Because it sorted by crow-flies distance first, this means that in most cases, most tiles will be skipped (the exception being if terrain features, namely water, prevent the shortest crow-flies tiles from being navigable without a lengthy detour).
 - As a corollary of the above, if it finds a path with length 1 (the shortest possible path, i.e. a tile right next door), it immediatley stops searching.

In my sample data, with a Babylon Warrior on turn 12, this takes the time down from 86ms to 1ms - very noticeable to not noticeable at all.  (Note that this was testing with a fixed seed of 123, which is not included in this branch)

Other notes:

 - This change relies on a high-level optimization to pathing, i.e. not calling the algorithm as much.  It's also possible to optimize the algorithm, either to make it faster, or to cache things so calls that would otherwise recalculate things don't have to recalculate things.  I suspect we'll use both of these options, but in this case the high-level one seemed a bit simpler, and proved to be quite effective.
 - I intentionally added some logging around this, of various levels, with the idea being to make it easier to notice if/when explorer pathing becomes slow in the future.  Longer term, I think we should enhance our pathing algorithm to keep metrics such as when it is called, how long it takes, etc. - as it's likely to be a major bottleneck if Civ3 is any indication.
 - See #273 for lots of detailed notes, including testing results.

Even with slow console logging and debug logging of explorer info (and with a few long calculations around bays), this gets us to turn 100 without the slowdowns we were seeing before.  Switching to file based logging extends that to 200 turns, at which point we're back to Aztec where it's likely the UI that starts slowing things down with tons of cities present.

### DIFF
--- a/C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
+++ b/C7Engine/AI/Pathing/DijkstrasLandAlgorithm.cs
@@ -40,6 +40,10 @@ namespace C7Engine.Pathing
 		}
 
 		public override TilePath PathFrom(Tile start, Tile destination) {
+			if (!destination.IsLand()) {
+				return TilePath.NONE;
+			}
+
 			// shortest distance from start to each tile on the continent
 			Dictionary<Tile, int> dist = new Dictionary<Tile, int>();
 			Dictionary<Tile, Tile> predecessors = new Dictionary<Tile, Tile>();
--- a/C7Engine/AI/UnitAI/ExplorerAI.cs
+++ b/C7Engine/AI/UnitAI/ExplorerAI.cs
@@ -54,7 +54,7 @@ namespace C7Engine
 				log.Information("No valid exploration locations for unit " + unit + " at location " + unit.location);
 				return false;
 			}
-			// log.Verbose($"Exploring for unit {unit}");
+			log.Verbose($"Exploring for unit {unit}");
 			KeyValuePair<Tile, float> topScoringTile = FindTopScoringTileForExploration(player, validNeighboringTiles, aiData.type);
 			Tile newLocation = topScoringTile.Key;
 
--- a/C7Engine/AI/UnitAI/ExplorerAI.cs
+++ b/C7Engine/AI/UnitAI/ExplorerAI.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using C7GameData;
 using C7GameData.AIData;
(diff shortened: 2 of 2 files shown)


QuintillusCFC (commented):
Self-reviewed since it's been a while.  Still makes sense.  Going to merge since this will significantly speed up mid-game turns for future work.

PR #283, opened 2022-07-05
282 - Fixed Seed

### LINKED ISSUE #282: Allow setting a fixed random seed
Setting a fixed random seed is very helpful for investigating performance, as well as bugs, as it allows reproducible behavior.

We're at a stage where this is useful, and should be added.

I'll add a commit that has the first stage of this.  For some reason I have to set the seed in 3 places, which isn't ideal, so more investigation is needed.  It also should preferable be configurable somewhere other than being hard-coded.

QuintillusCFC:
Closes #282 

This allows setting a fixed seed with one line of code being uncommented in GameData.cs.  Eventually, this will be configurable through the UI, but for now it allows debugging performance or AI issues with much more repeatability.

As a note, I made it static because somehow, passing the RNG as a parameter resulted in getting different results with the passed-by-parameter RNG each time I ran the game.  I'm not entirely sure why that was the case but it was definitely giving different results each time.  Overall however, "one RNG to rule them all" does seem like the sensible way to go to ensure we don't introduce another one somehow and throw off the eventual fixed RNG setting.

### DIFF
--- a/C7GameData/GameData.cs
+++ b/C7GameData/GameData.cs
@@ -34,7 +34,7 @@ namespace C7GameData
 		public GameData()
 		{
 			map = new GameMap();
-			rng = new Random();
+			rng = new Random(123);
 		}
 
 		public List<Player> GetHumanPlayers() {
--- a/C7GameData/GameData.cs
+++ b/C7GameData/GameData.cs
@@ -6,7 +6,7 @@ namespace C7GameData
 	public class GameData
 	{
 		public int turn {get; set;}
-		public Random rng; // TODO: Is GameData really the place for this?
+		public static Random rng; // TODO: Is GameData really the place for this?
 		public GameMap map {get; set;}
 		public List<Player> players = new List<Player>();
 		public List<TerrainType> terrainTypes = new List<TerrainType>();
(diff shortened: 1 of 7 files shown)


C7GameData/GameData.cs
```diff
@@ -6,7 +6,7 @@ namespace C7GameData
 	public class GameData
 	{
 		public int turn {get; set;}
-		public Random rng; // TODO: Is GameData really the place for this?
+		public static Random rng; // TODO: Is GameData really the place for this?
```
WildWeazel, line 9: IMO it's the right place for the _seed_ but not the RNG itself. I'm not really sure where it belongs though. Any time you're tempted to make an object static that should suggest a Component, but maybe that's not the best either. I'm kind of thinking it should be initialized somewhere in C7Engine (maybe CreateGame) and passed down by DI.

WildWeazel (approved):

C7GameData/GameData.cs
QuintillusCFC: I like that thought and distinction!  I probably will refactor this to storing the seed, and then having a RandomComponent, having properly looked at the component-event PR.  Might take a slight detour to investigating the CxC city placement first though, since that's been on my mind for a while now.

QuintillusCFC:
On second thought, I have CxxC working locally now, and in order to reduce conflicts and because the Component PR is not yet merged, I'm going to go ahead and merge this.  Then we can add the RandomComponent once that one is merged.  Otherwise, by keeping this branch open, I'll be building on it with other branches, potentially accumulating more conflicts.

PR #290, opened 2022-07-07
93 - CxxC

### LINKED ISSUE #93: Settler AI - Don't settle in CxC by default
Right now, the settler AI settles in CxC.  This shouldn't be the default; tiles that close to other cities should be skipped.

I would suggest making this configurable via a boolean; for some AI strategies (100k culture, Feudalism government type unit support, for example), it might make sense.

### LINKED ISSUE #207: AI settlers should favor tiles that are explored over tiles that aren't explored
In Babylon, without the Tile Visibility concept, the AI evaluated all nearby tiles when deciding where to settle.

We should change it so that the AI only considers tiles that have been explored, unless they have done a very poor job of exploring, or the only known tiles are so unappealing that taking a chance on random settler exploration is worth it.

WIth the TileKnowledge object on the Player object, it should be fairly easy to add this.  This will also allow removing the awkward method for finding nearby tiles.  However, we will likely need to add or strengthen a factor for how far away a tile is - we don't want the first settler walking halfway across the map because there's a cow over there.

QuintillusCFC:
This changes the city placement logic so there will always be at least two tiles between cities (CxxC pattern, rather than CxC, although the AI may choose a wider pattern if it deems it beneficial).

Closes #93 .  Closes #207

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -401,6 +401,9 @@ public class Game : Node2D
 						if (tile.cityAtTile != null) {
 							City city = tile.cityAtTile;
 							GD.Print($"  {city.name}, production {city.shieldsStored} of {city.itemBeingProduced.shieldCost}");
+							foreach (CityResident resident in city.residents) {
+								GD.Print($"  Resident working at {resident.tileWorked}");
+							}
 						}
 
 						if (tile.unitsOnTile.Count > 0) {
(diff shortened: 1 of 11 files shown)


QuintillusCFC:
The Greek, Dutch, and American AIs doing a nice job of settling in a CxxC pattern:

image

They've also done a decent job prioritizing city locations.  They tend to travel farther away early on than a human would, but the locations they haven't settled - jungle and plains - would be low-priority for a human, too.  The Dutch just beat the Americans in a race to The Hague's location.

QuintillusCFC:
Closes #93 

Closes #207 (this wound up being one line)

QuintillusCFC:
It's been a week, no reviews, it's also released to prod, no complaints.  Looked over again, going to merge.

PR #300, opened 2022-07-10
use 'allowCities' from terrain data

### LINKED ISSUE #120: Import/Use Civ3 Data about Whether Terrain Types Allow Cities
Currently, cities can be founded on any terrain that isn't Mountains, and this is hard-coded.

In reality, this data is stored in a BIQ file and is configurable; I've personally modified it in many of my scenarios.  We should import this into C7 and make it something that the AI considers when building cities - as well as something that controls whether the "Build City" option is enabled for the human player, or not.

Kright:
I add "allowCity" to C7 TerrainType. This lead to some fixes:

* update terrain types in .json map
* update settlerAI
* add methods to Tile and Unit
* add check to "buildCity" action

Closes: #120

### DIFF
--- a/C7Engine/AI/UnitAI/SettlerLocationAI.cs
+++ b/C7Engine/AI/UnitAI/SettlerLocationAI.cs
@@ -49,14 +49,8 @@ namespace C7Engine
 		private static Dictionary<Tile, int> AssignTileScores(Tile startTile, Player player, IEnumerable<Tile> candidates, List<MapUnit> playerSettlers)
 		{
 			Dictionary<Tile, int> scores = new Dictionary<Tile, int>();
-			candidates = candidates.Where(t => !SettlerAlreadyMovingTowardsTile(t, playerSettlers));
+			candidates = candidates.Where(t => !SettlerAlreadyMovingTowardsTile(t, playerSettlers) && t.IsAllowCities());
 			foreach (Tile t in candidates) {
-				//TODO: #120: Look at whether we can place cities here.  Hard-coded for now.
-				//Also likely move this into a "Where" clause above the loop so this loop is just the scoring/selection
-				if (t.overlayTerrainType.Key == "mountains") {
-					continue;
-				}
-
 				int score = GetTileYieldScore(t, player);
 				//For simplicity's sake, I'm only going to look at immediate neighbors here, but
 				//a lot more things should be considered over time.
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -398,8 +398,18 @@ public static class MapUnitExtensions {
 		}
 	}
 
+	public static bool canBuildCity(this MapUnit unit)
+	{
+		return unit.unitType.actions.Contains("buildCity") && unit.location.IsAllowCities();
+	}
+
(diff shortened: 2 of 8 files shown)


C7Engine/AI/UnitAI/SettlerLocationAI.cs
```diff
@@ -102,7 +102,7 @@ private static HashSet<Tile> GetCandidateTiles(Tile start)
 			Dictionary<Tile, int> scores = new Dictionary<Tile, int>();
 			foreach (Tile t in candidates) {
 				//TODO: #120: Look at whether we can place cities here.  Hard-coded for now.
-				if (t.overlayTerrainType.Key == "mountains") {
+				if (!t.IsAllowCities()) {
```
QuintillusCFC: The todo on line 104 can now be removed.

C7Engine/MapUnitExtensions.cs
```diff
@@ -398,8 +398,18 @@ public static void disband(this MapUnit unit)
...
 
+	public static bool canBuildCity(this MapUnit unit)
+	{
+		return unit.unitType.actions.Contains("buildCity") && unit.location.IsAllowCities();
+	}
+
 	public static void buildCity(this MapUnit unit, string cityName)
 	{
+		if (!unit.canBuildCity()) {
+			log.Warning("can't build city here");
```
QuintillusCFC: It might be helpful to also include unit.location in the warning message, e.g. $"can't build city at {unit.location}".  This shouldn't happen, but that way if it does it'll be easier to figure out where it's happening.

C7GameData/TerrainType.cs
```diff
@@ -17,6 +17,7 @@ public class TerrainType
         public int baseShieldProduction {get; set; }
         public int baseCommerceProduction {get; set; }
         public int movementCost {get; set; }
+        public bool allowCities { get; set; }
```
QuintillusCFC: Minor, but it might be nice to manually set the getter so that instead of `IsAllowCities` it reads something like `AllowsCities`.  This is a side effect of how C# configures properties by default and @WildWeazel might have a better solution as the C# expert, but in this case the "boolean getters start with `is`" rule makes it read awkwardly.

QuintillusCFC (commented):
Looks good, a couple minor comments but the core logic and style looks good to me.  I'm going to fire it up and run it once before giving full approval.

One other thought is that currently, the random start location generator can put starts on mountains.  That is in GameMap.cs, generateStartingLocations method.  There is a line `if (randTile.baseTerrainType.isWater())` that is probably the right place to check whether the tile allows cities.  As-is, my guess is civs that spawn on mountains/volcanoes will trigger the "can't build city" warning every turn, or just sit idle.

QuintillusCFC:
Ran the code, confirmed it's all working right other than that players can spawn on mountains/volcanoes.

(As a side note, the `generateDummyGameMap` method that was updated isn't currently used; it was used for our early random map generation before we realized it was a lot quicker to get maps working via importing them.  It might be used again someday, so it hasn't been deleted, just marked obsolete)

Kright:
There was conflict, so I rebased on current Development branch.

QuintillusCFC (approved):

QuintillusCFC:
Looks good!  Approved and merging!

PR #302, opened 2022-07-11
Set the initial zoom on the slider to 1.0, rather than 0.3

### LINKED ISSUE #301: The initial zoom-out takes you way far out
I got so used to this bug that I forgot it was a bug...

1. Start a new game.
2. Zoom out (tested with both mouse wheel and touchpad, specifically touchpad with zoom in/out along the right side)

Expected:

You are zoomed out a little bit.

Actual:

You are zoomed out so far you might as well be on the Moon.

If you tap "z" to restore 100% zoom, and then zoom out one wheel notch, then you only get zoomed out a little bit, as expected.  So it's a problem on initial zoom only.

QuintillusCFC:
This fixes the initial zoom-out (or zoom-in) so zooming out takes you to a 0.9 zoom level instead of a 0.2 zoom level (and 1.1 instead of 0.4 for zooming in).

Longer term, we probably shouldn't store this on a component that isn't going to be here long term, but it's still the only way to get back to the main menu, so for now I'm okay with a quick fix here and a refactor when we remove the component later.

Closes #301

### DIFF
--- a/C7/C7Game.tscn
+++ b/C7/C7Game.tscn
@@ -150,7 +150,7 @@ size_flags_vertical = 3
 min_value = 0.1
 max_value = 2.0
 step = 0.1
-value = 0.3
+value = 1.0
 
 [node name="QuitButton" type="Button" parent="CanvasLayer/SlideOutBar/VBoxContainer"]
 margin_top = 448.0


QuintillusCFC:
We've hit the 7 day limit, the one-liner still looks sensible and no complaints on the release, so I'm merging and moving to Released.

PR #303, opened 2022-07-11
Only allow selecting units belonging to the current player

### LINKED ISSUE #296: Clicking on one of your units that is fortified should select it
If you fortify a unit, currently you have to right-click to select it.  A regular left-click should work.  This worked at one point, but doesn't anymore.

### LINKED ISSUE #297: Babylonian units can be selected, and even ordered around!
May be related to #296 , or refactors for who the player is controlling.

You can select a Babylonian unit with a left click, and order it to move, with either the buttons or the num pad!

This only applies to Babylon, and is likely an artifact of earlier versions of unit selection.

QuintillusCFC:
Not units whose color matches that of Babylon.

Closes #296 , closes #297

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -359,7 +359,7 @@ public class Game : Node2D
 							if (tile != null) {
 								MapUnit to_select = tile.unitsOnTile.Find(u => u.movementPointsRemaining > 0);
 								//TODO: Better check for "current/human player"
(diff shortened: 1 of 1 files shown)


WildWeazel (approved):
is this adequate to remove the TODO now?

QuintillusCFC:
Good call, I mentioned the same thing in a PR review and then forgot it myself 45 minutes later.

PR #310, opened 2022-07-17
Build Server Uses .Net 6.0, and DRY Out GitHub Actions Code

### LINKED ISSUE #307: Build server can't build with tests at net6.0
Currently, the build server fails if the tests use net6.0, likely because its Docker image uses Mono, which imitates net4.7.2.

We should figure out a way to fix this, maybe:

 - Add a step to fetch net6.0 and build using that?
 - Figure out why the tests need to be on net6.0, and find a way to go back to net4.7.2
 - Something else?

### LINKED ISSUE #309: DRY out the GitHub Actions Configuration
Our Windows, Linux, and Mac build steps are about 83% shared (based on comparing them with KDiff and visually estimating the percentage).  However, they're all rewritten from scratch.  The more that is added to them, the more the overhead is noticeable, and the more tools like KDiff are needed to ensure we don't fat-finger something.

This is not great.  We should try to DRY out the configuration.

After some DuckDuckGo'ing and Googling, I think the best option may be Composite Actions.

StackOverflow on the topic: https://stackoverflow.com/questions/59757355/reuse-portion-of-github-action-across-jobs  The 3rd answer, by Cardinal, has a nice breakdown of the options and relative advantages.  I'm pretty sure we don't want reusable workflows, as we only want parts of the workflows (steps) to be reusable.  Strategies miiiight work but I'm not sure we have quite enough shared.  Composite Actions seem like the right balance.

Blog post on the topic: https://colinsalmcorner.com/github-composite-actions/  Uses this GitHub repo as reference: https://github.com/dotnet-architecture/eShopOnContainers/tree/dev/.github/workflows  And [this commit](https://github.com/dotnet-architecture/eShopOnContainers/commit/e57a97d5153e577c539129dc9054e40ee18aede7) shows exactly what happens when you add Composite Actions.  Look at basket-api.yml and .github/workflows/composite/build/action.yml in particular.

QuintillusCFC:
Closes #307 

Closes #309 

GitHub Actions isn't the world's most advanced CI product, but I still managed to DRY it out after some research on how to do that.  See #309 for more details about that.  This currently uses Composite Actions, which seemed like the most appropriate tool for the job, although I'm still not entirely convinced just building for all 3 platforms in one job wouldn't be a superior (if perhaps slightly slower) approach.

### DIFF
--- a/.github/workflows/composite/buildSetup/action.yml
+++ b/.github/workflows/composite/buildSetup/action.yml
@@ -1,0 +1,39 @@
+name: "Cross-Platform Setup"
+description: "Performs cross-platform build setup"
+
+runs:
+  using: "composite"
+  steps:
+  - name: Optionally set DEBUG_BUILD
+    if: github.event.inputs.debug-build == 'true'
+    run: |
+      echo "DEBUG_BUILD=-debug" >> $GITHUB_ENV
+  - name: Optionally set GH_USERNAME
+    if: github.event.inputs.include-username == 'true'
+    run: |
+      echo "GH_USERNAME=-${{ github.actor }}" >> $GITHUB_ENV
+  - name: Optionally set BUILD_TIME
+    if: github.event.inputs.include-datetime == 'true'
+    run: |
+      echo "BUILD_TIME=-$(date --iso-8601=minutes | sed -e 's/:/-/g' | sed -e 's/\+.*//')" >> $GITHUB_ENV
+  - name: Set .net6.0 SDK
+    run: |
+      wget https://packages.microsoft.com/config/debian/10/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+      dpkg -i packages-microsoft-prod.deb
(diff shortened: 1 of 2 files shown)


QuintillusCFC:
Okay, I'm marking this as ready.  The first commit does the .NET 6.0 part.  The remainders are DRYing out the code, there's a commit for each Workflow build.  No need to look at each one individually as most got reverted, but if you're curious what was tried, you can see the answer.

Basically it's still not as DRY as I'd like (mainly due to required boilerplate), but my net lines of code is still negative so I think it's good enough for now.  My goal here isn't to be an expert on GitHub Actions, and after last month's Serilog tangents, I'd prefer to keep this tangent short.

pcen:
looks like I can rebase #204 on this since the approach I took was pretty hacky, but it would be great to get unit tests running in CI

QuintillusCFC:
Merging after 7+ days with no-review, and self-checking.

PR #312, opened 2022-07-18
189 display roads

### LINKED ISSUE #189: Displaying existing roads from BIQ files
When loading a BIQ file that contains roads, they should be displayed on the map.

### LINKED ISSUE #190: Display existing roads from SAV and C7 JSON files
Building on #189 , roads should be displayed from SAV files (which will store them in a different location), and C7 JSON files.

This would be a logical follow-up for the person who adds BIQ road support, but could be done by someone else as well.

Kright:
Add layer for rendering roads and railroads. 
Add a few roads at top left corner of c7-static-map for demonstration.
Movement costs weren't touched, roads just displayed.

Closes #189 
Closes #190


C7/MapView.cs
```diff
@@ -983,6 +983,79 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
...
+	}
+
+	public override void drawObject(LooseView looseView, GameData gameData, Tile tile, Vector2 tileCenter) {
+		if (!hasRoad(tile)) return;
+
+		Rect2 screenTarget = new Rect2(tileCenter - tileSize / 2, tileSize);
+
+		if (!hasRailRoad(tile)) {
+			int roadIndex = 0;
+			foreach (var dirToTile in tile.neighbors) {
```
QuintillusCFC: On this line and line 1016, the `var` should be `Tile` to match our [current style guidelines](https://github.com/C7-Game/Prototype/pull/199) (most of this file was written before we'd standardized on any style guidelines, so it's a mix of styles).  In this case it's clear enough that the neighbor is another Tile, but it's not always so evident with `var`.

QuintillusCFC: I _think_ setting up the code formatting for VSCode (https://github.com/C7-Game/Prototype/wiki/Developing-and-Setting-Up-IDEs#code-formatting) might give hints about this automatically, if you're using VSCode (Rider should pick up the formatting/styles automatically).

C7/MapView.cs
```diff
@@ -983,6 +983,79 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
 	}
 }
 
+public class RoadLayer : LooseLayer {
```
QuintillusCFC: Optional: I've been trying to encourage keeping our file sizes moderate, and in that interest the ResourceLayer.cs file (C7/Map/ResourceLayer.cs) is the first map layer to be in the new Map folder.  Refactoring hasn't been a high priority since the map has been stable for a while, but long-term it would be nice to have the classes for drawing layers live in separate files in that folder, and the MapView live there too.  Wouldn't suggest doing all that here, but maybe the new class living in that folder.

QuintillusCFC (approved):
A couple minor style items, but it looks good.  Approved, no need for me to re-review it after the style change updates.

QuintillusCFC:
Oh, one thing.  I noticed this is linked to #190, but it appears that while it adds support for BIQs/JSON saves, it doesn't for SAVs yet.  Which is okay, that was always intended as the next step.  Just wanted to confirm that we should leave #190 open but note that it's just SAV support that remains.

C7/MapView.cs
Kright: Rider has refactoring extracting classes to separate files in directory. I extracted RoadLayer only, worked nice.

Kright:
> it appears that while it adds support for BIQs/JSON saves, it doesn't for SAVs yet.

I added roads to QueryCiv3.Sav, but didn't tested. Or you mean something else?

C7/MapView.cs
QuintillusCFC: A fellow Rider user!  Good to know I'm not the only one anymore!  I don't think I knew about that refactoring option though, will have to keep it in mind for next time.

QuintillusCFC:
> > it appears that while it adds support for BIQs/JSON saves, it doesn't for SAVs yet.
> 
> I added roads to QueryCiv3.Sav, but didn't tested. Or you mean something else?

So you did... so I dug into it more why it wasn't showing up.  TILE in the SAV structure is more complex than in the BIQ, and it looks like the place you added it would work for Vanilla and PTW, but not for Conquests.

Below the area where you added it, there is this section:

```
        // if civ 3 conquests file (for now we'll assume this is always true)
        private fixed byte HeaderText2[4];
        public int Length2;

        // TODO: Most flags need to be added
        private fixed byte Flags2[12];
```

Based on the notes I have from a private Civ forum, Flags2 contains:

```
   OverlayBits      4            12         Overrides TILE#1 OverlayBits.
   Unknown          1            13			   
   TerrainBits      1            14         Split in two nibbles for real and base. Overrides TILE#1 TerrainBits.
   Unknown          2            16			   
   BonusBits        4            20         Overrides TILE#1 BonusBits.
```

Whereas Flags1 contains:

```
Unknown          2            24
OverlayBits      1            25           See TILE#2 for c3c.
TerrainBits      1            26           Split in two nibbles for real and base. See TILE#2 for c3c.
BonusBits        1            27           See TILE#2 for c3c.
RiverData        1            28
```

What is present right now is Flags1[2], index 0 and 1, which is correct for PTW/Vanilla.  I believe this means that Flags2[0], index 0 and 1, are what is required for Conquests.  It definitely should be tested with a SAV file that has some roads built after the scenario starts, since I am not sure that any open-source project has implemented this before.

(As a side note, while I've shared the SAV structure info I have via e-mail with some of our collaborators, we haven't yet found a place to store it for project members)

Kright:
I loaded civ3, build city with the road near it, saved as .sav and opened it from C7. Map hasn't city, but road exits)

QuintillusCFC (approved):
I can confirm that as well.  Merging!

PR #315, opened 2022-07-19
258: Remove GD.Prints

QuintillusCFC:
Converts nearly all GD.Print statements to Serilog.

Exceptions are the right-click tile info (for devs knowing what's on a tile and its index, until we get that in the UI), GD.Prints related to logger configuration, and the Godot sink.

Note that this builds on the PR for #267 , so that one should be merged first.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -4,6 +4,7 @@ using System.Collections;
 using System.Diagnostics;
 using C7Engine;
 using C7GameData;
+using Serilog;
 
 public class Game : Node2D
 {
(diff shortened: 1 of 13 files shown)


C7/Game.cs
```diff
@@ -395,6 +398,8 @@ public override void _UnhandledInput(InputEvent @event)
 							new RightClickTileMenu(this, tile).Open(eventMouseButton.Position);
 
 						string yield = tile.YieldString(controller);
+						//These GD.Print statements are debugging prints for developers to see info about the tile
+						//For now I'm leaving them as GD.Print.  Could revisit this later.
```
pcen, line 402: In the future we could extend logging to have something like a `log.Console` method that wraps a `GD.Print` call but can be configured to no-op in release builds

pcen (approved):

C7/Game.cs
QuintillusCFC: I like the idea!  I'm a bit Serilog'ed out right now personally, but feel free to write up the idea somewhere where it'll be found again later (which means probably outside these PR comments).

PR #321, opened 2022-07-21
286 - Naval Units No Longer Defend Cities/Barb Camps

### LINKED ISSUE #286: Naval units shouldn't defend cities
Although I believe it was permissible in Civ2, in Civ3, naval units should not defend cities or barbarian camps.  In C7, they currently do.

Instead, if they are the only defensive units, they should be destroyed.

It probably wouldn't hurt to put this on a switch so players who prefer the legacy Civ2 style could re-enable it.

QuintillusCFC:
(See PR #322 first, as this builds upon that one)

#286 - Naval units (or other units of non-matching unit category) will no longer be able to defend tiles.  In practice, this means they won't defend cities or barbarian camps as for now those are the only tiles that can have units of a different category.

The solution should be generally applicable to land/sea units on water, once we get to transport boats/carriers.

This does not attempt to add mechanics for capturing units; workers and catapults will still try to defend themselves with their 0 defense stat, just like they always have.

Note that the `unit.OnEnterTile(newLoc);` method is called earlier because it now clears out remaining units on the tile, and in the old order the unit moved onto the tile before that happened (and thus would get cleared out).  Also note that OnEnterTile _is_ modified, but GitHub did a poor job of hiding irrelevant code; it starts on line 289 of MapUnitExtensions.cs (but GitHub starts showing the code from 290).

Closes #286

### DIFF
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -63,6 +63,22 @@ public static class MapUnitExtensions {
 		return unit.unitType.BaseStrength(role) * StrengthBonus.ListToMultiplier(unit.ListStrengthBonusesVersus(opponent, role, attackDirection));
 	}
 
+	public static bool CanDefendAgainst(this MapUnit unit, MapUnit attacker) {
+		//Basically, unit type must match.  Sea/air units in a city/airfield can't defend against land units.
+		//Land units on a boat or planes on a carrier can't defend against boats.  Anti-air is another category that should be checked before the direct combat.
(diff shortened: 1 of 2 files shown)


QuintillusCFC:
Closes #286

QuintillusCFC:
Odd, the Closes keyword didn't work.  Maybe because it is currently set to merge into another branch, not Development?  Or maybe it just doesn't work if you forget it when you open the PR.

QuintillusCFC:
Merging after 7-day window and self-review not finding any bugs.

PR #322, opened 2022-07-21
Naval Unit Improvements

### LINKED ISSUE #305: Boats should only heal in port
Babylon attacked my galley, I won, but it healed way far away from home the next turn.

Boats should only heal in cities.  Pending C3C verification.

### LINKED ISSUE #319: Boats should be able to enter friendly cities
Related to #305 , fixed that but then realized boats can't enter cities so with that fixed they can never heal...

QuintillusCFC:
Closes #305 
Closes #319 

This is part of a batch of naval improvements, and should be merged before #321, sorry for creating them in the wrong order!

The first commit changes it so boats now only heal in port.

The second commit allows boats to go back to port after heading out on the waves, since I realized they couldn't heal at all after the first commit!

This (potentially) concludes my night of fixing naval annoyances and making that part of the game a little bit better.  Unless I decide to teach the AI a little bit about how boats are supposed to work.


QuintillusCFC (commented):
Self-reviewed after 7 days, still looks sensible.  Going to merge.

PR #325, opened 2022-07-21
287 - Galleys No Longer Used for Land Defense

### LINKED ISSUE #287: AI shouldn't use naval units to defend cities/camps
Closely related to #286 , the AI should not task naval units with city/camp defense.

#286 should probably be done first since it's easier to verify that way, and because even in the original Civ3, you can still wind up with cases where naval units are the only ones present in cities (e.g. bombardment destroys all land units, but there are still ships parked there).

QuintillusCFC:
This prevents the barbarians from using galleys to defend their camps.

I think we are okay in terms of non-barbarian cities, though feel free to produce a counterexample.


QuintillusCFC (commented):
Self-reviewed after 7 days, looks sensible...

PR #327, opened 2022-07-21
318 - Can't build cities next to or on existing cities

### LINKED ISSUE #318: Shouldn't be able to settle cities where there's already a city, or immediately next to one
It shouldn't be possible to build a city on a site that already has a city, or right next to a tile that already has a city.

This applies to both the button (which shouldn't be visible in these cases) and to the 'b' shortcut.

QuintillusCFC:
Closes #318 .  This only applies to human players as the AI was already obeying this condition.

### DIFF
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -400,7 +400,16 @@ public static class MapUnitExtensions {
 
 	public static bool canBuildCity(this MapUnit unit)
 	{
-		return unit.unitType.actions.Contains("buildCity") && unit.location.IsAllowCities();
+		if (!unit.unitType.actions.Contains("buildCity")) {
+			return false;
+		}
+		if (unit.location.HasCity || !unit.location.IsAllowCities()) {
+			return false;
+		}
+		if (unit.location.neighbors.Values.Count(tile => tile.HasCity) > 0) {
+			return false;
+		}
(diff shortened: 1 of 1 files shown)


C7Engine/MapUnitExtensions.cs
```diff
@@ -400,7 +400,16 @@ public static void disband(this MapUnit unit)
...
 	public static bool canBuildCity(this MapUnit unit)
 	{
-		return unit.unitType.actions.Contains("buildCity") && unit.location.IsAllowCities();
+		if (!unit.unitType.actions.Contains("buildCity")) {
+			return false;
+		}
+		if (unit.location.HasCity || !unit.location.IsAllowCities()) {
+			return false;
+		}
+		if (unit.location.neighbors.Values.Count(tile => tile.HasCity) > 0) {
```
Kright: I suggest to use Any instead of Count.

Kright (changes requested):

C7Engine/MapUnitExtensions.cs
QuintillusCFC: Good idea.  I wound up taking the idea from #337 and using All with !HasCity, IMO it's slightly easier to read that way.

Kright (approved):

QuintillusCFC:
That was fast!  Merging...

PR #328, opened 2022-07-21
#106 - Fix the Seams

### LINKED ISSUE #106: World Map - Fix the Seams
At many zoom levels, there are ugly seams in the map.

This is to ideally fix that while retaining zoom.

The idea of only having a few fixes zooms (100%, 200%, 50%, etc.) to fix the problem has also been suggested, and may be either an interim or permanent solution.

QuintillusCFC:
Closes #106 

This uses option 4 among those listed.  I still agree with Jim that option 2 is the more elegant/correct, and least hackish, but decided to try option 4 because this has been on our todo list since mid-February, is our most visible UI issue, and option 2 sounds complex enough no one has tried it.

Surprisingly, a 0.1% increase in the texture area solved the problem and it looks pretty darn good.

I'm not opposed to switching to option 2 some day, but in the meantime I think this is a worthwhile improvement.


QuintillusCFC:
Going ahead and merging as there are no objections.  If we want to go with a more elegant solution later, that's great too, but a seamless experience in the meantime will be nice.

Issue #330, opened 2022-07-22
Handle Strategy Maps in biq's PRTO structure


QuintillusCFC:
While working on #326 , I was trying to figure out why Sengoku wasn't importing as expected.  For some reason it was trying to import the Ronin twice... but it was only in the unit list once.

Then I noticed that Rider said there were 46 PRTOs, but counting them manually, there were fewer than that.  Then I remembered... strategy maps!

Civ3 has a really weird file structure when you give multiple possible strategies to a unit.  Instead of storing these in, say, a bitfield, it creates an entire new copy of the unit.  One has one strategy, the other has the other.  If you add a third strategy, there is a third unit of the same type.

The Ronin has both the Offense and Explore strategies ticked, and thus is stored twice using StrategyMaps.

We should _not_ store multiple copies of units for strategy maps!  Thankfully, unlike in my editor, we won't have to write them out in that manner either, just have the sense to know they aren't really two units.

You might think that the way to check for strategy maps is the units having the same name, but it's actually to check if the "otherStrategy" flag is not -1.  If it's not -1, the unit is a StrategyMap of another unit.  Excerpt from my editor's PRTO import code that shows this:

```
                if (newPRTO.getOtherStrategy() != -1)    //strategy map
                {   //combine the strategies
                    
                    unit.get(newPRTO.getOtherStrategy()).AIStrategy = (unit.get(newPRTO.getOtherStrategy()).AIStrategy) | newPRTO.AIStrategy;
                    //Recalculate the constituent parts
                    unit.get(newPRTO.getOtherStrategy()).extractAIStrategy();
                }
```

So it isn't too bad once you know what's going on (and don't have to re-export it!).  Still doesn't make sense as a data structure, but we can handle it.

Issue #329, opened 2022-07-22
GameData's unitPrototypes shouldn't really be a Dictionary<string, UnitPrototype> anymore


QuintillusCFC:
When we were in the Wild West days of Aztec and to some extent Babylon, and unit prototypes were written in code and not imported, it was convenient to have them stored indexed by their name.  Just refer to unitPrototypes["Warrior"] and you have a Warrior!

However, this is increasingly an anachronism, and isn't helpful when we have imported, variable unit types.  I checked all of our references to unitPrototypes, and the ones that still use this style are all holdovers from those early days, and won't work reliably with "non-standard" unit names.  More often, especially with newer code, we refer to the dictionary's values and ignore the keys altogether.

If we ever do need to find a unit prototype by name (perhaps for debugging purposes?), we can use a LINQ .Where or .Find clause.

So this refactor is basically to convert this data structure into a List.

Issue #140, opened 2022-03-12
Add/fix custom graphics via scenario search path in legacy Civ3 scenarios


QuintillusCFC:
There's already some code around this, but it isn't really working right now.  We ought to fix it before we forget about it entirely.

PR #331, opened 2022-07-23
#326 - Use scenario-specified barbarian units

### LINKED ISSUE #326: Use scenario-specified default barbarian units
Some scenarios use custom barbarian untis, e.g. 06 Age of Discovery and 07 Sengoku.  Our code fails because it always looks for Warrior and Galley, and these scenarios don't have those units, but others.

We should use those units instead.

QuintillusCFC:
This allows scenarios that don't use Warrior/Horseman/Galley for barbarians to work in that regard.

Notably, 06 MP Age of Discovery now loads with Archer barbarians.

This work spawned #329 and #330 to make further improvements to this area.

---

Note: With this change, the whole list of unit prototypes is imported (rather than a whitelisted subset), and the limiter of what can be built is moved to CanBuildCity in City.cs.  Eventually this will disappear, but this change allows the barbarians to build units that aren't on that whitelist.

It should also be noted that 06 MP Age of Discovery is not really playable yet, since you can only build Settlers and Workers and that rapidly leads to crashes due to running out of citizens.  But it can load its barbarians properly, and a custom scenario with modified barbarians should work too.


C7/Game.cs
```diff
@@ -108,7 +108,13 @@ public override void _Ready()
...
 			errorOnLoad = true;
 			PopupOverlay popupOverlay = GetNode<PopupOverlay>(PopupOverlay.NodePath);
-			popupOverlay.ShowPopup(new ErrorMessage(ex.Message), PopupOverlay.PopupCategory.Advisor);
+			string message = ex.Message;
+			string[] stack = ex.StackTrace.Split("\r\n");	//for some reason it is returned with \r\n in the string as one line.  let's make it readable!
+			foreach (string line in stack) {
+				message = message + "\r\n" + line;
+			}
+
+			popupOverlay.ShowPopup(new ErrorMessage(message), PopupOverlay.PopupCategory.Advisor);
```
QuintillusCFC, line 117: This gives a more useful error message if a file fails to load for some reason.  Yeah, it's in the logs too (which is nice!), but having a hint in-your-face is nice too.  Eventually we probably want to roll this back but when it's mostly developers playing, I think it makes sense.

C7GameData/ImportCiv3.cs
```diff
@@ -217,50 +223,48 @@ public static C7SaveFormat ImportBiq(string biqPath, string defaultBiqPath)
...
+					prototype.actions.Add("fortify");
+				}
+				if (prto.Disband) {
+					prototype.actions.Add("disband");
+				}
+				if (prto.GoTo) {
+					prototype.actions.Add("goTo");
+				}
+				//Temporary check until #329/#330 are finished
+				if (!c7SaveFormat.GameData.unitPrototypes.ContainsKey(prototype.name)) {
```
QuintillusCFC, line 270: This temporary check is the only change to this section, the rest is indentation.  This change allows scenarios with strategy maps (which is nearly all of them) to be loaded now that all unit prototypes are imported.

WildWeazel:
bumping to resolve conflicts

QuintillusCFC:
Fixed the conflicts.  Still looks sensible to me.  I'd kind of forgotten that this was still here and not merged.

QuintillusCFC:
I'm going to go ahead and merge this since I self-reviewed it 3 days ago and there hasn't been another review since 7/23.  Also because I was looking at #140 and that will be easier once these changes are merged due to more scenarios being openable.

PR #332, opened 2022-07-23
191 refactor pathing algorithm

Kright:
Related to #191

Fractional movement points will be added in another pull request, because this is huge enough.

I rewrote dijkstra pathing algoritm for more abstract way. I decoupled it from game logic for movement possibility and costs.

DijkstrasAlgorithm.run(...) - pure algorithm, don't depend on game staff.
EdgeWalker interface passed to Dijkstra's algorithm and called back do determine movement costs between tiles.
WalkerOnLand - example of EdgeWalker imlementation for searching path on land.

BinaryMinHeap - custom data structure, needed for performance of Dijkstra's algorithm. "net472" doesn't have PriorityQueue, so I wrote it myself. Surprisingly the pathing algoritm turned out to be very simple.

I hope that existing BFSLandAlgorithm and DijkstrasLandAlgorithm could be replaced with new implementation and custom EdgeWalkers.

P.S. Algorithm already uses roads info when searching path, but units still untouched and they walk slow as without road.

### DIFF
--- a/C7Engine/AI/Pathing/BinaryMinHeap.cs
+++ b/C7Engine/AI/Pathing/BinaryMinHeap.cs
@@ -1,0 +1,83 @@
+using System;
+using System.Collections.Generic;
+
+namespace C7Engine.Pathing {
+	/**
+	 * https://en.wikipedia.org/wiki/Binary_heap
+	 * min heap
+	 */
+	public class BinaryMinHeap<TValue> where TValue: IComparable<TValue> {
+		private readonly List<TValue> data = new List<TValue>();
+
+		public int count { get => data.Count; }
+
+		// average O(1), worst case O(log N)
+		public void insert(TValue v) {
+			data.Add(v);
+			siftUp(data.Count - 1);
+		}
+
+		// extract them smallest value, O(log N)
+		public TValue extract() {
+			TValue result = data[0];
+			data[0] = data[data.Count - 1];
+			data.RemoveAt(data.Count - 1);
+			if (data.Count > 0) {
+				siftDown(0);
+			}
+			return result;
+		}
+
+		private void siftUp(int childIndex) {
+			if (childIndex == 0) return;
+			int parentIndex = getParentIndex(childIndex);
+			if (!isRightOrder(parentIndex, childIndex)) {
+				swap(parentIndex, childIndex);
+				siftUp(parentIndex);
+			}
+		}
+
+		private void siftDown(int parentIndex) {
+			int leftChild = getLeftChild(parentIndex);
+			int rightChild = leftChild + 1;
+			if (rightChild < data.Count) {
+				// two children
+				bool leftShouldBeHigher = isRightOrder(leftChild, rightChild);
+				int topChildIndex = leftShouldBeHigher ? leftChild : rightChild;
+				if (!isRightOrder(parentIndex, topChildIndex)) {
+					swap(parentIndex, topChildIndex);
+					siftDown(topChildIndex);
+				}
+				return;
+			}
+
+			if (leftChild >= data.Count) {
+				// no children
+				return;
+			}
+
+			// one children
+			if (!isRightOrder(parentIndex, leftChild)) {
+				swap(parentIndex, leftChild);
+			}
+		}
+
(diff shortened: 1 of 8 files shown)


C7Engine/AI/Pathing/BinaryMinHeap.cs
```diff
@@ -0,0 +1,82 @@
...
+				swap(parentIndex, leftChild);
+			}
+		}
+
+		private bool isRightOrder(int parentIndex, int childIndex) {
+			return data[parentIndex].CompareTo(data[childIndex]) <= 0;
+		}
+
+		private void swap(int index, int otherIndex) {
+			(data[index], data[otherIndex]) = (data[otherIndex], data[index]);
```
QuintillusCFC, line 69: That is nifty that you can swap things with this syntax in C#.  A new solution (to me) to the "how do you swap values without an intermediary variable?" question.

EngineTests/AI/Pathing/BinaryMinHeapTest.cs
```diff
@@ -0,0 +1,56 @@
...
+			Assert.Equal(result, sortedCopy);
+		}
+
+		public List<int> generateLargeList() {
+			List<int> lst = new List<int>();
+			for (int i = 0; i < 1000; ++i) {
+				lst.Add(i);
+			}
+
+			Random random = new Random(0x1337);
```
QuintillusCFC, line 33: I see how that random seed was chosen...

One minor issue, the first 39 lines of this file have spaces for indentation, whereas the rest of the PR uses tabs.  We've still got some legacy spaces in other files, but are trying to standardize on tabs.

C7Engine/AI/Pathing/DijkstrasAlgorithm.cs
```diff
@@ -0,0 +1,73 @@
...
+			Dictionary<Tile, Edge<Tile>> walkingMap = run(start, destination, edgeWalker);
+			return makePath(walkingMap, destination);
+		}
+
+		/**
+		 * return Dictionary: Node -> Edge(previousNode, node, distanceToNodeFromStart)
+		 * when stopWhenReachDestination == false it calculates distances to each available point
+		 */
+		public static Dictionary<TNode, Edge<TNode>> run<TNode>(TNode start, TNode destination,
+			EdgeWalker<TNode> edgeWalker, bool stopWhenReachDestination = false)
```
QuintillusCFC: Based on how it's used currently, i.e. with no caching, wouldn't it make more sense to make stopWhenReachDestination default to true?  Is there any benefit to building the whole tree if we don't reuse it?

I suspect rebuilding the whole tree is the reason that even with the binary heap improvement, I'm still seeing the "Explorer pathing took 179 milliseconds" warnings occasionally - it's saving time from the better data structure, but it's calculating the whole tree whereas it wasn't previously.

QuintillusCFC (commented):
Looks like a good general improvement!  See the comment about the default for calculating the whole tree - I think the extra whole-tree calculation is outweighing the data structure improvements at least part of the time as we currently use it.  But aside from that, this looks like a nice step forward in both generalization and efficiency.

C7Engine/AI/Pathing/DijkstrasAlgorithm.cs
Kright: It's a mistake. I should use stopWhenReachDestination = true by default, will fix it.

EngineTests/AI/Pathing/BinaryMinHeapTest.cs
Kright: Are you sure? I checked this file now and found nothing but tabs.

QuintillusCFC: Odd, it must have been a GitHub glitch, when I view it in Rider it is indeed tabs. Sorry for the false alarm!

C7Engine/AI/Pathing/DijkstrasAlgorithm.cs
QuintillusCFC: Yup, that makes a huge difference.  With it as true, there are no log statements in 140+ turns that hit the explorer warning time threshold.

PR #334, opened 2022-07-28
101: Tile Visibility

### LINKED ISSUE #101: Only tiles that have been explored should be visible
Tiles that haven't been explored shouldn't be visible (unless a debug mode flag is enabled; we should include one so we can enable it when developing the AI).

TBD whether this is Babylon or Carthage.

QuintillusCFC:
Closes #101 

Still a WIP, but you can try it out if you like.

There's probably a way to do it with shaders, but I'm the AI guy, not the graphics guy.  I just got inspired to make this originally-planned-for-Babylon feature work and figured no one else had done it and I had an idea how to do it, and it worked.


WildWeazel:
hmm, we should probably have someone who knows shaders...

QuintillusCFC:
> hmm, we should probably have someone who knows shaders...

I believe both @maxpetul and @Kright know shaders; Max has definitely added them and Kright mentioned them in a PR review recently.  I know we also had some references on where to learn them shared, but that hasn't risen to the top of my priority list yet...

QuintillusCFC:
Note to self on what remains:

 - Possibly optimize which tiles draw fog?
 - Likely actually draw the whole SE, S, SW tiles that haven't been discovered yet, so the fog can obscure them but features that overlap (rivers, but also mountains) are partially visible.
 - Do some software archaeology to try to figure out why Civ seems to use different graphics, was Vanilla 1.00 different?  Not that I remember, but if so why not?  Ultimately we may need to use an algorithmic option like Civ (esp for MP).
 - Shaders?  Perhaps most useful if we go the algorithmic route, but maybe useful regardless for the alpha?

Overall I think the existing changes are already a significant improvement and make the "eXploration" part of the game viable; these changes would be the polish.  If this PR were two or three times the size I'd probably split these improvements out,and maybe I will yet.

QuintillusCFC:
> * Likely actually draw the whole SE, S, SW tiles that haven't been discovered yet, so the fog can obscure them but features that overlap (rivers, but also mountains) are partially visible.

This one is actually sorta important for fog-peering.  Seed 123 through commit ab89a3c:

image

Looks flat, right?  But if we draw the bordering tiles:

image

We can see there's a mountain tile.  But also that we need the fog to draw for realsies in order to cover features of the other tiles - large forests, tall mountains, etc. - that escape the border for one tile away.

That might result in version 1.5 of fog of war, I want to be careful not to scope creep this to 500 lines of code, and "version 1.0" as submitted 2 days ago is a huge improvement for 4X gameplay.

I also checked the Civ3 GOTY Edition box, which is Vanilla, and it uses the same patterns we're all used to for fog of war.  Still not sure why the file is different than what we see in game, but that also seems like it can wait for version 1.5.

Thus I'm going to open this for review, with the understanding that like kright's road prototype PR, this isn't meant to be final-version never-again-modified.

QuintillusCFC (commented):
It's been a month and no reviews, so I re-examined it, merged dev into it, tested it... it's going to evolve a bit, but it's one giant leap forward for gameplay purposes so I'm going to merge it.

Issue #192, opened 2022-04-05
Teach the AI workers to build roads


QuintillusCFC:
Once we have the ability to display roads, we should teach the AI workers to build them.  This will be much more productive for them than their current behavior of scouting, and usually being attacked by barbarians.

If #137 has already been done, then some logic should also be added here so that the AI will occasionally build more workers than just what it starts with.

PR #336, opened 2022-07-28
add prototype of roads building

Kright:
Add button for building roads by workers. There is a prototype, the road built immediately.
AI can't build roads yet.

It's a preparation for #192

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -605,6 +605,13 @@ public class Game : Node2D
 			}
 				break;
 
+			case "buildRoad": {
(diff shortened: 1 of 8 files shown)


QuintillusCFC (commented):
Looks cool, one small thing, can you add the keyboard shortcut to the button?  It ought to work to add it in UnitButton.cs, line 53, just like the other ones in the file but with R.

This also reminds me that we should make those configurable someday, so I wrote up #341 for that, but I wouldn't delay road building in order to make the shortcuts configurable.

QuintillusCFC (approved):
Provisional approval once the road-building shortcut is ready.

PR #337, opened 2022-07-28
don't build city on existing city or near it, simplify code

Kright:
Don't allow to build city on the tile which has city already or city near it.


Kright:
Sorry, there was older pull request: #327

PR #340, opened 2022-07-28
use MovementPoints with float value instead of plain int

### LINKED ISSUE #191: Dijkstra's Algorithm - Consider Whether Tiles have Roads

Kright:
Closes #191

Roads allow values like 1/3.
I replaced int values with MovementPoints struct. It uses float and has helper methods for avoiding rounding errors.


WildWeazel (approved):
looks great!

PR #345, opened 2022-09-16
153 - Move resources into folders and copy all at once

### LINKED ISSUE #153: Copy all resources at once in Git Actions build
In the Git Actions build (controlled by export-c7.yml), right now there is a step to copy all the resources ("static files"), on lines 61, 107, and 157.  But we have to name each individual file.  We could do that for Aztec (1 file) and Babylon (3 files), but as we start adding units and other graphics, it will rapidly become unsustainable.

This card is to refactor that YML (and probably our file structure, putting all static files in a sub-folder of the C7 project) so that we don't have to update that file every time we add a new resource/static file.

Optional bonus points for figuring out how to de-duplicate the copying of static files across Windows/Linux, and maybe Mac.  Mac does it a little bit differently right now, so it might not be feasible to do so, at least across all three.  Ask JimOfLeisure (Puppeteer on CivFanatics) about the details of why the Mac step thinks different in this regard.

QuintillusCFC:
This is the precursor to being able to add our own units.  It adds two folder - Art and Text - and moves our resources into them.

So far, Art contains just the title PCX, and text contains credits.txt and the static JSON.

It should work just fine when we add sub-folders.

There's a test action build using these changes at https://github.com/C7-Game/Prototype/actions/runs/3070488664  I have already verified the Windows one works as expected.

As a side improvement, if the static copy fails for some reason, formerly it gave a warning in the GitHub Actions console but still succeeded.  Now it still prints the warning, but will also fail the build.  This wound up being much easier than trying to fail if the Godot build itself failed.


QuintillusCFC:
Self-reviewed.  Saw that the old static-JSON paths should be removed from .gitattributes and did that, but otherwise looks sensible.

PR #346, opened 2022-09-18
131 - Include a Unit with C7

### LINKED ISSUE #131: Include One of Our Own Units
- Package up a unit to ship with Carthage
 - Format can remain the legacy format for now
 - Work with the artists to find a unit that someone's willing to have be part of C7
 - Include it in the game

QuintillusCFC:
First we look for the unit on a scenario's mod path, which still is only among Civ3 files.

Then we check for a C7 native version.

Finally we check the base Civ paths.

Please review #345 first!


QuintillusCFC:
Self-reviewed.  Improved a method name, sorted a couple modifiers.  The general method for searching still looks solid, at least until we support mods stored relative to a C7 install.

PR #348, opened 2022-09-19
140 - Display Scenario-Specific Graphics

### LINKED ISSUE #140: Add/fix custom graphics via scenario search path in legacy Civ3 scenarios

QuintillusCFC:
Based on the scenario search path.

I'd thought we had this before, but it was only in the legacy Civ3Map, and only when you tossed a string for the folder in the source code.

Now, it stores the search path in Util, and the Civ3MediaPath method uses that automatically.  From a pure OO standpoint, the old way of having to pass in the search path as an argument may be cleaner, but we already were calling that method from 25 places and it just didn't seem to make sense to have that cognitive overhead in 25 places when it's only going to change when a new game/scenario is loaded.

Please review #131 first!

### DIFF
--- a/C7/Civ3Map/Civ3Map.cs
+++ b/C7/Civ3Map/Civ3Map.cs
@@ -81,7 +81,7 @@ public class Civ3Map : Node2D
 		// temp if
 		if(FileNameLookup[fileID] != null)
 		{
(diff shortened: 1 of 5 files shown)


QuintillusCFC:
Self-reviewed, and fixed up a comment  that my second commit had made out-of-date.  Merging.

PR #351, opened 2022-09-19
90 - Bonus Grassland Overlay and Production Effect

### LINKED ISSUE #90: Bonus Grassland Overlay
Add the bonus grassland overlay.

These use the "desert" TNT overlay, specifically the 4th row down (first row being row 1) in Art\Terrain\tnt.pcx.  I have not determined which one Civ uses, and in my editor use a deterministic calculation based on the tileIndex % 3.  I suggest we start with something equivalent; if someone really cares and knows how it's determined, great, but the important thing is it doesn't change every frame.

QuintillusCFC:
Bonus grasslands now show up, and production from the bonus is accounted for.

I added the check for grassland as the tile key because in Civ3 SAV files, tiles other than grassland can apparently have the bonus flag checked.  Either that or our import process has an error in it, but I checked that and it looks right to me.

### DIFF
--- a/C7/Map/TntLayer.cs
+++ b/C7/Map/TntLayer.cs
@@ -1,0 +1,42 @@
+using C7GameData;
+using Godot;
+using Resource = C7GameData.Resource;
+using Serilog;
+
+namespace C7.Map
+{
+	/// <summary>
+	/// Displays terrain yield overlays (from the tnt.pcx file).  These are most well known for letting you know where
+	/// there are bonus grasslands.
+	/// Note: I don't know why it's called tnt.
+	/// </summary>
+	public class TntLayer : LooseLayer
+	{
+		private ILogger log = LogManager.ForContext<ResourceLayer>();
+
+		private static readonly Vector2 tntSize = new Vector2(128, 64);
+		private ImageTexture txtTexture;
+
+		//Each row corresponds to a terrain.  For now we're only adding one, maybe someday we'll add full TNT support
+		private readonly int GRASSLAND_ROW = 0;
+		private readonly int BONUS_GRASSLAND_ROW = 1;
+		private readonly int PLAINS_ROW = 2;
+		private readonly int DESERT_ROW = 3;
+		private readonly int BONUS_GRASSLAND_TNT_OFF_ROW = 3;
+		private readonly int TUNDRA_ROW = 4;
+		private readonly int FLOOD_PLAIN_ROW = 5;
+
+		public TntLayer()
+		{
+			txtTexture = Util.LoadTextureFromPCX("Art/Terrain/tnt.pcx");
+		}
+		public override void drawObject(LooseView looseView, GameData gameData, Tile tile, Vector2 tileCenter)
+		{
+			if (tile.overlayTerrainType.Key == "grassland" && tile.isBonusShield) {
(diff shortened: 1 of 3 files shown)


C7/Map/TntLayer.cs
```diff
@@ -0,0 +1,42 @@
...
+	/// Displays terrain yield overlays (from the tnt.pcx file).  These are most well known for letting you know where
+	/// there are bonus grasslands.
+	/// Note: I don't know why it's called tnt.
+	/// </summary>
+	public class TntLayer : LooseLayer
+	{
+		private ILogger log = LogManager.ForContext<ResourceLayer>();
+
+		private static readonly Vector2 tntSize = new Vector2(128, 64);
+		private ImageTexture txtTexture;
```
WildWeazel: supposed to be t**n**tTexture?

C7/Map/TntLayer.cs
```diff
@@ -0,0 +1,42 @@
...
+		private readonly int FLOOD_PLAIN_ROW = 5;
+
+		public TntLayer()
+		{
+			txtTexture = Util.LoadTextureFromPCX("Art/Terrain/tnt.pcx");
+		}
+		public override void drawObject(LooseView looseView, GameData gameData, Tile tile, Vector2 tileCenter)
+		{
+			if (tile.overlayTerrainType.Key == "grassland" && tile.isBonusShield) {
+				Rect2 tntRectangle = new Rect2(0 * tntSize.x, BONUS_GRASSLAND_TNT_OFF_ROW * tntSize.y, tntSize);
```
WildWeazel: the first parameter will always be 0

C7/Map/TntLayer.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace C7.Map
+{
+	/// <summary>
+	/// Displays terrain yield overlays (from the tnt.pcx file).  These are most well known for letting you know where
+	/// there are bonus grasslands.
+	/// Note: I don't know why it's called tnt.
+	/// </summary>
+	public class TntLayer : LooseLayer
+	{
+		private ILogger log = LogManager.ForContext<ResourceLayer>();
```
WildWeazel: Are these layers sharing a log context or is this supposed to be TntLayer? A more generic copy-friendly but slightly inefficient way to do this might be something like ForContext<this.GetType())>() (I have not tested that)

WildWeazel (changes requested):

C7/Map/TntLayer.cs
QuintillusCFC: It looks like it isn't possible to use this.GetType() because this is technically in the initializer.  But that would be great, because it's definitely supposed to be TntLayer, it's a copy-paste error.  It's just complex enough syntax that I tend to copy and modify it, and sometimes forget the modify part.

C7/Map/TntLayer.cs
QuintillusCFC: True.  In Civ3, it alternates between 0 and 2 depending on some formula based on the tile, probably similar to forests/mountains.  I've just left it as 0 for now in lieu of a formula, the differences are miniscule anyway.  It now is always 0, no multiplication needed.

C7/Map/TntLayer.cs
QuintillusCFC: Yup.

WildWeazel (approved):

Issue #323, opened 2022-07-21
Crash due to failing to remove zombie unit from tile


QuintillusCFC:
This part of MapUnitExtensions.cs can cause a crash:

```
			if (!unit.location.unitsOnTile.Remove(unit))
				throw new System.Exception("Failed to remove unit from tile it's supposed to be on");
```

It seems to happen semi-randomly when barbarians move, although the fact that it's barbarians moving may be coincidental as they move a lot and engage in combat a lot.  I might also be seeing it more because I've been using the Shift key to skip animations, and there might be a timing issue involved in it.

Now that we have good logging, we should probably avoid unhandled exceptions like this, and prefer logging errors.  But it's also curious that this is happening at all.

Adding to Carthage so someone can pick it up if they please, it's never good to leave crashes around.

QuintillusCFC:
Just had it happen again after two galleys fought.  Not sure if that's always the case?  Have also seen a bunch of times where it doesn't crash after galleys fight.  And then it did happen when I pressed Shift during the fight.

QuintillusCFC:
Investigating at a breakpoint, it seems that both units wound up on the same tile and it tries to remove units from what presumably is the other tile in the combat.  Although it claims both of those units have 3 hitpoints (when it should be 0 and 1), and there's another tile right next to the combat that has two healthy galleys.

I think this would be easier to investigate with a fixed seed, and with the error converted to an error first rather than a crash, so the map could be queried for Tile IDs afterwards.  As it is, it's just tough to evaluate having new circumstances every time it happens.

QuintillusCFC:
Reproducing info: With commit 7976165cc24556993e279a7ea7ccc6622bc1cf71

And seed: 1626442457 (or 1453020662?  It prints the seed twice with different values for some reason)

On turn 19, this happens via BarbarianAI.cs, line 37

Fortify the Carthage units, enter Observer Mode.  On turn 11 Carthage should destroy a barb camp.  Play till turn 18, add a breakpoint.

QuintillusCFC:
Unit location = 10, 66

It's a galley moving northwest.

QuintillusCFC:
The barb camp west of Rome spawns a galley for turn 13.

Rome sends a galley out on turn 16.

Starting turn 18, Rome's galley is at 9, 69; the barbarian one is at 10, 66.

Rome's galley moves north to 9, 67.

The barbarian galley moves to 9, 67 (which is southwest, not northwest).

[...]

PR #353, opened 2022-09-19
Add an Observer Mode

### LINKED ISSUE #352: Add an Observer Mode
Now that we have tile visibility, an observer mode that lets you see everything would be really nice, especially when adding new map features.

QuintillusCFC:
Activated (and deactivated) with Ctrl+Shift+Alt+O.

This is slightly different than Civ3's Observer Mode, modeled more on Paradox's Observer Mode.  Specifically, it's different in that the AI will assume control of your civilization while you are in Observer Mode.

This makes it more useful for hands-off observation, and will be yet more useful, especially for testing how scenarios play out, or just if the player wants an advanced start, if an "autoplay" feature is added (which is featured in some Civ4 mods).

Known issue: If you enter Observation Mode immediately, before fortifying your units, each turn ended with the "Enter" key will cause two turns to happen.  If you fortify your units first, or use the NumPad enter key, that won't happen.

I've identified that the cause is at least in part that the Godot `BlinkyEndTurnButtonPressed` event we have is firing right after the Enter key event is processed in this case, but am not sure exactly why or why it happens only for Enter.  Presumably somehow it has the focus in that case?  But I suspect it's a deep enough rabbit hole that I don't really want to get into it, a good enough Observation Mode is good enough for Carthage.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -535,6 +535,22 @@ public class Game : Node2D
 			{
 				SetAnimationsEnabled(false);
 			}
+			else if (eventKeyDown.Scancode == (int)Godot.KeyList.O && eventKeyDown.Shift && eventKeyDown.Control && eventKeyDown.Alt) {
+				using (UIGameDataAccess gameDataAccess = new UIGameDataAccess()) {
+					gameDataAccess.gameData.observerMode = !gameDataAccess.gameData.observerMode;
+					if (gameDataAccess.gameData.observerMode) {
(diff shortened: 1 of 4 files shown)


QuintillusCFC (commented):
Self-reviewed, nothing alarming spotted.  Merging since this and logging the seed, which depends on it, will both be helpful in debugging, e.g. #323 needs them to be merged to become easier to investigate.

PR #354, opened 2022-09-20
335 - Log the Seed

### LINKED ISSUE #335: Log the seed, and allow setting the seed in the UI
One of the things I've wished we had multiple times was the ability to look up the seed after the fact.  Play a game, often just watching the AI, and see something interesting, but not being able to recreate it is unfortunate.

Civ3 does this by storing the seed in conquests.ini.

Either "New Game" or "Quick Start" should default to the last seed and allow you to set one.  Probably New Game should let you enter the seed; this can be the first part of our "New game setup" screens.

Technically, my guess is we'll have to log randomly-generated seeds by generating them and then passing them in to the random component, as it doesn't seem like there's a way to extract a seed from a Random generated without any parameters.  I could be wrong about this (and it's different in Java), but there doesn't seem to be such an API.

QuintillusCFC:
The seed is now logged to the log.  This makes it possible to reproduce odd behavior that you see when you hadn't been expecting odd behavior.

Also cleaned up several spare RNGs that were in use throughout the code.

I have not added the ability to set the seed in the UI.  It seems like it should be easy, I could do it in one line in Java, but I couldn't find a way to do it easily here.  I'm leaning towards punting on that until we're ready to have a game setup screen.

### DIFF
--- a/C7Engine/AI/CityProductionAI.cs
+++ b/C7Engine/AI/CityProductionAI.cs
@@ -113,8 +113,7 @@ namespace C7Engine
 				i++;
 			}
 
-			Random random = new Random();
-			double randomDouble = sumOfAllWeights * random.NextDouble();
+			double randomDouble = sumOfAllWeights * GameData.rng.NextDouble();
 			log.Verbose($"Random number in range 0 to {sumOfAllWeights} is {randomDouble}");
 			int idx = 0;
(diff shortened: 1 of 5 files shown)


QuintillusCFC (commented):
Self-reviewed, looks good, merging.

Issue #359, opened 2022-09-20
Refactor AI unit strategy assignment


QuintillusCFC:
Of the three AI decision-making types we have (unit strategy, city production, and strategic priority), unit strategy is currently the least sophisticated.  It's essentially a deterministic if-else chain, which is not going to be very maintainable as it grows, and doesn't allow for any element of chance.

City production has a more sophisticated process of weighing each possibility and then deciding among the options, and strategic priorities have that plus an architecture that is at least in theory pluggable.  They've both been worked on more recently.

QuintillusCFC:
Either as part of this or spun off, we should also use the `path` and `moveAlongPath` property/methods on Unit/UnitExtensions, rather than duplicating that logic in various AI behavior classes.

PR #360, opened 2022-09-20
357 - Teach the AI to take out barb camps.

### LINKED ISSUE #357: Teach AI to Take Out Barbarian Camps
Currently, the AI only destroys barbarian camps if it is lucky and its exploration units take one out.  Barbarian camps can stick around forever, even in the middle of AI empires.

The AI should be taught that they shouldn't let barbarians hang out right outside their gates.

QuintillusCFC:
The AI will no longer let barb camps stick around forever, but will take out barb camps within 3 tiles of a city.

This is still not the most proactive stance, but is a start towards the AIs being more competent at not letting the barbarians run roughshod over their lands.

Closes #357

### DIFF
--- a/C7Engine/AI/PlayerAI.cs
+++ b/C7Engine/AI/PlayerAI.cs
@@ -6,6 +6,7 @@ using C7GameData;
 using C7GameData.AIData;
 using C7Engine.AI;
 using C7Engine.AI.StrategicAI;
+using C7Engine.AI.UnitAI;
 using Serilog;
 
 namespace C7Engine
@@ -97,6 +98,9 @@ namespace C7Engine
 				log.Information("Set defender AI for " + unit + " with destination of " + ai.destination);
 				unit.currentAIData = ai;
 			}
+			else if (UnitAttackingBarbCamp(unit, player)) {
+				log.Information("Set unit " + unit + " to take out barb camp");
+			}
 			else if (unit.unitType.name == "Catapult") {
 				//For now tell catapults to sit tight.  It's getting really annoying watching them pointlessly bombard barb camps forever
 				DefenderAIData ai = new DefenderAIData();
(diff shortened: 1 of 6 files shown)


C7Engine/AI/UnitAI/CombatAI.cs
```diff
@@ -0,0 +1,41 @@
...
+			while (true) {
+				Tile t = explorerData.path.Next();
+				if (t == Tile.NONE) {
+					break;
+				}
+
+				//Todo: Is there a less clunky way to do this?
+				//Uhhhmm, actually yes, there's a move along path method.  The things is the path has
+				//to be on the unit, rather than the unit AI.  Looks like we're only using that for
+				//human-inputted paths (go to for tiles more than one tile away).
```
QuintillusCFC: See #359 .  I think we added the human-inputted paths after we'd added some AI ones, but there are several places that need to be refactored.  Rather than mix refactor and feature, I've left them separate for now.

C7Engine/AI/UnitAI/CombatAI.cs
```diff
@@ -0,0 +1,41 @@
...
+	/// <summary>
+	/// A unit whose intended role is combat, other than purely point-defense combat.
+	///
+	/// This is likely to evolve significantly, and likely have sub-classes; this is the very first iteration.
+	/// This first iteration is focused on defeating barbarians.
+	/// </summary>
+	public class CombatAI : C7Engine.UnitAI {
+		private Tile target;
+
+		private ILogger log = Log.ForContext<DefenderAI>();
```
WildWeazel: wrong context again

C7Engine/AI/UnitAI/CombatAI.cs
```diff
@@ -0,0 +1,41 @@
...
+
+		public bool PlayTurn(Player player, MapUnit unit) {
+			CombatAIData explorerData = (CombatAIData)unit.currentAIData;
+
+			while (true) {
```
WildWeazel: I'm not sure I follow why this is an infinite loop and it looks kind of dangerous. Are we trying to find where along the original path the unit is now? Would it make sense to represent the path as something like a stack or linked list and pop the next tile with each move?

C7Engine/AI/PlayerAI.cs
```diff
@@ -176,6 +180,39 @@ public static void SetAIForUnit(MapUnit unit, Player player)
...
+			if (unit.unitType.attack <= 0) {
+				return false;
+			}
+
+			List<Tile> reachableBarbCampsTiles = new List<Tile>();
+			foreach (Tile t in player.tileKnowledge.AllKnownTiles()
+				         .Where(t => unit.CanEnterTile(t, true) && t.cityAtTile == null && t.hasBarbarianCamp))
+			{
+				reachableBarbCampsTiles.Add(t);
+			}
```
WildWeazel: I see you've discovered LINQ but not fully on board yet ;)

`List<Tile> reachableBarbCampsTiles = player.tileKnowledge.AllKnownTiles().Where( ... ).toList()`

C7Engine/AI/UnitAI/CombatAI.cs
QuintillusCFC: This definitely looked more confusing coming back to it after a week than right after I wrote it.  I don't remember why it has an infinite loop.  Best guess is it's supposed to move until it's out of movement points or it engages in combat, but it currently won't move more than one tile even if it has movement points for it.

Definitely needs changes!  Thanks for calling this out while I still remember some of why I wrote it the way I wrote it!

The path _is_ represented as a stack (TilePath.java), but that hasn't been fully integrated into the AI yet.

C7Engine/AI/UnitAI/CombatAI.cs
QuintillusCFC: Should be fixed.  Really wish there were a way to make that automatic.

C7Engine/AI/UnitAI/CombatAI.cs
QuintillusCFC: This has been revamped.  Now it will use existing methods to follow its path to the barbarian camp.

C7Engine/AI/PlayerAI.cs
QuintillusCFC: Ah, I thought there ought to be a method like that but hadn't found it!  Switched!

WildWeazel (approved):

PR #363, opened 2022-09-28
349 - Fix cheap road/rail transport

### LINKED ISSUE #349: Moving onto Railroads is Free, and Moving onto Roads is Discounted
Discovered while playtesting for something else.  09 MP WWII in the Pacific is a good test scenario.

A unit is not charged any movement points when it moves onto a rail tile, even if it was not already on a rail tile.  This applies from non-roaded onto rail as well.

Also, a unit moving onto a road from a non-roaded tile is only charged 1/3 of the movement cost; it should be the full amount.

Great for moving units quickly, but also a bug...

QuintillusCFC:
Moving onto rails is no longer free, and moving onto roads is no longer discounted, unless the source tile also has rail/(roads or rail), respectively.

Also fix barbarians being charged for base terrain cost regardless of roads/rails, and a crash caused by right-clicking on a unit with a defend strategy on a non-city tile (which currently happens for catapults in scenarios that don't support Settlers).

### DIFF
--- a/C7Engine/AI/BarbarianAI.cs
+++ b/C7Engine/AI/BarbarianAI.cs
@@ -27,7 +27,7 @@ namespace C7Engine {
 						if (validTiles.Count == 0) {
 							//This can happen if a barbarian galley spawns next to a 1-tile lake, moves there, and doesn't have anywhere else to go.
 							log.Warning("WARNING: No valid tiles for barbarian to move to");
-							continue;
+							break;
 						}
 						Tile newLocation = validTiles[GameData.rng.Next(validTiles.Count)];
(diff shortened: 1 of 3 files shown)


QuintillusCFC:
Tagging Kright as a potential reviewer, since he added roads initially.

I should also note that when a barbarian moves across multiple roaded tiles, it only animates the last movement, which is a different issue.

Kright (approved):
Looks okay to me. 

Not related to this review, but I think that on line 30 in BarbarianAI is better to use "break" for avoiding infinite loop.

QuintillusCFC:
Yup, I agree.  Switched to `break`, now merging.

Issue #148, opened 2022-03-14
Unit Prototypes should store info about terrain they can move onto


QuintillusCFC:
In MapUnitExtensions.cs's move method, we currently have a check to ensure that units can only move on land.

Unfortunately, that applies to naval units as well.  Hence the infamous "Plains Galley" bug, where your galleys can sail across the prairies, but not across the coasts or seas.  "Prairie Schooner" got more literal.

There has been some debate on how this should be done, notably the possibility of units that have aspects of more than one of the unit types (Air/Land/Sea) in Civ3.  My thinking is that while those, along with the "wheeled" flag, should be indicators of what terrains are allowed, ultimately C7 should allow setting it per-terrain, and should store it that way.  This would allow the creation of units that can travel e.g. on mountains but not in jungle, while also allowing units that can travel in the jungle but not on mountains, in the same scenario.  Perhaps you want Tibet to have special mountain units and Kongo to have special jungle units, but you don't want the Kongolese using their special unit to scale Mt. Everest, or the Tibetan special units to conquer the equatorial area.  It would also simplify any checks after loading the scenario - just check if the target terrain is on the list of valid types.

There's also the matter of the legacy Civ3 naval units having optional "Sinks in Sea" and "Sinks in Ocean" flags, and that those stop applying after a certain period of time.  That is probably beyond the scope of this initial improvement, and could prove to provide some unwanted complexity.  My first thought was, "set the unit prototype to only allow coast, and once safe sea travel is researched, modify the prototype to allow sea, etc.".  But that has the problem that safe sea travel is researched per-civilization.  It also ignores the fact that you can send Galleys on sea and even ocean tiles manually, and the AI will do that as well (but only if they end up on "safe" terrain); this is different than the hard restriction imposed by the "wheeled" flag.  Yeah, we're probably going to need a compatibility layer for that, let's spin that out into a separate issue.

Related to #134 , as that may answer the question of where's the info going to be stored?

Sean-Brown:
Looking at the editor, it seems like units are put into a "class" (land, sea, or air) that determines what kind of moves they can make. Land and sea troops can additionally select terrains that they ignore movement cost on. That can be a place to start from, however I kind of wish these were combinable, for example sea planes which could be air/sea, troop transports that can be land/sea, etc. We can certainly implement it differently while still maintaining backward compatibility, I was just musing about how everything's laid out in the editor and how it presumably affects the logic.

[...]

PR #366, opened 2022-10-03
320 - Fix the overloaded NeighborsWater method.

### LINKED ISSUE #320: Disambiguate Overloaded NeighborsWater Method in Tile.cs
Our NeighborsWater method is currently used for two different purposes, which should have two different logical evaluations, and which quite reasonably both think it is the method they should be calling.  This is mostly on me as I added the first iteration of it, which focused purely on map graphics needs but gave it a name that was very inviting for other purposes.

As a result, right now, it looks like this:

```
foreach (Tile neighbor in getDiagonalNeighbors()) {
//check for whether it's a water tile
```

Aside from questions about whether you consider the "diagonal" tiles to be the ones that share edges or vertices (currently in the code, it means the ones that share edges), the thing that should raise eyebrows is why doesn't it check all the neighboring tiles?  For important questions such as, "can a tile on this city build naval units?", we want to check all neighboring tiles.

The reason it's written the way it is currently is that for the question, "do we want to display large forest/jungle/marsh graphics, or small ones?", the relevant question is "do the neighbors that share an edge with this tile contain water?  if so, use small graphics."  The method was written with this definition of neighboring water in mind, and got picked up for everything else.

So in essence this is a classic equivalent of when you get corruption in your database because one part of the business uses Term A and means one thing, and another part of the business uses Term A and means something else, and IT doesn't realize that Term A can mean two different things, and re-uses the same field.  Like in most of those cases, there were also a few months between the first use of NeighborsWater and the subsequent uses of it, so there was time to forget exactly what it meant in the first place.

I suggest we use the "_any_ neighboring tile is a water tile" definition for NeighborsWater, and have the graphics uses cases use the getDiagonalNeighbors method and LINQ to figure out what they need.  They're the specialized case here.

This is also a bug because if you found a city that only borders water via vertices, it cannot build Galleys; the same applies for barbarian camps that only neighbor water via vertices.

QuintillusCFC:
This fixes the bug where cities/barb camps that were connected to water only along a vertex could not build water units.

Closes #320

### DIFF
--- a/C7GameData/Tile.cs
+++ b/C7GameData/Tile.cs
@@ -79,7 +79,7 @@ namespace C7GameData
 		//Sometimes we care *specifically* about the Coast terrain, e.g. galleys can only move on that terrain, not Sea or Ocean
 		//Those cases should not use this method.
 		public bool NeighborsWater() {
-			foreach (Tile neighbor in getDiagonalNeighbors()) {
+			foreach (Tile neighbor in neighbors.Values) {
 				if (neighbor.baseTerrainType.isWater()) {
 					return true;
 				}
@@ -87,9 +87,14 @@ namespace C7GameData
 			return false;
 		}
 
-		public Tile[] getDiagonalNeighbors() {
-			Tile[] diagonalNeighbors =  { neighbors[TileDirection.NORTHEAST], neighbors[TileDirection.NORTHWEST], neighbors[TileDirection.SOUTHEAST], neighbors[TileDirection.SOUTHWEST]};
-			return diagonalNeighbors;
+		/// <summary>
+		/// Returns neighbors along edges only.
+		/// This is used by some graphics algorithms.
+		/// </summary>
+		/// <returns></returns>
+		public Tile[] getEdgeNeighbors() {
+			Tile[] edgeNeighbors =  { neighbors[TileDirection.NORTHEAST], neighbors[TileDirection.NORTHWEST], neighbors[TileDirection.SOUTHEAST], neighbors[TileDirection.SOUTHWEST]};
+			return edgeNeighbors;
 		}
 
 		public override string ToString()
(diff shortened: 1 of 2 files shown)


C7GameData/Tile.cs
```diff
@@ -79,17 +79,22 @@ public Tile()
 		//Sometimes we care *specifically* about the Coast terrain, e.g. galleys can only move on that terrain, not Sea or Ocean
 		//Those cases should not use this method.
 		public bool NeighborsWater() {
-			foreach (Tile neighbor in getDiagonalNeighbors()) {
+			foreach (Tile neighbor in neighbors.Values) {
```
WildWeazel, line 82: Remember that cities are a special case for diagonal neighbors. Otherwise a diagonal water neighbor only is not considered connected, ie the land connection takes priority. Is this taken into account when there is no city?

QuintillusCFC: Do you mean in terms of trade connections?  I guess I'm not sure exactly what case it is you mean where a diagonal (vertex) neighbor doesn't count for water, other than the graphics case mentioned in the comment.

So far this is only used when deciding what a city can build, whether a barbarian camp spawns boats, and how appealing a city site is to the AI.

WildWeazel: If you have water tiles at N/S and land tiles at E/W, that's an isthmus and sea units can't move between the water tiles. But if N is a city, sea units can move in and out of it. I just want to be sure this method isn't a general case for finding connected water tiles.

QuintillusCFC: Ohhhh, right!  Forgot about that.  Well, the isthmus case _is_ a problem, but it doesn't use this method.  For both barbarian and non-barbarian AIs, the `GetCoastNeighbors` method in `Tile.cs` is used, which doesn't check for isthmuses.  I'm not sure off the top of my head for human moves.  But I have seen AI galleys move across isthmuses.

I don't think this method could be easily misused for that, as it returns a boolean, and our movement/pathing methods are looking for tile objects that can be moved to.  I.e. this method doesn't tell you which neighboring tiles are water, so you can't use it to decide where to move.

Nonetheless, that is a good catch, I wrote it up as #369 .  I'd probably do it after #148 , which I see as the next iteration in improving the accuracy of movement possibilities.

WildWeazel (approved):

PR #367, opened 2022-10-04
241 - Remove citizens from tiles when a city is destroyed.

### LINKED ISSUE #241: Citizens should be removed from tiles when their city is destroyed
Noticed while play-testing.  If a city is farming a nice wheat tile, and then it gets destroyed, and a future city is re-settled there, the wheat tile remains claimed by the other destroyed city, so the new one has to work a different tile.

QuintillusCFC:
Also renamed method to conform with C# standards, and move the destroy logic out to CityInteractions so it can be reused when we add voluntary (right click menu) city destruction.

### DIFF
--- a/C7Engine/EntryPoints/CityInteractions.cs
+++ b/C7Engine/EntryPoints/CityInteractions.cs
@@ -21,5 +21,16 @@ namespace C7Engine
 			tileWithNewCity.cityAtTile = newCity;
 			tileWithNewCity.overlays.road = true;
 		}
+
+		public static void DestroyCity(int x, int y) {
+			Tile tile = EngineStorage.gameData.map.tileAt(x, y);
+			tile.DisbandNonDefendingUnits();
+			while (tile.cityAtTile.size > 0) {
+				tile.cityAtTile.RemoveCitizen();
+			}
+			tile.cityAtTile.owner.cities.Remove(tile.cityAtTile);
+			EngineStorage.gameData.cities.Remove(tile.cityAtTile);
+			tile.cityAtTile = null;
+		}
 	}
 }
--- a/C7Engine/EntryPoints/CityInteractions.cs
+++ b/C7Engine/EntryPoints/CityInteractions.cs
@@ -25,9 +25,7 @@ namespace C7Engine
 		public static void DestroyCity(int x, int y) {
 			Tile tile = EngineStorage.gameData.map.tileAt(x, y);
 			tile.DisbandNonDefendingUnits();
-			while (tile.cityAtTile.size > 0) {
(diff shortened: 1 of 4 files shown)


C7Engine/EntryPoints/CityInteractions.cs
```diff
@@ -21,5 +21,16 @@ public static void BuildCity(int x, int y, string playerGuid, string name)
 			tileWithNewCity.cityAtTile = newCity;
 			tileWithNewCity.overlays.road = true;
 		}
+
+		public static void DestroyCity(int x, int y) {
+			Tile tile = EngineStorage.gameData.map.tileAt(x, y);
+			tile.DisbandNonDefendingUnits();
+			while (tile.cityAtTile.size > 0) {
+				tile.cityAtTile.RemoveCitizen();
```
WildWeazel: seems like we could use a convenience function RemoveCitizens(int n) and maybe RemoveAllCitizens()

C7Engine/EntryPoints/CityInteractions.cs
```diff
@@ -21,5 +21,16 @@ public static void BuildCity(int x, int y, string playerGuid, string name)
...
+			}
+			tile.cityAtTile.owner.cities.Remove(tile.cityAtTile);
+			EngineStorage.gameData.cities.Remove(tile.cityAtTile);
+			tile.cityAtTile = null;
```
WildWeazel, line 31: It seems like a code smell to do each of these separately at this level, but I'm not quite sure how it should be done. I suppose the component model will play a role in that, since there would be a component for handling city data.

C7Engine/EntryPoints/CityInteractions.cs
QuintillusCFC: Added both of them, and it does indeed clean it up a bit.  It also provides a bounds check in the RemoveCitizen(int n) method which protects against problems caused by removing too many citizens, so that is nice.

C7Engine/EntryPoints/CityInteractions.cs
QuintillusCFC: Yeah, the thing is there are multiple places that reference the city object so it is convenient to reference it from various viewpoints, but that means they all have to be cleaned up too.  On the balance, I'd rather have to clean up a few reference but have those different views than to only be able to access cities via `EngineStorage.gameData.cities`.

My intention with this method is that it will be a one-stop-shop for city destruction, so e.g. it can be re-used when we add the right-click "Abandon city" action.  It may move and it may be modified, but by having one place that handles everything, the fact that there are a few references to clean up shouldn't result in forgetting to clean them up.

WildWeazel (approved):

PR #372, opened 2022-10-16
323 - Fix zombie units causing crashes

### LINKED ISSUE #323: Crash due to failing to remove zombie unit from tile

QuintillusCFC:
This fixes a longstanding mysterious crash, which thanks to improved logging and the ability to set a fixed seed, was determined to be caused by zombie units.

These units no longer move around and cause crashes, and should be properly garbage collected the next turn now that they don't cause crashes.

You can use fixed seed 1626442457 to test this solving a crash that happens after turn 18.  See the issue card for full notes.

Please review 357 first!  The reproducible case for this depends on the changes made there.

### DIFF
--- a/C7Engine/AI/BarbarianAI.cs
+++ b/C7Engine/AI/BarbarianAI.cs
@@ -21,7 +21,7 @@ namespace C7Engine {
 			// issue for the barbs but will be for similar loops elsewhere in the AI logic.
 			foreach (MapUnit unit in player.units.ToArray()) {
 				if (UnitIsFreeToMove(unit)) {
-					while (unit.movementPoints.canMove) {
+					while (unit.hitPointsRemaining > 0 && unit.movementPoints.canMove) {
 						//Move randomly
 						List<Tile> validTiles = unit.unitType.categories.Contains("Sea") ? unit.location.GetCoastNeighbors() : unit.location.GetLandNeighbors();
 						if (validTiles.Count == 0) {
--- a/C7Engine/AI/BarbarianAI.cs
+++ b/C7Engine/AI/BarbarianAI.cs
@@ -21,7 +21,7 @@ namespace C7Engine {
 			// issue for the barbs but will be for similar loops elsewhere in the AI logic.
 			foreach (MapUnit unit in player.units.ToArray()) {
 				if (UnitIsFreeToMove(unit)) {
-					while (unit.hitPointsRemaining > 0 && unit.movementPoints.canMove) {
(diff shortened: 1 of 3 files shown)


C7Engine/AI/BarbarianAI.cs
```diff
@@ -21,7 +21,7 @@ public class BarbarianAI {
 			// issue for the barbs but will be for similar loops elsewhere in the AI logic.
 			foreach (MapUnit unit in player.units.ToArray()) {
 				if (UnitIsFreeToMove(unit)) {
-					while (unit.movementPoints.canMove) {
+					while (unit.hitPointsRemaining > 0 && unit.movementPoints.canMove) {
```
WildWeazel: If this is the part that avoids the crash, I feel like it's just masking the problem. Why is the unit still showing up here once it's been killed? Is it specifically units that die before running out of MP, or are there already dead units still in player.units? Either way it seems like a more fundamental problem than is being addressed here.

WildWeazel:
We'd better fix this soon, Halloween is coming up....

C7Engine/AI/BarbarianAI.cs
QuintillusCFC: The unit is removed from player.units when it dies, but it is still in player.units.ToArray() from line 22.  The array is created because otherwise, we would have a ConcurrentModificationException (or the C# equivalent thereof) when removing it from player.units while iterating over that collection.

It is specifically units that die before they run out of MP.

I'm going to push up a change that results in the unit.move method returning false if the unit is killed while moving, allowing the while loop to be broken out of in that case.  It's more code, but means we won't keep iterating on the zombie unit for another `while`  (I'd also like to refactor some of the unit.move and animation mechanics at some point, but that might take past Halloween)

QuintillusCFC:
Self-reviewed after 7 days since last comments, merging (not much time left till Halloween!).

PR #373, opened 2022-10-19
176 - Fix Credit Formatting

### LINKED ISSUE #176: Fix credits formatting
Make the credits screen use Noto Sans font with formatting for bold tags, etc.

QuintillusCFC:
Fixes #176 

This adds bold/italics to the credits, which drastically improves the formatting, and also updates it to be current.

I still might extract the theming into the base theme, since it will be useful elsewhere too, but it's getting late here and the method is ready to be looked at.

Also, anyone who wants to update their attribution in the credits, this is a great time for it!  @Kright , I have just added you via your GitHub username, but can change the attribution to what you prefer if it's different.

An idea for a future PR is making the links in the credits work, which would also allow linking attributions to GitHub accounts/websites/CFC accounts/etc. if desired.  It looks like this is probably doable with a bit more time.


WildWeazel (approved):
Looks nice! Let's remember to add art credits as we go

PR #374, opened 2022-10-22
343 - Cities that grow negatively will now shrink back down to the next size

### LINKED ISSUE #343: Cities that Can't Supply Enough Food Grow Negatively Forever
Best illustrated with a screenshot:

image

Rome grew to size 2 but couldn't supply enough food, so every turn the counter of how long until it grows decreases by 1.  Instead it should shrink to size 1.

QuintillusCFC:
Closes #343 

This is part of the bug-squashing effort.  You can use seed 643368717 to test this; move your Settler a few tiles SW so it can build a city surrounded by desert and one Plains tile, and after 20 turns it'll grow to size 2, when it should show -- city growth and a red pop size in the UI, and the next turn it will shrink back down to size 1 and start growing again.

The last commit is a minor refactor of TurnHandling.cs to break a large method into a few logical parts.  Other than extracting methods, it's the same.  I didn't have to change anything there but it was my first port of call when investigating this and I was reminded that I'd been wanting to split that method up for some time.

Note that if you keep playing 643368717, you'll soon encounter the zombies of #323 .  They seem to be more ravenous as Halloween approaches.

### DIFF
--- a/C7/Map/CityLayer.cs
+++ b/C7/Map/CityLayer.cs
@@ -50,7 +50,7 @@ namespace C7.Map {
 			looseView.DrawTextureRectRegion(cityTexture, screenRect, textRect);
 
 			int turnsUntilGrowth = city.TurnsUntilGrowth();
-			string turnsUntilGrowthText = turnsUntilGrowth == int.MaxValue ? "- -" : "" + turnsUntilGrowth;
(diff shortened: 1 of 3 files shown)


QuintillusCFC:
Self-reviewed after 7 days, made an improvement to the method documentation, merging.

PR #381, opened 2023-01-16
370 - Add Kyriakos's Persian city graphics, fix custom graphics

### LINKED ISSUE #370: Add City Set Graphics
As the second step of adding community graphics, the plan is to add Kyriakos's Persian set (https://forums.civfanatics.com/threads/five-culture-ancient-sets.629287/, third one down in the preview images).

The simplest version of this is so rename it/ship it as the rROMAN graphics that the game currently uses, but if you are feeling more ambitious, you could also set it as the Mediterranean one and import the Civ3 info about which city graphics a civ should use.

### LINKED ISSUE #380: Fix Custom Graphics Not Displaying (Again)
While testing #370 , I noticed that custom graphics weren't working again.  That's weird, right?  #140 was supposed to fix them.

It turns out that the merge-conflict-resolution before merging that had a mistake, where a / in the search path got removed, so it could never find the custom graphics.  Once that was established as the cause it was easy to fix.

QuintillusCFC:
Since these are a different size (600x547) than the base ones (501x380), a code update is also needed.  Most of the extra size is pink space, so I think it's mostly just giving it a bigger possible canvas.

Checking with the Manhattan graphics, arbitrary size city graphics are supported in Civ3, so we should support that as well.  We might need a flag for high-res when we get to that, but can cross that bridge when it's time to do so.

Also fixed custom graphics not displaying since that had been broken in a merge into #140 prior to merging it.

Closes #370 
Closes #380


QuintillusCFC:
Self-reviewed after 13 days; merging.

PR #383, opened 2023-01-31
#382 - Remember install location for next time when it is set.

### LINKED ISSUE #382: Remember Civ3 Install Location When Set Manually
While setting up C7 on my new laptop, I realized that if you set the Civ3 location with the button, it doesn't remember it when re-launched.  You have to have set Civ3_Home (or have it in the Windows registry) for it to stick.

We should have it remember.  Since setting environment variables tends to require administrator privileges, this is likely via setting it in a config file and then checking that config file.

Side note: I have realized that at least with the Steam version, you have to run Civ3 after installing it for it to show up in the registry.  This makes it slightly more likely to happen on Windows, although this will still mostly benefit non-Windows systems.

QuintillusCFC:
Closes #382 

This results in not having to either set the Civ3_Home environment variable, or manually select the install location every time.  A nice convenience if you don't have Civ3 in the Windows Registry and aren't familiar with environment variables.


WildWeazel (approved):

WildWeazel:
@QuintillusCFC is this still relevant? I seem to remember something related to this going in already

QuintillusCFC:
> @QuintillusCFC is this still relevant? I seem to remember something related to this going in already

Yeah, it is.  You're thinking of https://github.com/C7-Game/Prototype/pull/386, which gives auto-detect on Linux/Mac if the Steam version of Civ3 is findable.

This handles the case where Civ3 isn't findable, regardless of OS.  Rather than setting the Civ3Home environment variable, which is a bit of a tough ask for a non-technical user (I'm not sure I know any non-developers, or at least developer-adjacent professionals, who would know how to do that), you can just select the location once in the UI, and with this update, it'll remember that in the future.

QuintillusCFC:
I'm pretty sure I hadn't merged this because there was a conflict, and I haven't been doing much C7 work since then.  Resolved it tonight (it was easy), now merging.

PR #384, opened 2023-02-04
#172 - City Scene

### LINKED ISSUE #172: Create a City UI scene
Create a single reusable and scriptable Godot scene to display a more faithful city UI: the shaded, color-coded box containing the size, name, growth time, production project and time, capital icon, and barracks/harbor/airport icons (some discussion may be needed about how to generalize those), with a container for the city graphic and overlay animation. The scene should have functions to update/set each of these elements either programmatically or via editor controls.

This may obviate #171

QuintillusCFC:
Closes #172 

These changes break our city-drawing code into its own scene, and also break out the city label into its own scene.

Along the way I added a couple small optimizations, such as only redrawing the label background (which involves a lot of operations) if the label background has changed.

I also added in making it so only the capital has the capital star, and logging of turn times.

### DIFF
--- a/C7/Map/CityLabelScene.cs
+++ b/C7/Map/CityLabelScene.cs
@@ -1,0 +1,211 @@
+using System;
+using C7GameData;
+using ConvertCiv3Media;
+using Godot;
+using Serilog;
+using Serilog.Events;
+
+namespace C7.Map {
+	public class CityLabelScene : Node2D {
+		private ILogger log = LogManager.ForContext<CityLabelScene>();
+
+		private City city;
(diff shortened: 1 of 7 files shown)


pcen (approved):

WildWeazel:
can this go in now or wait (ha) for Godot 4?

QuintillusCFC:
It can go in now, and may require some changes for Godot 4, particularly in the files in the C7/Map folder which are Godot Node2Ds.

I'll go ahead and merge it since Paul has approved it, with that note.

PR #396, opened 2023-02-11
139 - Smarter Barbarians

### LINKED ISSUE #139: Barbarians Try to Attack Your Units and Destroy Your Roads and Cities
Instead of wandering around aimlessly, barbarians should seek loot, from your and other AI civilizations.

They should walk towards your cities, and generally try to be as much as a nuisance as possible.

This could be with individual units, or with small groups of them.  I like the idea of having it vary a bit by tribe - maybe this tribe just attacks Rambo style, but that other tribe attacks less frequently but with some coordination.  Kind of open ended, but more interesting than "beeline to nearest city" is encouraged.  If they find the city is defended, but some Iron is available to pillage?  I'd take the easy target if I were a barbarian.

QuintillusCFC:
Closes #139 

This is a set of changes to make the barbarians more dangerous.  More changes can be added to make them more dangerous, and may be as part of this PR, or later.

It also organizes them into tribes, with the idea of various tribes having different priorities.

So far the changes center around making them more opportunistic:

 - If a barbarian unit starts its turn next to an undefended non-combat unit, goodbye undefended unit
 - If a barbarian unit starts its turn next to a non-barbarian unit it thinks it has a chance of beating, it will attack

Corollary, the barbarians aren't the best at figuring out if they think they have a chance of winning yet.  I added a very simple risk calculator that takes into account defensive bonuses, but not the round-based nature of combat, meaning the barbarians will be more optimistic than they should be.

Long-term, I envision adding a more accurate combat odds calculator, and perhaps a multi-unit combat calculator (do my five units have a chance of taking that city?).  Which combat odds calculator the barbarians (and AI) use could vary based on difficulty level.

### DIFF
--- a/C7Engine/AI/BarbarianAI.cs
+++ b/C7Engine/AI/BarbarianAI.cs
@@ -16,35 +16,42 @@ namespace C7Engine {
 				throw new System.Exception("Barbarian AI can only play barbarian players");
 			}
 
-			// Copy unit list into temporary array so we can remove units while iterating.
-			// TODO: We also need to handle units spawned during the loop, e.g. leaders, armies, enslaved units. This is not so much an
-			// issue for the barbs but will be for similar loops elsewhere in the AI logic.
-			foreach (MapUnit unit in player.units.ToArray()) {
-				if (UnitIsFreeToMove(unit)) {
-					while (unit.movementPoints.canMove) {
-						//Move randomly
-						List<Tile> validTiles = unit.unitType.categories.Contains("Sea") ? unit.location.GetCoastNeighbors() : unit.location.GetLandNeighbors();
-						if (validTiles.Count == 0) {
-							//This can happen if a barbarian galley spawns next to a 1-tile lake, moves there, and doesn't have anywhere else to go.
-							log.Warning("WARNING: No valid tiles for barbarian to move to");
-							break;
-						}
-						Tile newLocation = validTiles[GameData.rng.Next(validTiles.Count)];
-						//Because it chooses a semi-cardinal direction at random, not accounting for map, it could get none
-						//if it tries to move e.g. north from the north pole.  Hence, this check.
-						if (newLocation != Tile.NONE) {
-							log.Debug("Moving barbarian at " + unit.location + " to " + newLocation);
-							if (!unit.move(unit.location.directionTo(newLocation))) {
+			BarbarianPlayer barbs = (BarbarianPlayer)player;
+
+			foreach (BarbarianTribe tribe in barbs.getTribes())
+			{
+				// Copy unit list into temporary array so we can remove units while iterating.
+				foreach (MapUnit unit in tribe.GetUnits().ToArray()) {
+					if (UnitIsFreeToMove(unit)) {
+						while (unit.movementPoints.canMove) {
+							//Move randomly
+							List<Tile> validTiles = unit.unitType.categories.Contains("Sea")
+								? unit.location.GetCoastNeighbors()
+								: unit.location.GetLandNeighbors();
+							if (validTiles.Count == 0) {
+								//This can happen if a barbarian galley spawns next to a 1-tile lake, moves there, and doesn't have anywhere else to go.
+								log.Warning("WARNING: No valid tiles for barbarian to move to");
 								break;
 							}
-						} else {
-							//Avoid potential infinite loop.
-							break;
+
+							Tile newLocation = validTiles[GameData.rng.Next(validTiles.Count)];
+							//Because it chooses a semi-cardinal direction at random, not accounting for map, it could get none
+							//if it tries to move e.g. north from the north pole.  Hence, this check.
+							if (newLocation != Tile.NONE) {
+								log.Debug("Moving barbarian at " + unit.location + " to " + newLocation);
+								if (!unit.move(unit.location.directionTo(newLocation))) {
+									break;
+								}
+							} else {
+								//Avoid potential infinite loop.
+								break;
+							}
 						}
 					}
 				}
 			}
 		}
+
 		private static bool UnitIsFreeToMove(MapUnit unit)
 		{
 			if (!unit.location.hasBarbarianCamp) {
--- a/C7Engine/AI/BarbarianAI.cs
+++ b/C7Engine/AI/BarbarianAI.cs
@@ -41,7 +41,7 @@ namespace C7Engine {
 
 							//Check if there are any undefended units that can be taken!
 							foreach (Tile tile in validTiles) {
-								if (tile.unitsOnTile.Exists(mapUnit => UndefendedUnit(mapUnit))) {
+								if (tile.unitsOnTile.Exists(mapUnit => IsUndefendedUnit(mapUnit))) {
 									bool alive = unit.move(unit.location.directionTo(tile));
 									// TODO: Restructure so we can avoid gotos.
 									goto nextMovementPoint;
@@ -85,7 +85,7 @@ nextUnit: ;
 			}
 		}
 
-		private static bool UndefendedUnit(MapUnit unit) {
+		private static bool IsUndefendedUnit(MapUnit unit) {
 			if (unit.owner.isBarbarians) {
 				return false;
 			}
--- a/C7Engine/AI/BarbarianAI.cs
+++ b/C7Engine/AI/BarbarianAI.cs
@@ -34,6 +34,15 @@ namespace C7Engine {
 								break;
 							}
 
+							//Check if there are any undefended units that can be taken!
+							foreach (Tile tile in validTiles) {
+								if (tile.unitsOnTile.Exists(mapUnit => UndefendedUnit(mapUnit))) {
+									unit.move(unit.location.directionTo(tile));
+									// TODO: Restructure so we can avoid gotos.
+									goto nextMovementPoint;
+								}
+							}
+
 							Tile newLocation = validTiles[GameData.rng.Next(validTiles.Count)];
 							//Because it chooses a semi-cardinal direction at random, not accounting for map, it could get none
 							//if it tries to move e.g. north from the north pole.  Hence, this check.
@@ -46,10 +55,21 @@ namespace C7Engine {
 								//Avoid potential infinite loop.
 								break;
 							}
+nextMovementPoint: ;
 						}
 					}
 				}
 			}
+		}
+
+		private static bool UndefendedUnit(MapUnit unit) {
+			if (unit.owner.isBarbarians) {
+				return false;
+			}
+
+			return unit.location.unitsOnTile.Count(mapUnit => {
+				return mapUnit.unitType.defense > 0;
+			}) == 0;
 		}
 
 		private static bool UnitIsFreeToMove(MapUnit unit)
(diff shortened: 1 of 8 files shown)


C7Engine/MapUnitExtensions.cs
```diff
@@ -430,12 +436,23 @@ public static void disband(this MapUnit unit)
...
+			if (!player.isBarbarians) {
+				if (player.units.Contains(unit)) {
+					player.units.Remove(unit);
+					break;
+				}
+			} else {
+				foreach (BarbarianTribe tribe in ((BarbarianPlayer)(player)).getTribes()) {
+					if (tribe.GetUnits().Contains(unit)) {
+						tribe.RemoveUnit(unit);
+						goto endDisband;
```
WildWeazel: Yikes, I didn't know C# even supported `goto`. Can't this just be a `return`?

C7Engine/MapUnitExtensions.cs
```diff
@@ -430,12 +436,23 @@ public static void disband(this MapUnit unit)
 		// EngineStorage.animTracker.endAnimation(unit, false);   TODO: Must send message instead of call directly
 		unit.location.unitsOnTile.Remove(unit);
 		gameData.mapUnits.Remove(unit);
-		foreach(Player player in gameData.players)
+		foreach (Player player in gameData.players)
 		{
-			if (player.units.Contains(unit)) {
-				player.units.Remove(unit);
+			if (!player.isBarbarians) {
```
WildWeazel: More broadly though, a `MapUnit` should know which player it belongs to if it doesn't already, and `Player` should be able to handle disbanding its own unit (with `BarbarianPlayer` providing an override) rather than having to do this search here.

C7Engine/MapUnitExtensions.cs
```diff
@@ -295,6 +295,12 @@ public static void OnEnterTile(this MapUnit unit, Tile tile)
 		if (tile.hasBarbarianCamp && (!unit.owner.isBarbarians)) {
 			tile.DisbandNonDefendingUnits();
 			EngineStorage.gameData.map.barbarianCamps.Remove(tile);
+			BarbarianPlayer barbarians = (BarbarianPlayer)EngineStorage.gameData.players.Find(player => player.isBarbarians);
+			foreach (BarbarianTribe tribe in barbarians.getTribes()) {
+				if (tribe.GetCamps().Contains(tile)) {
+					tribe.RemoveCamp(tile);
```
WildWeazel, line 301: eventually, when the components are in place... this kind of thing should be event based

C7Engine/AI/BarbarianAI.cs
```diff
@@ -11,40 +12,89 @@ public class BarbarianAI {
 
 		private ILogger log = Log.ForContext<BarbarianAI>();
 
+		// Threshold at which barbarians take risky actions.
+		// In the future this may vary based on barbarian settings, and perhaps the situation
+		const double COLLEGE_TRY_THRESHOLD = 1.0 / 3;
```
WildWeazel, line 17: just curious... college?

C7Engine/AI/BarbarianAI.cs
```diff
@@ -11,40 +12,89 @@ public class BarbarianAI {
...
+							}
+nextMovementPoint: ;
 						}
 					}
+nextUnit: ;
 				}
 			}
 		}
+
+		private static bool UndefendedUnit(MapUnit unit) {
```
WildWeazel: Let's stick with a Is/Has naming convention for booleans... `IsUndefendedUnit`

C7Engine/AI/BarbarianAI.cs
```diff
@@ -11,40 +12,89 @@ public class BarbarianAI {
...
+			if (unit.owner.isBarbarians) {
+				return false;
+			}
+
+			return unit.location.unitsOnTile.Count(mapUnit => {
```
WildWeazel, line 93: seems like this would be a good utility function `GetDefenderCount`

C7Engine/AI/UnitAI/CombatOdds.cs
```diff
@@ -0,0 +1,27 @@
+using System.Collections.Generic;
+using C7GameData;
+
+namespace C7Engine.AI.UnitAI {
+	/**
+	 * This is the start of a combat odds calculator.
+	 * Its primary purpose is to make the AI well-informed about the risks it is taking, so it can defeat the human
+	 * more easily.
+	 * Its secondary purpose is to allow an eventual Civ4-style UI giving the human a helpful tooltip about the odds,
```
WildWeazel, line 9: I agree, so there is a better place for this? Do we have utils in the game-logic layer?

C7Engine/AI/UnitAI/CombatOdds.cs
```diff
@@ -0,0 +1,27 @@
...
+	public class CombatOdds {
+		public static double OddsOfVictory(MapUnit attacker, MapUnit defender) {
+			//Yanked from MapUnitExtension's `fight` method
+			IEnumerable<StrengthBonus> attackBonuses  = attacker.ListStrengthBonusesVersus(defender, CombatRole.Attack , attacker.facingDirection),
+				defenseBonuses = defender.ListStrengthBonusesVersus(attacker, CombatRole.Defense, attacker.facingDirection);
+
+			double attackerStrength = attacker.unitType.attack  * StrengthBonus.ListToMultiplier(attackBonuses),
+				defenderStrength = defender.unitType.defense * StrengthBonus.ListToMultiplier(defenseBonuses);
+
+			return attackerStrength / (attackerStrength + defenderStrength);
```
WildWeazel, line 24: This seems like per-hit odds, isn't it?

C7Engine/AI/BarbarianAI.cs
```diff
@@ -11,40 +12,89 @@ public class BarbarianAI {
...
 							}
-						} else {
-							//Avoid potential infinite loop.
-							break;
+
+							//Check if there are any undefended units that can be taken!
+							foreach (Tile tile in validTiles) {
+								if (tile.unitsOnTile.Exists(mapUnit => UndefendedUnit(mapUnit))) {
+									bool alive = unit.move(unit.location.directionTo(tile));
+									// TODO: Restructure so we can avoid gotos.
```
WildWeazel, line 46: Right, these are making me nervous. To me this suggests breaking up the nested loops into functions that return a boolean or nullable object and check the return value to decide the next step

C7Engine/AI/BarbarianAI.cs
```diff
@@ -11,40 +12,89 @@ public class BarbarianAI {
...
+										if (alive) {
+											goto nextMovementPoint;
+										}
+										goto nextUnit;
+									}
+								}
+							}
+
+							Tile newLocation = validTiles[GameData.rng.Next(validTiles.Count)];
+							//Because it chooses a semi-cardinal direction at random, not accounting for map, it could get none
```
WildWeazel, line 69: Do the neighbor functions not account for map edges?

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -81,40 +81,41 @@ private static bool PlayPlayerTurns(GameData gameData, bool firstTurn)
...
-					tile.unitsOnTile.Add(newUnit);
-					gameData.mapUnits.Add(newUnit);
-					barbPlayer.units.Add(newUnit);
-					log.Debug("New barbarian galley added at " + tile);
+			BarbarianPlayer barbPlayer = (BarbarianPlayer)gameData.players.Find(player => player.isBarbarians);
+			foreach (BarbarianTribe tribe in barbPlayer.getTribes()) {
+				foreach (Tile tile in tribe.GetCamps()) {
+					//7% chance of a new barbarian.  Probably should scale based on barbarian activity.
+					int result = GameData.rng.Next(100);
+					log.Verbose("Random barb result = " + result);
```
WildWeazel, line 89: If this is happening at each camp maybe include the location

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -81,40 +81,41 @@ private static bool PlayPlayerTurns(GameData gameData, bool firstTurn)
...
+						newUnit.unitType = gameData.barbarianInfo.barbarianSeaUnit;
+						newUnit.experienceLevelKey = gameData.defaultExperienceLevelKey;
+						newUnit.experienceLevel = gameData.defaultExperienceLevel;
+						newUnit.hitPointsRemaining = 3;
+						newUnit.isFortified = true; //todo: hack for unit selection
+
+						tile.unitsOnTile.Add(newUnit);
+						gameData.mapUnits.Add(newUnit);
+						tribe.AddUnit(newUnit);
+						log.Debug("New barbarian galley added at " + tile);
```
WildWeazel, line 117: sounds like a call for factory methods!

WildWeazel (changes requested):

C7Engine/MapUnitExtensions.cs
QuintillusCFC: You know, it actually can in this case.

I wound up with `goto` as an alternative to Java's ability to break out of an outer for loop, e.g.:

```
outer: foreach (someVar in outerCollection): {
    //find corresponding innerCollection
    foreach (innerVar in innerCollection): {
        if (innerVar.shouldBreakLoop()) {
            break outer;  //Only supported in Java, not C#
        }
    }
}
//do some additional tasks
```

This is the idiom that I don't know how to C#-ize, and which leads to my uses of `goto`.  In this case there are no additional tasks, so `return` works, but in most of the other `goto` uses I've added, there are additional tasks once the outer loop is broken.

What is the C# way of doing this?  I suppose I could add an additional variable to set and check at the end of the outer loop, and if it's true, break out of it?  But I'm not sure that's actually any cleaner than `goto`.

C7Engine/MapUnitExtensions.cs
QuintillusCFC: I agree.  I will see if I can find a better way to implement this.

(It's also cool to see comments on my pull request!)

C7Engine/AI/BarbarianAI.cs
QuintillusCFC: Yeah, it's a reference to this idiom in English: https://www.merriam-webster.com/dictionary/give%20it%20the%20old%20college%20try

Chosen because the barbarians will (currently) make an attempt at winning even if the odds aren't great, which is how I've often seen that idiom used.  Going to give something a try but not necessarily expecting a lot of success.  Maybe it's a bit too local of an idiom?  I have lived my entire life in either colleges or college towns...

C7Engine/MapUnitExtensions.cs
QuintillusCFC: This has now been refactored and is part of BarbarianAI and is a break.

C7Engine/MapUnitExtensions.cs
QuintillusCFC: This has now been refactored as suggested.  Surprisingly we only had 7 virtual functions previously.

C7Engine/AI/BarbarianAI.cs
QuintillusCFC: Changed.

PR #405, opened 2023-05-30
Godot4: Use Godot Action System for Input

pcen:
the actual dif is ~500 lines (change rewrites the sample save file and 400 lines of whitespace changes in 1 file)

# use Godot action keybindings for input:
- before input buttons were hardcoded to Godot.Key.<key>, now actions triggered by input are determined by action keybindings in Godot (Project -> Project Settings... -> Input Map) queried by Godot `Input.` methods
- simplifies Unit Buttons: instead of sending a signal that calls into Game, Unit Buttons trigger actions also through Godot `Input` class
- for consistency, unit actions use the same names as corresponding Godot keybinding action names, defined in C7GameData/Actions.cs (ie. r key is mapped to the "unit_build_road" action in Godot, and now the worker prototype actions contains "unit_build_road" instead of "buildRoad")

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -93,7 +93,7 @@ public partial class Game : Node2D
 			}
 
 			Toolbar = GetNode<Control>("CanvasLayer/ToolBar/MarginContainer/HBoxContainer");
-			Player = GetNode<CharacterBody2D>("CharacterBody2D");
+			// Player = GetNode<CharacterBody2D>("CharacterBody2D");
 			//TODO: What was this supposed to do?  It throws errors and occasinally causes crashes now, because _OnViewportSizeChanged doesn't exist
 			// GetTree().Root.Connect("size_changed",new Callable(this,"_OnViewportSizeChanged"));
 
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -382,26 +382,25 @@ public partial class Game : Node2D {
 							new RightClickTileMenu(this, tile).Open(eventMouseButton.Position);
 
 						string yield = tile.YieldString(controller);
-						//These GD.Print statements are debugging prints for developers to see info about the tile
-						//For now I'm leaving them as GD.Print.  Could revisit this later.
-						GD.Print($"({tile.xCoordinate}, {tile.yCoordinate}): {tile.overlayTerrainType.DisplayName} {yield}");
+						log.Debug($"({tile.xCoordinate}, {tile.yCoordinate}): {tile.overlayTerrainType.DisplayName} {yield}");
 
 						if (tile.cityAtTile != null) {
 							City city = tile.cityAtTile;
-							GD.Print($"  {city.name}, production {city.shieldsStored} of {city.itemBeingProduced.shieldCost}");
+							log.Debug($"  {city.name}, production {city.shieldsStored} of {city.itemBeingProduced.shieldCost}");
 							foreach (CityResident resident in city.residents) {
-								GD.Print($"  Resident working at {resident.tileWorked}");
+								log.Debug($"  Resident working at {resident.tileWorked}");
 							}
 						}
 
 						if (tile.unitsOnTile.Count > 0) {
 							foreach (MapUnit unit in tile.unitsOnTile) {
-								GD.Print("  Unit on tile: " + unit);
-								GD.Print("  Strategy: " + unit.currentAIData);
+								log.Debug("  Unit on tile: " + unit);
+								log.Debug("  Strategy: " + unit.currentAIData);
 							}
 						}
-					} else
-						GD.Print("Didn't click on any tile");
+					} else {
+						log.Debug("Didn't click on any tile");
+					}
 				}
 			}
 		} else if (@event is InputEventMouseMotion eventMouseMotion && IsMovingCamera) {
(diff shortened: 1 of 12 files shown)


C7/Text/c7-static-map-save.json
WildWeazel, line 1: just checking, was this file supposed to be rewritten?

C7Engine/MapUnitExtensions.cs
WildWeazel, line 1: Looks like this file was reformatted. We need to come up with some IDE settings to standardize the style. Are there any changes here other than the action names?

C7/Game.cs
WildWeazel, line 1: same for this file?

C7/Text/c7-static-map-save.json
pcen: yes, the only change is consistent naming of unit prototype capabilities

C7Engine/MapUnitExtensions.cs
pcen: I'm using vscode and it formats based on the .editorconfig file which was added [here](https://github.com/C7-Game/Prototype/pull/199). The only non-formatting change is the action names

C7/Game.cs
pcen: the main change in Game.cs is that input is handled differently - the godot docs say input actions should be handled in _Process (which in turn calls processActions). Since anywhere that an action can be triggered - including from unit buttons - triggers an action, they're now all processed in Game.cs

C7/Game.cs
```diff
@@ -580,107 +443,159 @@ public override void _UnhandledInput(InputEvent @event)
...
+			} else {
+				mapView.setCameraZoomFromMiddle(0.5f);
+				VSlider slider = GetNode<VSlider>("CanvasLayer/SlideOutBar/VBoxContainer/Zoom");
+				slider.Value = 0.5f;
+			}
+		}
+
+		if (Input.IsActionJustPressed("toggle_animations")) {
+			SetAnimationsEnabled(false);
+		} else if (Input.IsActionJustReleased("toggle_animations")) {
```
WildWeazel: Is this part now redundant with line 428?

C7/Game.cs
```diff
@@ -580,107 +443,159 @@ public override void _UnhandledInput(InputEvent @event)
...
+		}
+
+		if (Input.IsActionJustPressed(C7Action.UnitSentryEnemyOnly)) {
+			// unimplemented
+		}
+
+		if (Input.IsActionJustPressed(C7Action.UnitBuildCity) && CurrentlySelectedUnit.canBuildCity()) {
 			using (var gameDataAccess = new UIGameDataAccess()) {
 				MapUnit currentUnit = gameDataAccess.gameData.GetUnit(CurrentlySelectedUnit.guid);
+				GD.Print(currentUnit.Describe());
```
WildWeazel: Use Serilog instead of GD.Print if this is still needed

C7/Game.cs
```diff
@@ -580,107 +443,159 @@ public override void _UnhandledInput(InputEvent @event)
...
-		this.setSelectedUnit(UnitInteractions.getNextSelectedUnit(gameData));
-	}
+	// Handle Godot keybind actions
+	private void processActions() {
 
-	private void setGoToMode(bool isOn)
-	{
-		inUnitGoToMode = isOn;
-	}
+		if (Input.IsActionJustPressed("end_turn") && !this.HasCurrentlySelectedUnit()) {
```
WildWeazel: Can all these action names be added to C7Action?

C7/Text/c7-static-map-save.json
WildWeazel: I figured out how to show the side-by-side diff without whitespace - much cleaner comparison :)

C7Engine/MapUnitExtensions.cs
WildWeazel: Great, I'll make a task if there's not one to reformat everything at once... after we decide on the formatting. I have opinions :)

C7/Game.cs
```diff
@@ -436,108 +404,12 @@ public override void _UnhandledInput(InputEvent @event)
...
-			{
-				SetAnimationsEnabled(false);
-			}
-			else if (eventKeyDown.Keycode == Godot.Key.O && eventKeyDown.ShiftPressed && eventKeyDown.IsCommandOrControlPressed() && eventKeyDown.AltPressed) {
+		} else if (@event is InputEventMouseMotion eventMouseMotion && IsMovingCamera) {
+			GetViewport().SetInputAsHandled();
+			mapView.cameraLocation += OldPosition - eventMouseMotion.Position;
+			OldPosition = eventMouseMotion.Position;
+		} else if (@event is InputEventKey eventKeyDown && eventKeyDown.Pressed) {
+			if (eventKeyDown.Keycode == Godot.Key.O && eventKeyDown.ShiftPressed && eventKeyDown.IsCommandOrControlPressed() && eventKeyDown.AltPressed) {
```
WildWeazel, line 411: I don't yet understand how this API works but here we have a few hardcoded Keycodes and elsewhere there are references to C7Action. Can these be replaced with new actions?

C7/Game.cs
pcen: it is, thanks for catching. I removed the else if condition on 428

C7/Game.cs
pcen: updated all the string literals with C7Action string constants. I didn't add an action for this since it's a debug feature for development

WildWeazel (approved):

PR #407, opened 2023-06-21
[godot4] Use Godot TileMap for Rendering

pcen:
There is a larger performance overhead for rendering the tilemap compared to how the map is currently rendered, but the frame rate does not decrease nearly as much when more of the map becomes visible

The following features are added in pull requests based on this branch:

_https://github.com/C7-Game/Prototype/pull/408 adds units to this branch - merged_
_https://github.com/C7-Game/Prototype/pull/412 adds cities - merged_
_https://github.com/C7-Game/Prototype/pull/419 adds grid view - merged_
_https://github.com/C7-Game/Prototype/pull/421 adds fog of war - merged_
_https://github.com/C7-Game/Prototype/pull/422 adds horizontal wrapping - merged_


pcen:
I'm going to merge this unless there's any objections (plus it can always be reverted). Refactoring large modules to use Godot 4 features involves big diffs (in this case, across all PRs about 2000 and it's still not done). I can't feasibly break this change into, say, 10 different 200 line PRs that are easily reviewable for two reasons
- it takes maybe 800 lines to get just base terrain rendering, so I can't think of smaller incremental changes to get tile maps working
- even with the 3 current PRs based on the previous branch, rebasing if something in the base branch (godot4) is difficult

Since it's logistically too difficult (for me at least) to keep adding features or refactoring that depend on these changes to get merged I'm effectively blocked in the mean time. I'm happy to wait for reviews, but if merging stale self-reviewed PRs is acceptable I'd like to do so since I'd love to make more progress with moving to godot 4

QuintillusCFC:
Merging stale self-reviewed PRs is acceptable.  There's been some debate of waiting 1 week versus 2 (I'm partial to 1, as it can slow things down too much otherwise).

That said, I do have some concerns about replacing the `godot4` branch with this one until some more of the incremental changes are done.  I still need to check out 408/412 as they may address my concerns, but right now `godot4` is getting kind-of close to parity with the base `Development` (Godot 3) branch, more so than I'd expected.  But the TileMap version isn't entirely there yet.

Some of this is related to units and likely fixed in 408/412, some of it is the whole map being visible (no fog of war yet).  I also crashed it while pressing Ctrl+G to turn on the grid, with this error:

```
Stack overflow.
Repeat 130585 times:
--------------------------------
   at C7.Map.MapView.get_showGrid()
--------------------------------
   at Game.processActions()
   at Game._Process(Double)
   at Godot.Node.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name ByRef, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant ByRef)
   at Godot.CanvasItem.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name ByRef, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant ByRef)
   at Godot.Node2D.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name ByRef, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant ByRef)
   at Game.InvokeGodotClassMethod(Godot.NativeInterop.godot_string_name ByRef, Godot.NativeInterop.NativeVariantPtrArgs, Godot.NativeInterop.godot_variant ByRef)
   at Godot.Bridge.CSharpInstanceBridge.Call(IntPtr, Godot.NativeInterop.godot_string_name*, Godot.NativeInterop.godot_variant**, Int32, Godot.NativeInterop.godot_variant_call_error*, Godot.NativeInterop.godot_variant*)

```

I checked base `godot4` and the grid works as expected there so it's something in the switch to TileMap.  Grids are by their nature at the very edge of tiles...

I'm curious about the performance not decreasing as much when more of the map becomes visible.  Do you mean revealed in-game, or as you use a larger view (e.g. 4K versus 720p), or both, or something different?  It would be interesting to see some stats on the relative performance, both the increased overhead and the improvements, if you have them.

QuintillusCFC:
I should note, it's just replacing `godot4` that I have concerns with.  Kind of like how we want `godot4` to reach 98%+ parity with `Development` before making it the default, IMO before switching to tilemaps they should be at 98%+ parity with the existing rendering.  Most likely, this would mean making `pcen/use-tilemaps` the base for tilemaps, and merging the other tilemap-based changes into it, until as a whole the tilemaps version is as good or better than the non-tilemaps version.

pcen:
> I should note, it's just replacing `godot4` that I have concerns with. Kind of like how we want `godot4` to reach 98%+ parity with `Development` before making it the default, IMO before switching to tilemaps they should be at 98%+ parity with the existing rendering. Most likely, this would mean making `pcen/use-tilemaps` the base for tilemaps, and merging the other tilemap-based changes into it, until as a whole the tilemaps version is as good or better than the non-tilemaps version.

Yeah I guess this is where I'm torn, because as you mentioned, for example, the grid crashing is fixed in 408/412 (not sure which off the top of my head). If I merge all of the TileMap changes into this branch so it's ready to replace the base godot4 branch I'd expect a +/- 2500 line PR. Maybe it's better anyways since each 'incremental' PR is so large, at least someone can play-test the full change.

As for performance, by "more map" I mean that as more tiles become visible (at the same resolution) the performance of godot 3.5 drops more. In my case (a 2k monitor as the display for an m1 macbook air) when enough tiles are visible to fill the entire monitor without any fog of war I get ~60fps on 3.5 and ~200fps on 4.x with TileMap. But it would probably be worth doing a slightly more rigorous comparison

QuintillusCFC:
> > I should note, it's just replacing `godot4` that I have concerns with. Kind of like how we want `godot4` to reach 98%+ parity with `Development` before making it the default, IMO before switching to tilemaps they should be at 98%+ parity with the existing rendering. Most likely, this would mean making `pcen/use-tilemaps` the base for tilemaps, and merging the other tilemap-based changes into it, until as a whole the tilemaps version is as good or better than the non-tilemaps version.
> 
> Yeah I guess this is where I'm torn, because as you mentioned, for example, the grid crashing is fixed in 408/412 (not sure which off the top of my head). If I merge all of the TileMap changes into this branch so it's ready to replace the base godot4 branch I'd expect a +/- 2500 line PR. Maybe it's better anyways since each 'incremental' PR is so large, at least someone can play-test the full change.
> 
> As for performance, by "more map" I mean that as more tiles become visible (at the same resolution) the performance of godot 3.5 drops more. In my case (a 2k monitor as the display for an m1 macbook air) when enough tiles are visible to fill the entire monitor without any fog of war I get ~60fps on 3.5 and ~200fps on 4.x with TileMap. But it would probably be worth doing a slightly more rigorous comparison

The practice I've found that works best for large changesets of a branch that won't realistically be ready in one PR due to size, is to create a new branch as a base, and review/test incrementally, using that feature base branch.  Then when the feature is all done, the with-all-the-changes base branch only needs a basic regression test; its constituent parts have all been reviewed.  Basically that is to make sure everything plays nicely together and that there aren't any files missing from the commits (which in theory should have been caught already anyway, and usually that is the case).

That is quite a significant increase.  I'm not sure we really have any FPS benchmarks at this point; most of the performance effort has gone towards AI move calculations.  But almost 3.5x as good is a nice boost indeed.

Also good to know it runs on an M1; we didn't have any confirmation of that previously.  I just acquired one for iOS development, but haven't set up C7 on it yet, and I don't plan to make it my main machine anyway.  That means we now have someone who regularly runs each of the "most likely" platforms; me on Windows, you on Apple Silicon, WildWeazel on Linux, and Jim (inactive) on Intel Mac.

pcen:
I think that definitely makes sense, this branch is essentially the base for tilemaps
https://github.com/C7-Game/Prototype/pull/408 adds units to this branch
#412 adds cities to #408
#419 adds grid view
#420 adds fog of war
#422 adds horizontal wrapping

once all of these are merged into this branch (and #412 into #408) map rendering should be on par with the current gd4 map

WildWeazel:
I started reviewing this a while back but obviously it's been growing. What files/commits are native to this PR?

pcen:
> I started reviewing this a while back but obviously it's been growing. What files/commits are native to this PR?

The diff in this PR is basically the minimal changes required to render terrain using TileMaps. All the other PRs I outlined in my last comment are based on this branch (or some are based on each-other), but all reintroduce features on top of this diff. Once approved / merged, I'm merging them into this branch, which eventually will be one huge diff to merge into the godot4 branch

WildWeazel:
Right, I'd just like to review everything exactly once and I'm not sure how to get a clean diff without the other branches. Was there a last commit directly to this branch before merges started?

pcen:
I believe commit 8e3504174861ae5bb23d6d24f133688aa717d3eb is the last commit before other branches are merged

C7/Map/Civ3TerrainTileset.cs
```diff
@@ -0,0 +1,103 @@
+using Godot;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace C7.Map {
+
+	class TerrainPcx {
+		private static Random prng = new Random();
```
WildWeazel, line 9: I think we have a "master" RNG somewhere in the game data so that everything can be reproduced from the same seed. If that's not sensible to import here, let's put it somewhere that is.

C7/Map/Civ3TerrainTileset.cs
```diff
@@ -0,0 +1,103 @@
...
+		}
+		private int abcIndex(string terrain) {
+			List<int> indices = new List<int>();
+			for (int i = 0; i < abc.Count(); i++) {
+				if (abc[i] == terrain) {
+					indices.Add(i);
+				}
+			}
+			return indices[prng.Next(indices.Count)];
+		}
```
WildWeazel, line 33: `indices = abc.Where(i => i == terrain).Select(i => i.Index).toList()` or something like that
If selecting a random item from a collection will be a common operation, we may want an extension method for it

C7/OldMapView.cs
WildWeazel, line 1: Is this file still going to be used or is it transitory?

C7/Map/Civ3TerrainTileset.cs
WildWeazel, line 1: Since this file is specific to Civ3 file formats, I'm thinking it should belong in ConvertCiv3Media

C7/Map/TileSetLoader.cs
WildWeazel, line 1: Also seems like it belongs in ConvertCiv3Media

C7/Map/MapView.cs
```diff
@@ -0,0 +1,445 @@
...
+		public int worldEdgeRight {get; private set;}
+		public int worldEdgeLeft {get; private set;}
+		private int width;
+		private int height;
+		public bool showGrid {
+			get => showGrid;
+			set {
+				showGrid = value;
+			}
+		}
```
WildWeazel: this is redundant, can just be `public bool showGrid {get; set;}`

C7/Map/MapView.cs
```diff
@@ -0,0 +1,607 @@
...
+				string right = terrain[(x + 1) % width, y];
+				bool even = y % 2 == 0;
+				string top = "coast";
+				if (y > 0) {
+					top = even ? terrain[x, y - 1] : terrain[(x + 1) % width, y - 1];
+				}
+				string bottom = "coast";
+				if (y < height - 1) {
+					bottom = even ? terrain[x, y + 1] : terrain[(x + 1) % width, y + 1];
+				}
```
WildWeazel, line 240: Is it valid for these conditions to be false and the strings left as "coast"? What does that imply?

C7/Map/MapView.cs
```diff
@@ -0,0 +1,445 @@
...
+				setTerrainTile(cell, pcx.atlas, texCoords);
+			}
+			for (int x = 0; x < width; x++) {
+				for (int y = 0; y < height; y++) {
+					lookupAndSetTerrainTile(x, y, x, y);
+				}
+			}
+			for (int y = 0; y < height; y++) {
+				lookupAndSetTerrainTile(width - 1, y, -1, y);
+			}
```
WildWeazel, line 257: only because I've seen several of these "do for all values in range" blocks, there are more concise ways to do this with LINQ using SelectMany, but I'm not sure it provides any real advantage

WildWeazel (changes requested):
Looking great! Just some minor cleanup

C7/OldMapView.cs
pcen: nope, totally transitory!

C7/Map/Civ3TerrainTileset.cs
pcen: It is specific to civ3, but it can support loading from pcx or png where the tiles are laid out in the same layout. unless we're going to create a new layout we could just genericize this implementation slightly

C7/Map/TileSetLoader.cs
pcen: same as above, unless we change the layout of tilesets, the implementation of this class is 90% generic and 10% hardcoded for civ3 at the moment - instead of hardcoding paths to civ3 files we can configure this somewhere, but I don't think the tileset loading logic belongs in convert civ3 media

C7/Map/MapView.cs
pcen: TL;DR, explanation below: these conditions will always be false for tiles on non-wrapping edges. "coast" has no implication here since it's never visible, and is the cleanest default since it always gives _a_ visually correct terrain tile. 

civ terrain tiles represent the corners of 4 game tiles - so at the edges of the map, there is 1/4 of a terrain tile that is "off map". When choosing which terrain tile to use for the intersection of the 3 edge tiles, we need to choose a valid terrain tile combination (terrain tiles can only represent the intersection of 3 different terrain types, never 4). One of the following 2 predicates is always true of the terrain types of the corner of 3 tiles at an edge of the map:
A - one of those three tiles is coast => choosing coast for the 4th "fake, off map" tile is OK
B - two of those tiles have the same terrain type (and it is not coast), and the combination of terrain types will allow for the 4th "fake, off map" tile to be coast

If this were not a tilemap (ie. how the 3.x branch does it), we could instead sample from the corners of 4 different terrain tiles to draw the terrain underneath a game tile. Since we need to pick instead a terrain tile to draw under the joining corner of tiles, coast is a legal option which will give us a valid terrain tile to draw, and the fact that the 4th fake tile corner is hard coded to be coast is inconsequential since it's outside of the playable / visible map.

C7/Map/MapView.cs
pcen: Not familiar with SelectMany so I read the documentation:
Projects each element of a sequence to an [IEnumerable<T>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1?view=net-7.0), flattens the resulting sequences into one sequence, and invokes a result selector function on each element therein.

For me at least, having to read and understand this is much more difficult than reading the more verbose for loops. Perhaps the linq query can generate faster code so maybe there's a tradeoff there.

C7/OldMapView.cs
WildWeazel: Glad I asked!

C7/Map/Civ3TerrainTileset.cs
pcen: We should probably have a settings struct that's easy to initialise with default values instead of injecting a GameData into everything. This RNG is purely cosmetic at the moment as it just determines which terrain texture variant to use

C7/Game.cs
```diff
@@ -289,42 +296,23 @@ enum GameState {
...
-	}
-	public void _on_LeftButton_pressed() {
-		mapView.cameraLocation += new Vector2(-128, 0);
-	}
-	public void _on_UpButton_pressed() {
-		mapView.cameraLocation += new Vector2(0, -64);
-	}
-	public void _on_DownButton_pressed() {
-		mapView.cameraLocation += new Vector2(0, 64);
+	private void onSliderZoomChanged(float value) {
```
benskywalker, line 316: On my end it looks like zoom isn't working right now. If we either change the function here to _on_Zoom_value_changed, or if we change line 271 of C7Game.tscn to match this function, it starts working again. Might be a worthwhile fix. I'm not sure what naming convention we want to follow here.

pcen: thanks, not sure when this broke. It should be named in camel case, not sure why a lot of these callbacks are _ prefixed, maybe when you create a signal in Godot this is/was the default naming convention is uses to generate the function

WildWeazel:
Are you making any new commits directly to this branch or just merging? I think I've reviewed everything up to the first merge.

pcen:
> Are you making any new commits directly to this branch or just merging? I think I've reviewed everything up to the first merge.

Just merging at this point, however the only thing missing from this branch to achieve visual parity with the current rendering is unit hit point indicators.

QuintillusCFC:
Checking out this branch and `godot4`, there are a couple other items beyond hitpoints that are different vis-a-vis visual partity with `godot4`.  One is rivers, which is going to be covered by https://github.com/C7-Game/Prototype/pull/433.  The other I noticed is the city display.

Godot 3:

image

Godot 4:

image

(Note the darker shading around the city, but that isn't specific to tilemaps.  Also the width is too long, and the '3' isn't centered.  But these aren't tilemap-specific, just noting them as something to address before 3.5.x is deprecated)

Tilemaps:

image

For some reason the "Chariot: 10" now escapes its container.

The changes are getting close... tilemaps and Godot 4 are both a lot of work.

WildWeazel:
This was not merged into godot4 for Carthage because it seems a bit more experimental, there are some unresolved questions and regressions called out above, and it had produced some conflicts with the last-minute fixes in godot4. This along with the savegame branch should be prioritized next before they get any further out of date.

PR #408, opened 2023-06-26
Render Units on TileMap

pcen:
rebased

### DIFF
--- a/C7/Map/MapView.cs
+++ b/C7/Map/MapView.cs
@@ -3,6 +3,8 @@ using C7GameData;
 using System;
 using Serilog;
 using System.Linq;
+using System.Collections.Generic;
+using C7Engine;
 
 namespace C7.Map {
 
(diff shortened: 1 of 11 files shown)


QuintillusCFC:
Units seem to be working here, although some of the other things that have been fixed in the base `godot4` branch, such as textures on the buttons, aren't, and for some reason the lower-right status area is in the middle of the screen on this branch.

image

Also note sure if it's just this branch, but the AI isn't building cities in it for some reason.  I know that was working when I last looked at Godot 3.5.

Most of this is probably resolved by the refactor you mentioned.

pcen:
~~This branch is behind the godot4 base so once I rebase all the visual fixes will be in.~~ ~~Rendering cities on tilemap is based on this branch, so is it possible that the AI is building cities but they aren't being rendered?~~ I just checked the log file, cities are being built but they are not rendered on this branch. If you test on https://github.com/C7-Game/Prototype/pull/412 AI should be building cities.

Rebased on godot4 branch so UI should look normal again

WildWeazel:
UI looks good to me in the latest and unit animations seem to be working but I'm not sure what has changed & is expected. Not seeing HP bars for example.

pcen:
> UI looks good to me in the latest and unit animations seem to be working but I'm not sure what has changed & is expected. Not seeing HP bars for example.

Yeah, HP bars need to be reimplemented on top of the TileMap branch still. Everything else should have parity with the main 4.x branch

C7/Map/MapView.cs
```diff
@@ -34,17 +36,96 @@ public partial class MapView : Node2D {
...
+				}
+				if (unit.HasPriorityAsDefender(bestDefender, currentlySelected)) {
+					bestDefender = unit;
+				}
+				if (game.animTracker.getUnitAppearance(unit).DeservesPlayerAttention()) {
+					interesting = unit;
+				}
+			}
+			// Prefer showing the selected unit, secondly show one doing a relevant animation, otherwise show the top defender
+			return selected is not null ? selected : (interesting is not null ? interesting : bestDefender);
```
WildWeazel: you can condense this with the null coalescing operator:
`return selected ?? interesting ?? bestDefender`

WildWeazel (approved):

PR #412, opened 2023-06-29
Godot 4 Render Cities on TileMap

pcen:


C7/PlayerCamera.cs
```diff
@@ -44,4 +44,13 @@ public partial class PlayerCamera : Camera2D
 		Vector2 target = map.tileToLocal(tile);
 		Position = target;
 	}
+
+	public bool isTileInView(Tile tile, MapView map) {
+		Rect2 visible = getVisibleWorld();
+		Vector2 target = map.tileToLocal(tile);
+		float size = 30;
```
WildWeazel, line 51: does this account for zoom level? is there a constant that can be referenced instead of the literal 30?

C7/Map/MapView.cs
```diff
@@ -50,6 +51,14 @@ public partial class MapView : Node2D {
 			return position + offset;
 		}
 
+		public void addCity(City city, Tile tile) {
+			GD.Print($"adding city at tile ({tile.xCoordinate}, {tile.yCoordinate})");
```
WildWeazel: Serilog instead of GD

C7/Map/CityLabelScene.cs
WildWeazel, line 1: quite a few numeric literals in this file that could be assigned to constants to make it easier to read and adjust

WildWeazel (changes requested):

C7/PlayerCamera.cs
pcen: It does, the 30 is 30 pixels in world space. The Rect2 returned from getVisibleWorld accounts for zoom since it uses the camera's transform from GetCanvasTransform. I can make 30 a constant in the camera class. Really, it should check if any of the orthogonal tile is in view, but a conservative bounding box inside the tile is much easier to calculate and I assume a fast bounds check

C7/Map/CityLabelScene.cs
pcen: agreed, I'd rather do a separate refactor for constants since I just ported the existing city scene with minimal changes to draw it on the tile map (in world space - cities can't actually be drawn in the tilemap)

WildWeazel: A static class for all world drawing constants sounds good

QuintillusCFC:
One concern visually:

image

The extra space between the 1 and the star is present in base Godot 4 (as a regression from Godot 3.5), so it isn't a concern here.  But the vertical spacing issues aren't present in base Godot 4.x.  

Base Godot 4:

image

Godot 3:

image

pcen:
> One concern visually:
> 
> image
> 
> The extra space between the 1 and the star is present in base Godot 4 (as a regression from Godot 3.5), so it isn't a concern here. But the vertical spacing issues aren't present in base Godot 4.x.
> 
> Base Godot 4:
> 
> image
> 
> Godot 3:
> 
> image

I just ran the head of this branch on my laptop and I can't reproduce this:
Screenshot 2023-07-31 at 4 15 57 PM

Maybe using pixel offsets isn't the right way to format text

WildWeazel:
and mine seems to be somewhere in between:
image
could this be a function of screen resolution, or system UI scaling?

WildWeazel (approved):

PR #413, opened 2023-06-30
rough prototype code only - save by value

pcen:
edit: I'm using this code as reference for current PRs, but this branch shouldn't be branched off of or used directly

this is a proof of concept based on https://github.com/C7-Game/Prototype/discussions/377 for save game format

### how to try this out
- start a new game as usual
- the static save is (temporarily) a GOTM save file converted to C7 json format -> roads everywhere!
- build carthage
- `command + s` to save
- close the game and start it again -> load game -> load C7/test.json

I've done some previous experiments with save game structure previously attempting to serialise GameData directly and use c# references, but this seems like the most promising way to serialise saves.

- save files are very hard to edit without custom IDs
- save files are very hard to edit when the object graph of GameData is serialised automatically - Newtonsoft and Text.Json both cannot consistently determine references
  - with a custom format, there is a single copy of every entity at the top level of the save, and all references are done via EntityIDs serialised as strings
- loading from SAV and editing saves (ie in a yet-to-be-written C7 map editor) is easier since any object in the C7 json only needs to be updated in one place, and SaveGame is responsible for inflating this structure into the GameData object graph used at runtime


QuintillusCFC:
This branch got auto-closed when I deleted the already-merged use-tilemaps-render-cities branch.  Not sure why it didn't just change the base to the use-tilemaps branch?  At any rate, since it says it's experimental and for reference only, I'll leave it closed - it also helps reduce the confusion around which PRs relate to which other ones.

PR #414, opened 2023-07-09
[Godot 4] Fix Popup Bugs

pcen:
Incrementally removing bugs from the godot4 branch:

- prevents polling for input actions in Game.cs when a popup is visible
- fixes behaviour when the build city popup is open and esc. is pressed: before it would close the popup and show the exit game popup at the same time. now, the first esc. press will only close the current popup

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -27,7 +27,6 @@ public partial class Game : Node2D {
 	public AnimationManager civ3AnimData;
 	public AnimationTracker animTracker;
 
-	Hashtable Terrmask = new Hashtable();
 	GameState CurrentState = GameState.PreGame;
 
 	// CurrentlySelectedUnit is a reference directly into the game state so be careful of race conditions. TODO: Consider storing a GUID instead.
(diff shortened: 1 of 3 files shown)


C7/Game.cs
```diff
@@ -440,6 +442,15 @@ enum GameState {
 
 	// Handle Godot keybind actions
 	private void processActions() {
+		if (Input.IsActionJustPressed(C7Action.Escape) && popupOverlay.Visible) {
+			popupOverlay.OnHidePopup();
+			return;
+		}
+
+		// never poll for actions if UI elements are visible
+		if (popupOverlay.Visible) {
```
WildWeazel: what's the distinction between Visible and ShowingPopup()?

pcen: hmm I wrote ShowingPopup to use here but I must have forgotten to change it... currently there is no difference aside from PopupOverlay.ShowingPopup signalling clear intent vs. the overlay itself being visible which is less clear imo. Will fix

WildWeazel: bump

pcen:
discovered https://github.com/C7-Game/Prototype/issues/424 while cleaning up this change

Sean-Brown:
Getting back into this and not very familiar with all the SDKs involved, but the code changes still make sense to me.

QuintillusCFC (approved):

QuintillusCFC:
Reviewed code, the changes make sense, and should prevent the multi-popup issue generally.  Approved and merging.

PR #415, opened 2023-07-09
[Godot 4] Make ParameterWrapper type safe and remove memory leak

pcen:
- makes ParameterWrapper class take the type parameter instead of the value getter, making it impossible to attempt to cast `object value` to the wrong type
- makes ParameterWrapper inherit from `RefCounted` godot class so it does not leak memory


WildWeazel (approved):
good find!

WildWeazel:
merge if ready

PR #418, opened 2023-07-31
[4.x] Default font size 14

QuintillusCFC:
Sets the default font size to 14, and sets it to use Noto Sans.  14 was the default in Godot 3.x; this causes it to match existing functionality.  Downside, for some reason the overrides to font size for things like headers in dialogs aren't working; I verified that they weren't without these changes either.

See Popup.js, AddHeader method as an example of where we have an override that worked in 3.5 but doesn't in 4.x yet.

Also, switch to Godot 4.1.  We may not want to merge this until pcen has also signed-off on 4.1.

### DIFF
--- a/C7/C7.csproj
+++ b/C7/C7.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Godot.NET.Sdk/4.0.3">
+<Project Sdk="Godot.NET.Sdk/4.1.1">
   <PropertyGroup>
     <TargetFramework>net7.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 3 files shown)


pcen:
tested out 4.1 and everything looks good to me

WildWeazel:
Is 13 for buttons intentional or a typo?

WildWeazel:
bump
@pcen should this go directly to godot4 or into your staging branch?

pcen:
> bump
> @pcen should this go directly to godot4 or into your staging branch?

I've been hesitant to merge anything into godot4 because I really want to avoid merge conflicts with tilemaps branch, but this should be fine to merge into godot4 directly

pcen (approved):

PR #419, opened 2023-07-31
[4.x] TileMap: Grid View

pcen:
Fixes TileMap from crashing on ctrl+g
Implements grid overlay on the tilemap


WildWeazel (approved):
Looks good and works for me! Just thinking for later, this may be another good use of a shader instead of drawing another image, if there's any advantage.

pcen:
> Looks good and works for me! Just thinking for later, this may be another good use of a shader instead of drawing another image, if there's any advantage.

I _think_ a custom shader for the tilemap would render slightly faster, and make toggling the grid faster

Issue #103, opened 2022-02-14
Fog of war - consider tile elevation and obstacles


QuintillusCFC:
Certain type of terrain affect fog of war.

Hills and mountains and volcanoes let you see farther, but also prevent you from seeing past them.

This issue is to consider and implement that.

We should probably do this by adding a Height property to TerrainType, and setting it appropriately when importing from Civ3.

TBD whether Babylon or Carthage.

QuintillusCFC:
Some notes:

**Mountain rules**
Mountains see two unless blocked
Other mountains block mountains
Hills do _not_ block mountain visibility
Forest/jungle do _not_ block mountain visibility

![image](https://user-images.githubusercontent.com/5287977/191105873-b9e178c9-908d-4fd0-b2d6-423addcdd33b.png)
_Abe Lincoln can see two tiles across the jungle from the Mountain_

**Hills rules**

Hills see two unless blocked
Hills block hills
Forest/jungle block hills
Mountains block hills

![image](https://user-images.githubusercontent.com/5287977/191105371-55d7cc6c-d534-42be-a264-c603e39c137f.png)
_The Scout can see distance 2, with nothing blocking the view_

**Flatland rules**

Hills and mountains visible at distance 2 over Marsh
Mountains visible at distance 2 over Jungle and Forest
Hills _not_ visible at distance 2 over Jungle
Cannot see forests/jungle at distance 2
Mountains visible at distance 2 over Hills

![image](https://user-images.githubusercontent.com/5287977/191104973-92905bf4-a7a3-4113-a8a7-5339d0756c8d.png)
_The settler can see the Hill at distance 2 across the Marsh_

![image](https://user-images.githubusercontent.com/5287977/191106547-07446596-7ce7-4ade-b41e-5e96f5a1a13e.png)
_Theodora cannot see the Hill across the Jungle_

![image](https://user-images.githubusercontent.com/5287977/191106618-93a7e9c6-9c64-4bec-ba50-b053eeb9f64a.png)
_Theodora can see the Mountain across the Jungle_

![image](https://user-images.githubusercontent.com/5287977/191107466-d0b36268-0ab8-412e-8450-9e62bcd056dd.png)
_Theodora cannot see the forest that is SE of the game_

![image](https://user-images.githubusercontent.com/5287977/191108572-10540c1a-d36c-4cbc-b47b-2e2052e4466c.png)
_The Worker can see the Mountain despite the Hill in front of it_

**Coast rules**

[...]

Issue #102, opened 2022-02-14
Fog of war when know tiles are no longer visible [Part 1]


QuintillusCFC:
Builds on #101 .  If you no longer have a unit near a tile, it should have fog of war.

For part 1 of this, I suggest counting only tiles next to/within the 3x3 city boundaries, or adjacent to a unit, as visible.  We can add improvements based on terrain type and expanding borders later; this is likely to be a sizeable enough improvement with the smaller scope.

TBD whether Babylon or Carthage.

JimOfLeisure:
As alluded to in my comment in that task, tile visibility is stored in the Civ3 SAV file as a 32-bit bitmap. I called it "isVisible", and Lucien is calling it "ExploredBy". The following bitmap I called "isVisibleNow" and Lucien named "VisibleTo" and I believe is the state you mention here.

<https://github.com/C7-Game/Prototype/blob/c78b9804df659b774d310f53b3bced15753a4a2d/QueryCiv3/SavSections/Tile.cs#L47>

I'm not sure we need to store "isVisibleNow" in a C7 native sav file, since it's recalculated every unit move and border change, but I guess there should definitely be a getter or something to indicate if the tile should be fully lit instead of dimly lit as indicated by "ExploredBy" (or whatever its new implementation is, which is probably a getter, too).

Incidentally, there is some additional functionality as indicated by the save file. Each tile in the native Civ3 SAV has a per-civ bitmap of tile improvements in addition to the real current state. So an explored tile that is no longer directly visible has its improvements changed, the viewing civ in question only sees their last active view of the improvements.

Interestingly this does not apply to forests and marshes as there is no memory for them, so you always see the current base and overlay terrain state. I *think* resources are live-view only if they disappear or reappear, but I don't think I've ever tried to test that out to be sure.

esbudylin:
Implemented in #620

PR #421, opened 2023-08-01
[4.x] TileMap: Fog of War Layer

pcen:
Adds fog of war on par with the 3.5 implementation. Fog of war PCX files are loaded differently, since the alpha seems to be the inverse of the colour intensity, which gives a more realistic fog effect:
Screenshot 2023-08-01 at 4 17 56 PM


C7/Map/MapView.cs
```diff
@@ -517,5 +575,17 @@ public partial class MapView : Node2D {
 
 			updateBuildingLayer(tile);
 		}
+
+		public void discoverTile(Tile tile, TileKnowledge tk) {
+			HashSet<Tile> update = new HashSet<Tile>(tile.neighbors.Values);
```
WildWeazel, line 596: I'm not sure I follow what's happening here but I think it could be significantly shortened with LINQ and without the allocation. Is it building a set of unique tiles within a radius of 2 (and if so why 2)?

WildWeazel (commented):
It works in game but since I was too lazy to test.. does it still black out the unknown tiles if the FOW image is blank?

pcen:
> It works in game but since I was too lazy to test.. does it still black out the unknown tiles if the FOW image is blank?

If the fog of war is blank (as in fully transparent or missing from the tilemap atlas), only discovered tiles, and tiles adjacent to discovered tiles will be rendered. The adjacent ones need to be rendered since the fog of war reveals the edge of these tiles

C7/Map/MapView.cs
pcen: currently (in the 3.6 branch) we reveal all adjacent tiles to the tile a unit enters, hence all immediate neighbours need to be updated. If any of those neighbouring tiles wasn't already discovered, then the fog of war of their neighbours will change (generally if all the surrounding tiles are previously unexplored, the fog of war over these tiles will change to reveal the edge of their terrain) hence the radius of 2. In the future (https://github.com/C7-Game/Prototype/issues/103 and https://github.com/C7-Game/Prototype/issues/102), the calculation for which tiles to update the fog of war layer for will be much more sophisticated, but for now I just went for parity with the 3.6 branch

WildWeazel: Good point that the algorithm will need to change. In that case I won't try to optimize this now.

WildWeazel:
Right, I forgot that the bordering unexplored tiles are technically shown underneath the FOW. I think that's why Civ3 ultimately rendered it dynamically instead of using this overlay image that's easily disabled. We might consider the same, but that's kind of a gameplay decision.

WildWeazel (approved):

pcen:
> Right, I forgot that the bordering unexplored tiles are technically shown underneath the FOW. I think that's why Civ3 ultimately rendered it dynamically instead of using this overlay image that's easily disabled. We might consider the same, but that's kind of a gameplay decision.

The TileMap can also do something more sophisticated when rendering tiles: if the tile is only partially revealed by fog of war, we can only render the terrain below it so no additional information is leaked. Ultimately though, in a client/server architecture the server would have to only make the client aware of some features of the tile when it's partially revealed

pcen:
will merge this after https://github.com/C7-Game/Prototype/pull/412 is merged into the base tilemap branch, as this can then be merged into the base branch as well

WildWeazel:
> will merge this after #412 is merged into the base tilemap branch, as this can then be merged into the base branch as well

Is #412 still waiting on anything? It looks like my comments were resolved but it's still marked "Changes requested"

pcen:
> > will merge this after #412 is merged into the base tilemap branch, as this can then be merged into the base branch as well
> 
> Is #412 still waiting on anything? It looks like my comments were resolved but it's still marked "Changes requested"

Not waiting on anything else. I think it remains changes requested until it gets another review

PR #422, opened 2023-08-02
[4.x] TileMap: Horizontal Wrapping

pcen:


C7/Map/MapView.cs
```diff
@@ -61,10 +70,18 @@ public partial class MapView : Node2D {
 			CityScene scene = new CityScene(city, tile);
 			scene.Position = tilemap.MapToLocal(stackedCoords(tile));
 			AddChild(scene);
-			cityScenes.Add(city, scene);
+			cityScenes.Add(tile, scene);
+		}
+
+		private Vector2I horizontalWrapOffset(HorizontalWrapState wrap) {
+			return wrap switch {
```
WildWeazel, line 85: I didn't know about this syntax. Pretty cool!

WildWeazel:
Seems to work well after several circumnavigations. I'm still wrapping my head around how this works, pun intended, but is there any potential issue with scrolling indefinitely?

pcen:
> Seems to work well after several circumnavigations. I'm still wrapping my head around how this works, pun intended, but is there any potential issue with scrolling indefinitely?

The camera never goes more than one full map-width beyond the "main" map. Once it reaches that point, the camera's location is translated one map-width back towards the origin, so in terms of infinite scrolling it will just keep wrapping back towards the origin

C7/Map/MapViewCamera.cs
```diff
@@ -0,0 +1,116 @@
...
+
+		public void centerOnTile(Tile tile, MapView map) {
+			Vector2 target = map.tileToLocal(tile);
+			setPosition(target);
+		}
+
+		public bool isTileInView(Tile tile, MapView map) {
+			Rect2 visible = getVisibleWorld();
+			Vector2 target = map.tileToLocal(tile);
+			float size = 30;
```
WildWeazel, line 110: What's this number? Can it be derived from something, and it not made a constant somewhere? I think we'd discussed having a class to hold all map related constants.

C7/Map/MapViewCamera.cs
```diff
@@ -0,0 +1,116 @@
...
+		}
+		private float wrapEdgeMargin = 0;
+		float wrapRightEdge {get => map.worldEdgeRight - wrapEdgeMargin; }
+		float wrapLeftEdge {get => map.worldEdgeLeft + wrapEdgeMargin; }
+		private bool enteringRightWrap(Rect2 v) => hwrap != HorizontalWrapState.Right && v.End.X >= wrapRightEdge;
+		private bool enteringLeftWrap(Rect2 v) => hwrap != HorizontalWrapState.Left && v.Position.X <= wrapLeftEdge;
+		private bool atEdgeOfRightWrap(Rect2 v) => hwrap == HorizontalWrapState.Right && v.End.X >= wrapRightEdge + map.pixelWidth;
+		private bool atEdgeOfLeftWrap(Rect2 v) => hwrap == HorizontalWrapState.Left && v.Position.X <= wrapLeftEdge - map.pixelWidth;
+		private HorizontalWrapState currentHwrap(Rect2 v) {
+			return v.Position.X <= wrapLeftEdge ? HorizontalWrapState.Left : (v.End.X >= wrapRightEdge ? HorizontalWrapState.Right : HorizontalWrapState.None);
```
WildWeazel, line 58: Would this be better expressed as one of those switch statements or is that possible here? Also is it correct to use v.Position in one case and v.End in the other?

C7/Map/MapViewCamera.cs
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		// checkWorldWrap determines if the camera is about to spill over the world map and will:
+		// - move the second "wrap" tilemap to the appropriate edge
+		//   to give the illusion of true wrapping tilemap
+		// - teleport the camera one world-width to the left or right when
+		//   only the "wrap" tilemap is in view
+
+		private void checkWorldWrap() {
```
WildWeazel, line 67: How's this going to work when someone makes a tiny map and zooms all the way out? ie if both left and right edges are visible at the same time

WildWeazel: Also, just want to go ahead and point out that we should be able to handle horizontal and/or vertical wrap.

C7/Map/MapViewCamera.cs
```diff
@@ -0,0 +1,116 @@
...
+			if (enteringRightWrap(visibleWorld)) {
+				map.setHorizontalWrap(HorizontalWrapState.Right);
+			} else if (enteringLeftWrap(visibleWorld)) {
+				map.setHorizontalWrap(HorizontalWrapState.Left);
+			}
+			if (atEdgeOfRightWrap(visibleWorld)) {
+				Translate(Vector2.Left * map.pixelWidth);
+			} else if (atEdgeOfLeftWrap(visibleWorld)) {
+				Translate(Vector2.Right * map.pixelWidth);
+			}
```
WildWeazel, line 82: not necessarily needed right now, but Civ has a gradient where the scroll speed increases as the cursor approaches and pushes into the map edge

C7/Map/MapViewCamera.cs
pcen: not really possible, each predicate is too verbose and it makes the actual logic hard to follow 
position is start https://docs.godotengine.org/en/stable/classes/class_rect2.html#class-rect2-property-end

C7/Map/MapViewCamera.cs
pcen: I'm not sure, but I was thinking instead of something super fancy just limit the minimum possible zoom to never be able to span the entirety of a map. The benefits of a tilemap is that in the average case of normal sized maps we get a huge performance boost. Maybe we can just not load maps that are too small until this edge case is handled more elegantly than limiting zoom

WildWeazel:
If nothing else let's merge this :)

PR #423, opened 2023-08-07
Incrementing IDs

pcen:
Object IDs are discussed here: https://github.com/C7-Game/Prototype/discussions/210
Also here for save format: https://github.com/C7-Game/Prototype/discussions/377

This change is a prerequisite for implementing something like my save format draft https://github.com/C7-Game/Prototype/pull/413 and I intend this to be an incremental change towards an actual PR for getting working save game. Having said that, regardless of how save game is implemented in the future, this PR refactors and removes string-based GUID for units and cities, and an ad-hoc ID system that was being used for tiles. The ID class consists of a string key and a unique integer value. I did not implement more complex custom ID's that would allow for things like <key>-<city-name> for cities, since this would require more complex ID parsing from saves, but can always be added in the future as a refactor of the ID class (or probably making it an interface...).

This change should not affect gameplay.


Sean-Brown (approved):
After reading the discussions, the code changes make sense

QuintillusCFC (approved):
Makes sense to me as well.  I'm going to see if this can be merged into the `godot4` branch, keeping it from depending upon the tilemaps.

QuintillusCFC:
Nope, GitHub is not smart enough to figure out that I'd like to essentially rebase or cherry-pick to a different base.  I will see if I can do that locally and thus allow decoupling of saves from tilemaps to proceed.

QuintillusCFC:
Okay, well, I accidentally cherry-picked right to `godot4` rather than a branch to open a PR to merge into `godot4`, but at any rate this is now merged into `godot4`.  I'll merge it into the `tilemaps` as originally intended as well, so it's clear that this PR is all merged.

PR #426, opened 2023-09-20
Code Path to Save Current Game and Load Arbitrary Saves

pcen:
to save the current game: ctrl + s -> saves to hardcoded path `./Text/save.json` (if running Godot from the command line, invoke in C7 subfolder)

can now load .sav as well as .json files from main menu


C7/Art/grid.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
pcen, line 1: apparently we should check all `*.import` files into source, but ***not*** Godot's `.import` folder

WildWeazel:
@pcen This one slipped through the cracks- what's the status of this branch? Did these changes make it into godot4 when you split out your save work? If not there are likely conflicts by now, and maybe some redundancy.

pcen:
It's not merged. I think it was mostly UI change to allow loading and saving different saves, I don't think it added too much functionality. We should re-base if necessary and merge into Godot 4 branch

WildWeazel:
This did not make it into the Carthage merge (though technically that ship hasn't sailed yet...) as there are significant conflicts between it and godot4 by this point, and I don't know enough about it to sort it out. This along with the tilemaps branch should be prioritized next before they get any further behind.

esbudylin:
The save/load functionality has been on the main branch for quite some time, so this PR is likely no longer relevant.

PR #427, opened 2023-09-21
godot4 fix: remove autoloaded nodes

pcen:
The reason for this change is that we are seeing issues opening the C7 project on the godot4 branch in the editor on linux:
image
Which may potentially be the same as [this issue](https://github.com/godotengine/godot/issues/36787), which is reproducible in Godot 4.1 by creating a new project and adding an autoloaded C# node. This fix removes all autoloaded nodes:
GlobalSingleton becomes a singleton managed by C# using System.Lazy
LogManager becomes a purely static class with a static constructor running before methods are accessed, and currently LogManager.ShutDown() will need to be called whenever the application is closing in order to flush the log (ie. for every scene, have the top level node override _ExitTree and invoke LogManager.ShutDown, *if* exiting the tree implies the entire application is closing. Obviously we need a better solution for flushing the log in the future)

### DIFF
--- a/C7/Credits.tscn
+++ b/C7/Credits.tscn
@@ -1,6 +1,6 @@
-[gd_scene load_steps=2 format=2]
+[gd_scene load_steps=2 format=3 uid="uid://ouangx1h81vh"]
 
-[ext_resource path="res://Credits.cs" type="Script" id=1]
+[ext_resource type="Script" path="res://Credits.cs" id="1"]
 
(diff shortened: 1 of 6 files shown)


benskywalker:
I am no longer experiencing this issue on Linux, and it seems like a phantom issue caused by incompatibility with Vulkan or having dotnet 6 instead of 7.  Seems like this might be fine to close?

pcen:
closing for now until this issue comes up again

PR #428, opened 2023-09-22
Fix 350 truncate movement

benskywalker:
I'm going ahead and opening this PR because things look good on my end. I hope I've done this right, and I'm concerned by how big a diff this is showing. I only changed three files: MapUnit, MovementPoints, and LowerRightInfoBox

I just pulled from godot4 for my branch and until I made these changes it said they were up to date (up to date with Prototype / godot4, not just my fork. I could use some advice on this one, let me know if I've done something wrong here. 

Solving this issue: https://github.com/C7-Game/Prototype/issues/350

Issue #350, opened 2022-09-19
Truncate Movement Point Display


QuintillusCFC:
Right now a unit that has moved on a road displays its movement like this:

![image](https://user-images.githubusercontent.com/5287977/191117005-344ce313-9595-4dc6-af87-fde0f3066e80.png)

I actually kind of like that it shows the fractional part, versus in Civ3 where it would show 1/1 even though it had already used 1/3 of its movement points.  But it's a little silly that it shows that many digits.

I propose we default to two digits, truncated.  So with default road speeds, it would be 1/1, 0.66/1, 0.33/1, and 0.  With 4-movement roads, it would be 1, 0.75, 0.5, 0.25.  It reminds the player of how much movement they have left if they jump back and forth between units a bit, but it's as overkill as 0.6666666.

benskywalker:
I'll take this one. I'm just working on finding out which branch I need to base it off of, and then I'll fork this repo and have a PR. I've already got the changes working locally. For consideration: would it be better to have   (2/3)/1  instead of 0.67/1 ?

WildWeazel:
I was just thinking fractions would be great if you can get it to work with any numbers. There's probably a function somewhere in .NET to produce that.

benskywalker:
> I was just thinking fractions would be great if you can get it to work with any numbers. There's probably a function somewhere in .NET to produce that.

Hmm, ok so I've given this some more thought. Irregular fractions might pose an issue. For example, (15/4)/4 is probably less clear than 3.75/4

We could fix that by making it a mixed number like (3+3/4)/4 or (3 3/4)/4
That would induce some ui clutter though, and I'm not sure if this is worth it. The fractions would probably be pretty useful if the movements costs on roads / railroads are more irregular (eg: roads cost 1/6 and railroads cost 1/12). In that case it would pay dividends, and I'm happy to go forward with that route if that's what we prefer

QuintillusCFC:
The other UI consideration is that some units will have defensive bombard, e.g. an archer would display like:

2.1(1) 0.5/1

If Civ3 conventions were followed and it had traveled one tile on a road that cost half a movement point.

So if we had fractions, it would be:

2.1(1) (1/2)/1

Or for a catapult-type unit:

0.0(4) (3/4)/1
or
0.0(4) 0.75/1

If roads cost 1/4 of a movement point.

[...]

PR #429, opened 2023-09-22
Fixing Movement UI With Mixed Numbers

benskywalker:
Fixing issue #350 . This is another PR since I closed #428 due to a merge mistake on my end. This pr adds mixed numbers for remaining movement points. Feedback on whether this change is worthwhile is requested. @QuintillusCFC 
image

### DIFF
--- a/C7GameData/MovementPoints.cs
+++ b/C7GameData/MovementPoints.cs
@@ -38,55 +38,63 @@ namespace C7GameData
 		{
 			// Modified from Stack overflow https://stackoverflow.com/a/32903747
 			// Only supports positive values
-			double absolute_accuracy=0.01;
+			// Might need to mess with accuracy; I want to avoid a movement of 1 resulting in 99/100
+			double absoluteAccuracy=0.005;
 			double val = remaining;
 			if(val < 0)
 			{
 				return "0";
 			}
 
-			int whole_num = (int) Math.Floor(val);
-			val -= whole_num;
+			int wholeNum = (int) Math.Round(val,0);
+			
 
-
-			if(val <= absolute_accuracy)
+			// This is important for important cases like 0.999. We have to round (possibly up) and then check if the whole number is close enough. Otherwise, we'll floor and create a mixed number 
+			if(Math.Abs(wholeNum - val) <= absoluteAccuracy)
 			{
-				return whole_num.ToString();
+				return wholeNum.ToString();
+			}
+			
+			wholeNum = (int) Math.Floor(val);
+			val -= wholeNum;
+			if(val <= absoluteAccuracy)
+			{
+				return wholeNum.ToString();
 			}
 
 			// The lower fraction is 0/1
-			int lower_n = 0;
-			int lower_d = 1;
+			int lowerN = 0;
+			int lowerD = 1;
 
 			// The upper fraction is 1/1
 			// Start with an upper guess of 1 and a lower guess of 0, then converge through a binary search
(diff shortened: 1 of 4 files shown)


C7GameData/MovementPoints.cs
```diff
@@ -33,5 +33,100 @@ public struct MovementPoints {
...
+					}
+					return "(" + whole_num.ToString() + " " + middle_n.ToString() + "/" + middle_d.ToString() + ")";
+				}
+			}
+
+		}
+
+		/// Binary seek for the value where f() becomes false.
+		/// Used for mixed number calculation. Taken from https://stackoverflow.com/a/32903747
+		void Seek(ref int a, ref int b, int ainc, int binc, Func<int, int, bool> f)
```
pcen: nit: might be nice to modify this to return (int, int) instead of mutating reference parameters - that is `(int, int) Seek(int a, int b, int ainc, int binc, Func<int, int, bool> f)`

pcen:
Looks good, thanks for working on this and congrats on your first PR in C7!!

nit on style: we use camelCase for variables so for consistency we should update the SO code too.

Also, are there other UI elements that will require nicely displaying floats in different formats? if so we could put the number formatting in a static class designed for formatting numbers, but I'm not sure if any other cases exist

benskywalker:
> Looks good, thanks for working on this and congrats on your first PR in C7!!
> 
> nit on style: we use camelCase for variables so for consistency we should update the SO code too.
> 
> Also, are there other UI elements that will require nicely displaying floats in different formats? if so we could put the number formatting in a static class designed for formatting numbers, but I'm not sure if any other cases exist

Thank you! Good points, I've refactored for camelCase. I can't think of any other uses for this, but I'm happy to move it if we can find one. It's kind of niche because it only applies when we need to display a decimal to users, AND when that decimal is intrinsically created by fractions. Maybe if we ever want to show Effective Combat Strength as a preview?

I've also fixed a bug where 0.999 would sometimes show as 99/100 or 999/1000. Now it should reliably show 1.  I've also modified Seek as you suggested. 
@pcen , if you have the bandwidth it might be good if you could check out the branch to review it. My dev environment is still not completely stable, so a doublecheck might be nice

pcen:
Something interesting I noticed is that after I cloned your repo, but before I opened the project in the Godot editor, just running it from the command line gave the following errors:
Screenshot 2023-09-23 at 2 25 30 AM
And when running, the lower right-hand corner UI box was empty for me:
Screenshot 2023-09-23 at 2 24 22 AM
After opening the project in the editor, then running it from either the command line or in the editor looks normal and doesn't give any errors. I'm curious as to what difference it makes opening the project in the editor (the git diff doesn't change, so whatever the diff is is gitignored...) But running it after I open it in the editor looks great!

pcen (approved):

benskywalker:
That's odd, I've never tried running it from the command line before! I'll try that going forward

benskywalker:
It looks like we're good to merge unless anyone has other concerns

pcen:
@WildWeazel can we get another stamp before merging?

WildWeazel:
I ran from the command line but had checked out this PR from my workspace, and did not see those errors. Seems like some kind of clean build issue.
I had thought about where else this might apply, and could only come up with the right-click menu which is already covered. If it turns up elsewhere, it's easy enough to extract.
Anyway, congrats on your first merge! That means I have a new task to do... update the credits!

Issue #94, opened 2022-02-14
Settler AI - Consider tiles in the BFC, and potential yield improvements


QuintillusCFC:
This is a more advanced one, after #92 and #91 .  Currently, the AI only considers tiles immediately next to cities.  This is pretty much okay early on since those will be the tiles that the city works.  However, there should be _some_ consideration of the tiles in the BFC (big fat cross) as well, and the AI should also consider whether tile improvements can increase the yield - notably, Hills and Mountains get +2 shields with a mine, versus +1 for other tiles, which is worth something, whereas tundra cannot be irrigated, and volcanoes cannot be mined; volcanoes should be considered less valuable than mountains despite the same initial tile yield.

The improvement weights should likely be considerably less than the base yields, to avoid the AI hamstringing itself by choosing tiles with low yields early on.  Desert can be improved just as much as grassland, but is a much inferior choice early on, for example.  The AI should also likely consider an either/or for food/shield improvement, since you cannot make both.  Intuitively, I'm thinking 1/3 the weight of base yields, although that may be too high or too low.  Make it configurable.

BFC weights should also be considerably less; we don't want the AI choosing locations because there are a bunch of cows right _outside_ the new city's borders.  But if there are two locations the AI is considering and one has Game outside the borders and the other doesn't, it's a tiebreaker.

It may be necessary to bump the base yields to make these lower weights feasible, e.g. the value of a base commerce might be 20 instead of 2.  Alternately, we could use floats so we have decimal weights when dividing.

benskywalker:
I'm taking a look at this, and I think it also makes sense to consider defensive bonus as a factor, rather than hard-coding a bonus for hills (as currently happens [here](https://github.com/C7-Game/Prototype/blob/5b044b091a0d0721359dea846d8d4a63e9959f82/C7Engine/AI/UnitAI/SettlerLocationAI.cs#L63C3-L63C44)).
I can calibrate it so hills will retain a similar advantage, and this will help if the defense values change. Also, is there any reason not to make scores a double? I'll look into it

benskywalker:
I can't see any negative consequences of changing the score to a double, so I'm doing that locally. The only downside is the change propagated into other AI Scores, but maybe those should be doubles anyway for precision? 

[...]

PR #430, opened 2023-09-23
[godot4] [AI] Fix Issue 94 - Improve SettlerLocationAI

benskywalker:
This PR should solve issue #94. I have added the BFC tile logic for AI settlers, and it is working.

I left some log statements in we'll probably want to take out once this is ready to merge.

I also added the structure for calculating the irrigation and mining bonus for terrain types. The only issue is it looks like biq.cs is not pulling that information in yet. That's not really my area - @QuintillusCFC , that data would be in the biq file somewhere, right? Terr.cs does have a [section](https://github.com/C7-Game/Prototype/issues/94) for these bonuses, but it isn't getting input yet. 

I'm just opening the PR so we can see where we're at on this

### DIFF
--- a/C7Engine/AI/UnitAI/SettlerLocationAI.cs
+++ b/C7Engine/AI/UnitAI/SettlerLocationAI.cs
@@ -55,14 +55,16 @@ namespace C7Engine
 				//For simplicity's sake, I'm only going to look at immediate neighbors here, but
 				//a lot more things should be considered over time.
 				foreach (Tile nt in t.neighbors.Values) {
+					double improvementScore = GetTileImprovementScore(nt, player);
 					double yieldScore = GetTileYieldScore(nt, player);
 					log.Information("Neighbor tile has score of " + yieldScore);
+					log.Information("Neighbor tile has improvement score of " + improvementScore);
 					score += yieldScore;
 				}
 				//Also look at the next ring out, with lower weights.
 				foreach (Tile outerTile in t.neighbors.Values)
 				{
-					double outerTileScore = GetTileYieldScore(outerTile, player) / 3;
+					double outerTileScore = (GetTileYieldScore(outerTile, player) + GetTileImprovementScore(outerTile, player)) / 3;
 					score += outerTileScore;
 					log.Information("Outer ring tile has yield score of " + outerTileScore);
 				}
@@ -104,6 +106,19 @@ namespace C7Engine
 				score += LUXURY_RESOURCE_BONUS;
 			}
 			return score;
+		}
+
+		private static double GetTileImprovementScore (Tile t, Player owner)
+		{
+			double irrigationBonus = t.irrigationYield(owner);
+			double mineBonus = t.miningYield();
+
+			// Food is more important than production 
+			double irrigationValue = irrigationBonus * 5;
+			double mineValue = mineBonus * 3;
+
+			// Since we can only irrigate OR mine, we just return the max of the two
+			return Math.Max(irrigationValue,mineValue);
 		}
 
 		public static bool IsInvalidCityLocation(Tile tile) {
--- a/C7Engine/AI/UnitAI/SettlerLocationAI.cs
+++ b/C7Engine/AI/UnitAI/SettlerLocationAI.cs
@@ -113,8 +113,10 @@ namespace C7Engine
 			double irrigationBonus = t.irrigationYield(owner);
 			double mineBonus = t.miningYield();
 
-			// Food is more important than production 
-			double irrigationValue = irrigationBonus * 5;
+			// Food is more important than production
+			// Irrigation only applies to freshwater tiles, although we should add a check for electricity later.
+			// Also this doesn't account for the ability to chain irrigation.
+			double irrigationValue = irrigationBonus * 5 * (t.providesFreshWater() ? 1 : 0);
 			double mineValue = mineBonus * 3;
 
 			// Since we can only irrigate OR mine, we just return the max of the two
--- a/C7Engine/AI/UnitAI/SettlerLocationAI.cs
+++ b/C7Engine/AI/UnitAI/SettlerLocationAI.cs
@@ -19,15 +19,15 @@ namespace C7Engine
 
 		//Figures out where to plant Settlers
 		public static Tile findSettlerLocation(Tile start, Player player) {
-			Dictionary<Tile, int> scores = GetScoredSettlerCandidates(start, player);
+			Dictionary<Tile, double> scores = GetScoredSettlerCandidates(start, player);
 			if (scores.Count == 0 || scores.Values.Max() <= 0) {
 				return Tile.NONE;	//nowhere to settle
 			}
 
-			IOrderedEnumerable<KeyValuePair<Tile, int> > orderedScores = scores.OrderByDescending(t => t.Value);
+			IOrderedEnumerable<KeyValuePair<Tile, double> > orderedScores = scores.OrderByDescending(t => t.Value);
 			log.Debug("Top city location candidates from " + start + ":");
 			Tile returnValue = null;
-			foreach (KeyValuePair<Tile, int> kvp in orderedScores.Take(5))
+			foreach (KeyValuePair<Tile, double> kvp in orderedScores.Take(5))
 			{
 				if (returnValue == null) {
 					returnValue = kvp.Key;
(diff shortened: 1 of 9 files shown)


benskywalker:
Just pushed out some more improvements. We now have a way to calculate whether a city is coastal, which does apply to unit production (lake cities can not build galleys). I've also added the logic for land isthmuses, which is working for the purposes of calculating lakes (small lakes separated from the ocean by a non-tile isthmus do not count as part of the ocean anymore).

pcen:
I was reading the IsLake implementation and it got me thinking how we deal with Tiles throughout the entire code base. I was going to suggest that searchedWater be a HashSet but I realised, as you've implemented it, that there's probably no difference in performance when there's only 21 tiles in the list/set. Since we can determine the index of a Tile given it's x,y coords, we could uniquely identify tiles by their index, and use bitsets in a lot of places. Especially for TileKnowledge, it would save a lot of memory allocations to use bitsets instead of HashSets of tiles

benskywalker:
That's an interesting idea, and it does sound more efficient. I worry it would make the code a bit harder to understand / debug though. Maybe I'm just wary because I haven't used bitsets much, but I wonder if it would be worthwhile considering the many hands that will modify pieces of our code

benskywalker:
I'm sorry, I deleted my previous comment because I was definitely commenting out of my ignorance about bitsets. That sounds like a very good idea for efficiency, and it would save a lot of unnecessary references for simple stuff. Because of the memory saved, it might be feasible to calculate things like IsLake for all tiles at the start of the game, and then it could just be a single bit.

C7Engine/AI/UnitAI/SettlerLocationAI.cs
```diff
@@ -97,6 +108,21 @@ private static int GetTileYieldScore(Tile t, Player owner)
 			return score;
 		}
 
+		private static double GetTileImprovementScore (Tile t, Player owner)
+		{
+			double irrigationBonus = t.irrigationYield(owner);
+			double mineBonus = t.miningYield();
```
QuintillusCFC, line 114: Minor, but the tradeBonus should also be considered.  It's disabled by default for volcanoes, and can be disabled or increased in mods as well.  The commerce yield weight in the tile score is 2 which seems largely appropriate.

C7Engine/AI/UnitAI/SettlerLocationAI.cs
```diff
@@ -39,36 +39,44 @@ public class SettlerLocationAI {
...
+				double score = GetTileYieldScore(t, player);
 				//For simplicity's sake, I'm only going to look at immediate neighbors here, but
 				//a lot more things should be considered over time.
 				foreach (Tile nt in t.neighbors.Values) {
-					score += GetTileYieldScore(nt, player);
+					double improvementScore = GetTileImprovementScore(nt, player);
+					double yieldScore = GetTileYieldScore(nt, player);
+					log.Information("Neighbor tile has score of " + yieldScore);
+					log.Information("Neighbor tile has improvement score of " + improvementScore);
+					score += yieldScore;
```
QuintillusCFC, line 62: Shouldn't the improvementScore be included in this?  I would argue that it should have a lesser weight than the yieldScore, since in most cases the improvements won't exist when the tile is being settled, but greater than zero weight.

C7Engine/AI/UnitAI/SettlerLocationAI.cs
```diff
@@ -39,36 +39,44 @@ public class SettlerLocationAI {
...
+				}
+				//Also look at the next ring out, with lower weights.
+				foreach (Tile outerTile in t.neighbors.Values)
```
QuintillusCFC, line 65: In this block, improvement score is included, but isn't it still looping over the same tiles (t.neighbors.Values in both cases)?

C7Engine/AI/UnitAI/SettlerLocationAI.cs
```diff
@@ -39,36 +39,44 @@ public class SettlerLocationAI {
...
 
 				//Prefer hills for defense, and coast for boats and such.
-				if (t.baseTerrainType.Key == "hills") {
-					score += 10;
-				}
-				if (t.NeighborsWater()) {
-					score += 10;
-				}
+				// In this scale, the defense bonus of hills adds up to a bonus of +10, which is equivalent to the previous hardcoded bonus. This just opens up possibilities with editing terrain.
+				score += t.baseTerrainType.defenseBonus.amount * 20.0;
```
QuintillusCFC, line 74: Definitely like that conceptually, and it will be useful in scenarios which allow settling on volcanoes and mountains, or which set different bonuses for tundra/plains/etc.

One gotcha that should be noted in a comment for now is that forests and jungles provide defensive bonuses, but are removed when a city is built on them - thus the bonus should not factor in here.  We don't yet have a place where we capture that "disappears when city is built on them" attribute as it isn't moddable in Civ3 and is just an intrinsic property of the terrain.

C7Engine/AI/UnitAI/SettlerLocationAI.cs
```diff
@@ -39,36 +39,44 @@ public class SettlerLocationAI {
...
+				
+				// Need to add a way to check freshwater source, and separately to check if coast is lake or coast tile. This score would not apply if the city only borders a lake, although we still need a freshwater bonus
```
QuintillusCFC, line 76: I like the idea of having checks for both freshwater sources and coasts.  The initial bonus was to encourage the AI to settle on coasts instead of one tile off the coasts, which it does to a baffling degree in Civ3, preventing it from both building ships and as importantly, harbors which boost the coastal food yield.

I would advise leaving the water bonus in for now as it's likely to encourage the coastal pattern, and if it also encourages settling by lakes for now, that's not a bad thing as it guarantees fresh water (once irrigation is implemented).  But with a comment about making this more nuanced in the future.

QuintillusCFC: Update: This can just use the new Coastal() method, right?  And then have a comment about eventually checking for nearby freshwater (which IMO could be both less weighty and more complex, as it can be chained across cities so local fresh water sources aren't _necessarily_ that important).

C7GameData/Tile.cs
```diff
@@ -55,12 +58,148 @@ public class Tile
...
+
+				// Skip land tiles and remove from the stack. This should be checked before they're added but just in case
+				if(!currentTile.IsWater())
+				{
+					currentTile.isLake = 0;
+					contiguous.Remove(currentTile);
+					continue;
+				}
+				contiguousWater += 1;
+				if(contiguousWater > 20)
```
QuintillusCFC, line 106: Let's put this 20 magic number in a constant, maybe LAKE_THRESHOLD, to set it up for eventual moddability.  One of the goals of C7 is to reduce the number of hard-coded limitations in Civ3, and while lake size isn't a huge point of consternation, it is ultimately hard-coded.

QuintillusCFC: Also minor, but this long block of if/else/foreach is a good example of why I recommend the style:

```
if (something) {
   //do something
}
```

rather than:

```
if (something)
{
    //do something
}
```

When there are a lot of checks like this, the extra lines from that leading { add up to making it harder to read the logic at a glance due to having to scroll more to see the whole picture, with each scroll adding a little bit of time and overhead re-focusing the eye on where the most-recently-read line is now.  Not a big deal and I see we've been inconsistent about it, but that's the reason I eventually switched to that style from originally giving the opening curly brace its own line.

QuintillusCFC: One last style comment, I realized that `i` and `contiguousWater` are both counting the same thing - how many water tiles we have.  They're just zero-indexed to account for the array and one-indexed because that's how humans think of counting things.  We want to say, if we find a 21st water tile, it's not a lake.  

I'd propose just having one counter variable, something like waterTileCount.  Start it at 0 and increment it in the same place `contiguousWater` is incremented, and it will both grab the right tile from the array and read nicely.  That way someone reading this code won't be trying to figure out the difference between two counters and why one is incremented early and one is incremented late - it'll be easier to verify that the logic works correctly.

C7GameData/Tile.cs
```diff
@@ -55,12 +58,148 @@ public class Tile
...
+				{
+					if(currentTile.IsWater())
+					{
+						currentTile.isLake = 1;
+					}
+				}
+				return true;
+			}
+			else
+			{
```
QuintillusCFC, line 162: I'm pretty sure that this block needs to check the value of the `contiguousWater` variable (or what it eventually is named, see above).  In the case that there are fewer than LAKE_THRESHOLD tiles examined, it's a lake, right?  But if we don't already know that one of the other tiles is a lake, it won't pass the `if` check on line 149.

Maybe the change should be on line 149, checking `if (lake || contiguousWater < LAKE_THRESHOLD)`?

This method is also ripe for a test in C7GameDataTests.  Not saying it needs to be added now, and we might not have enough convenience methods to make it easy to set up the needed data at this point, but the logic is just complex enough that it's a good example of where a test would be high-value.

C7GameData/Tile.cs
```diff
@@ -55,12 +58,148 @@ public class Tile
...
+				Tile t2 = neighbors[d2];
+
+				if(!t1.IsWater() && !t2.IsWater())
+				{
+					// An isthmus separates diagonal water tiles with coastal tiles in both complementary directions
+					return true;
+				}
+			}
+			return false;
+		}
```
QuintillusCFC, line 201: Nicely done.  I doubt this would be called repeatedly during runtime, so I wouldn't go to a lot of extra effort to cache it, but it's never going to change and it would just be storing one variable, so the extra effort isn't a lot.  Your call.

C7GameData/Tile.cs
```diff
@@ -261,6 +443,32 @@ public static (int, int) toCoordDiff(this TileDirection dir)
...
+		// Essentially finds the two other tiles that neighbor BOTH diagonal tiles
+		public static (TileDirection, TileDirection) diagonalComplementary(this TileDirection dir)
+		{
+			switch(dir)
+			{
+				case TileDirection.NORTH: return (TileDirection.NORTHEAST, TileDirection.NORTHWEST);
+				case TileDirection.EAST: return (TileDirection.SOUTHEAST, TileDirection.NORTHWEST);
+				case TileDirection.SOUTH: return (TileDirection.SOUTHEAST, TileDirection.SOUTHWEST);
+				case TileDirection.WEST: return (TileDirection.NORTHWEST, TileDirection.SOUTHWEST);
+				default: throw new ArgumentOutOfRangeException("TileDirection is not Diagonal");
```
QuintillusCFC, line 455: I have mixed feelings on this solution to giving an invalid direction.  Crashing the game is bad.  We had a bug in early 2022 where a method would throw an uncaught exception if it got bad data, and it did sometimes, and the game would mysteriously crash.  This doesn't look as likely to cause that issue but all it would take is someone passing NORTHEAST into this method because they are thinking of Civ4 tiles where NORTHEAST is diagonal.

It may, or may not, be better to instead log an error, but to return a value (a tuple of the same direction passed in, in both parts of the tuple?) rather than crash.  Is the resulting bug in game logic worse than a crash?  It varies, but IMO crashes are the quickest way to negative reviews and players quitting a game (or just never restarting it after the crash).

C7GameData/GameMap.cs
```diff
@@ -57,6 +57,32 @@ public void tileIndexToCoords(int index, out int x, out int y)
 			}
 		}
 
+		// Compute the outer ring of the BFC of tiles. Might not be necessary for tiles which can't be settled?
+		public void computeOuterRing()
+		{
+			// List of all the BFC tiles NOT including direct Neighbors
+			// Essentially, this is every outer tile except North->North, South->South, West->West,East->East
+			(TileDirection direction1, TileDirection direction2) [] outerRingDirections = {(TileDirection.NORTHWEST, TileDirection.NORTH),(TileDirection.NORTHWEST, TileDirection.NORTHWEST),(TileDirection.NORTHWEST, TileDirection.WEST),(TileDirection.SOUTHWEST, TileDirection.WEST),(TileDirection.SOUTHWEST, TileDirection.SOUTHWEST),(TileDirection.SOUTHWEST, TileDirection.SOUTH),(TileDirection.SOUTHEAST, TileDirection.EAST),(TileDirection.SOUTHEAST, TileDirection.SOUTHEAST),(TileDirection.SOUTHEAST, TileDirection.SOUTH),(TileDirection.NORTHEAST, TileDirection.NORTH),(TileDirection.NORTHEAST, TileDirection.NORTHEAST),(TileDirection.NORTHEAST, TileDirection.EAST)};
```
QuintillusCFC, line 65: I like this solution to the outer ring problem.  Certainly more elegant than how I did it in my editor, in a language that (at the time) didn't have tuple support.

One minor suggestion for readability would be having each of the tiles on its own line here, e.g.:

```
(TileDirection direction1, TileDirection direction2) [] outerRingDirections = {
    (TileDirection.NORTHWEST, TileDirection.NORTH),
    (TileDirection.NORTHWEST, TileDirection.NORTHWEST)
    ...
}
```

That would make it easier to verify that the expected 12 are there.  Which they are... but I did validate it just to make sure there weren't any erroneous tiles in the ring.

QuintillusCFC (changes requested):
I believe there are a couple logical errors such as in the isLake method, and not utilizing the outerRing or mining/irrigation improvements in the AI evaluation.  Also noted a few readability suggestions and a few areas that solve the problems elegantly.

All in all this will be a nice improvement in the AI's ability to decide where to settle once merged, and some of the calculations like the ring and isCoastal on Tile.cs will also be useful for future improvements.

C7GameData/Tile.cs
```diff
@@ -55,12 +58,148 @@ public class Tile
 
 		public TileOverlays overlays = new TileOverlays();
 
+
+		// -1 = not initialized. 0 = false, 1 = true
```
Sean-Brown, line 62: alternatively it could be a nullable boolean, i.e.

```
public bool? isLake = null;

public bool IsLake()
{
   if (isLake.HasValue) return isLake.Value;
   ...
}
```

Sean-Brown: Also alternatively it can be a computed property of this class

```
public bool IsLake
{
   get
   {
      if (!IsWater()) return false;
      ...
   }
}
```

WildWeazel:
@Sean-Brown would be up for finishing this one? I haven't heard from @benskywalker

ajhalme:
Closing this off - #913 implements the main ideas and features generalised tile scoring to multiple ranks.

PR #431, opened 2023-09-23
Adding river deltas. Algorithm needs to be checked

### LINKED ISSUE #107: Rivers - Add Delta Support
The file deltaRivers.pcx contains four delta graphics.  These are displayed when rivers empty into the coast (exact formula TBD).  We should include them in our river graphics.

benskywalker:
Fix #107. Deltas seem to work

This looks good
Looks good
Preexisting issue?
This issue is probably unrelated.
Some rivers disappear at the coast. Not sure what's going on here
This might be an issue.

Another success

### DIFF
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -408,9 +408,11 @@ public partial class RiverLayer : LooseLayer
 	public static readonly Vector2 riverSize = new Vector2(128, 64);
 	public static readonly Vector2 riverCenterOffset = new Vector2(riverSize.X / 2, 0);
 	private ImageTexture riverTexture;
+	private ImageTexture riverDeltaTexture;
 
 	public RiverLayer() {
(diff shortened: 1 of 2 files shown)


benskywalker:
I didn't mean to push this IsLake algorithm to this PR, but it should be fine. It kind of fits. It's mostly needed for #430, but hopefully these can both be added in fairly short order

pcen:
This looks awesome! It would be great if you could look into rebasing onto https://github.com/C7-Game/Prototype/pull/407, since we are in the process of updating the tile rendering to use Godot's TileMap class. The logic for choosing which tiles to render deltas on / which delta texture to use should be the same, but you'd need to add some code to load the delta pcx into a tile atlas and then index into that and add it to the tilemap

benskywalker:
@pcen, I've been working on that. I'm closing this PR because it's easier to just make a new branch, but I resolved the issue with rivers not showing in the tilemaps branch at all. #433

PR #433, opened 2023-09-23
Fix 107 Merge River Deltas into Tilemaps

benskywalker:
This PR will merge the river deltas into tilemaps. It also fixed a problem where the tilemaps branch never showed *any* rivers. There's still work to be done; currently the deltas will often show up in the ocean or on a land tile. I should check for river directionality, which will be somewhat more complex


benskywalker:
image
This delta is showing up *in* the ocean. Also the rivers are always rendered in the middle of tiles, which is a problem since you can't easily tell what edge they're on (bad for debugging but also for playing). 

This delta needs to be forward a bit 
image
I've got to find out how to render rivers on the edge of tiles in this new system

pcen:
The way the TileMap renders terrain is terrainTilemap is offset such that it renders the base terrain at the intersection of 4 tiles, so it is offset from tilemap which renders everything else such as hills and forests that fall on a tile. This is done because the terrain that should be rendered on each tile is 4 corners from terrain pcx tiles. I'm not sure how rivers are rendered but it's likely that they could be added as another layer in terrainTilemap?

benskywalker:
Ok, thanks for the info. In that case I think we just need to make it rivers are not offset from the base grid. That way they'll start at corners of tiles like (I think) they do in the base game. Here's an example from the base game:
image

And C7
image

Slight differences are normal, but these rivers consistently pass through the middle of every tile, rather than mainly sticking to the edge

benskywalker:
This is proving more difficult than expected. Rivers are currently marked as a layer in the tilemap, but making them show up at the corner of tiles (or at least shifted a bit) is tricky. 
image
Whenever I mess with setCell in mapView to try and offset rivers there, they don't show up properly. I tried to offset them by half a tile width in x, y, or both, but none of that worked. It might be possible to offset them and I'm just doing it wrong, so I'll keep looking into this

pcen:
I don't think it's really possible to set the offset of a tile in a TileMap - the `stackedCoords` are coordinates that index into a tile. We can set an offset for a particular atlas texture in the TileMap. This is done in the TileSetLoader in TileSetLoader.cs. You can trace the code, but for example:
```
{Atlas.JungleVolcano, new AtlasLoader("Art/Terrain/Volcanos jungles.pcx", 4, 4, mountainSize, 12)},
```
The last parameter is passed into the loader, and eventually gets assigned to `TextureOrigin`, which gives that texture a relative offset to the tile it's in. I'm pretty sure rivers are rendered on tile edges they lay along. I'd suggest adding the river tiles to `terrainTileset` since the entire terrain tilemap is offsest half a tile in both x and y, so if we render a river overtop of base terrain on the terrainTilemap I think the river texture will line up perfectly with tile boundaries. This thread gives a decent explanation of how terrain is rendered: https://forums.civfanatics.com/threads/terrain-editing.622999/

WildWeazel:
@benskywalker have you or can you check the selection logic against the Java code linked to in #107 ? I think that's the best reference we have.

benskywalker:
> @benskywalker have you or can you check the selection logic against the Java code linked to in #107 ? I think that's the best reference we have.

I haven't, but I'll get on that once I've got some more dev time

esbudylin:
Should be resolved by #892

PR #439, opened 2024-04-13
Fix Godot 4 City Graphics Regressions

QuintillusCFC:
These fixes bring the city graphics/label up to equality with Godot 3.5.

1. City label being too wide.  This is fixed by specifying the size of the font used for the label so Godot can calculate it correctly.  (Note it doesn't use the cache, as otherwise it would change the size everywhere).
2. Remove the square box around the city graphics.  This was caused by using shadows when they shouldn't have been used; this is now configurable. (Not 100% sure why this only affected Godot 4)
3. Pop size label now is the correct size


pcen (approved):
nice!

PR #440, opened 2024-04-13
Fix Popups and Font Colors

QuintillusCFC:
Note that the last couple are migrating a bit away from the original goal, but these are all fidelity fixes for Godot 4.

1. Set the header size on popups.  Affects pop-ups such as Main Menu (Esc), Build City, Disband, and Error, all in one line of code.
2. Fix the font colors not being correct (orange, blue) on hover or focus (respectively).  This was because Godot 4 changed the names from e.g. font_color_hover to font_hover_color, and we had the old names.
3. Fix the font size of the button-label text being slightly too small, and not aligned with the buttons (this applies on the popups, e.g. Esc menu, and has also been applied to the Main Menu buttons)
4. Don't load shadows when importing buttons.  Godot 3.5/no-shadows is the faithful way of importing them.
5. Turn off our ancient pre-Aztec sidebar/End Turn buttons; the former was uncollapsed by default and obscuring the buttons on the Build City dialog
6. Right-click box font should be black, not orange.
7. Fix shadow issues with advisor pop-up graphics.
8. Highlighted build city text should be black, not white, as should the cursor

The "pressed" state still results in a white background, but that is also the case in Godot 3.5.


pcen:
for the now-unnecessary offsets, can we just remove them instead of setting them to 0? also, not within the scope of making these fixes, but we might benefit from having a common theme(s?) constructor so that we can change our themes in one place and get a consistent looking UI throughout

pcen (approved):

pcen:
I rebased this change and tested, so I'm going ahead and merging

PR #442, opened 2024-04-17
also search for steam copy of civ3 on windows

pcen:
Title, this also upgrades the Godot version from 4.1 to 4.2.

### DIFF
--- a/C7/C7.csproj
+++ b/C7/C7.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Godot.NET.Sdk/4.1.1">
+<Project Sdk="Godot.NET.Sdk/4.2.1">
   <PropertyGroup>
     <TargetFramework>net7.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
--- a/C7/project.godot
+++ b/C7/project.godot
@@ -12,7 +12,7 @@ config_version=5
 
 config/name="C7"
 run/main_scene="res://MainMenu.tscn"
-config/features=PackedStringArray("4.1", "C#")
+config/features=PackedStringArray("4.2", "C#")
 boot_splash/image="res://icon.png"
 config/icon="res://icon.png"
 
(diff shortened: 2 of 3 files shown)


QuintillusCFC:
Interesting, I'm guessing you are running Steam and the registry method wasn't working?

I checked in regedit (on my Win10 box that has never used the CD version), and my key `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Infogrames Interactive\Civilization III` does list an `Install_Path` of `C:\Games\Steam\steamapps\common\Sid Meier's Civilization III Complete\` (and a `CD_Path` that points to my GOG install...).  Which makes me curious if the problem is that the key can be missing when installed from Steam, or that there's something non-reliable about the registry lookup.  I had some headaches getting the registry lookups to work successfully on all variants of Windows (32 vs 64 bit, XP versus Vista+) on my editor, so I wouldn't be surprised if the key is there but the lookup is failing for some reason.

If the answer is "sometimes the key isn't present", another note for making things more robust - while probably 99% of people install Steam to the default location, I'm that 1% of people who customize it.  It looks like the install path is stored in the registry at `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Valve\Steam`, with the `InstallPath` key; in my case it is `C:\Games\Steam`.  I realize the irony of looking up a registry key to deal with a missing registry key.

Definitely a nice improvement already, making Civ III auto-detection better is important for a smooth out-of-box experience.

(I'm also thinking that adding a Wiki entry to document our findings on the best way to detect Civ install locations may be a good idea.  CD versus Steam versus GOG, but our Linux expert WildWeazel could likely contribute thoughts on the best way for Linux/Proton, and if Jim comes back someday, he may have thoughts on how to improve that on Mac)

QuintillusCFC:
I also just realized that because Steam supports Steam libraries on secondary hard drives, some percentage of the users will have Civ installed on a drive other than the one where Steam itself lives.  I have Steam libraries on three separate drives on my desktop, and will eventually have games on two drives on my laptop.

I'm not sure yet how to auto-detect those cases.  That should probably be a separate enhancement issue.  I'd ballpark maybe 10% of users might have multiple Steam drives?  And some portion of those would have Civ III on the secondary drive.  Probably fairly common on desktops with smaller SSDs and larger HDDs, particularly older desktops, but pretty rare on the more common laptop use case.

pcen:
Yep, in my case Windows 11 machine, only installed Civ 3 via steam, and registry lookup wasn't working. I should check to see if there is a different registry key that contains the Civ 3 path.

For the case of multiple drives, we could probably enumerate mounted drives and check if the default steam path exists on any of the drives

Issue #441, opened 2024-04-13
Godot 4 Regressions


QuintillusCFC:
Keeping track of differences spotted between Godot 3.5 and Godot 4.  This is only the outstanding ones; Paul has fixed a bunch of them, and I just fixed quite a few of the remaining.

- [x] When you press Build City, the highlighted city name has white font; it should have black font (likely a style issue)
- [x] The I-Beam in the same dialog is white (and thus hard to see), rather than black (likely a style issue?)
- [x] The cursor is also white rather than black and hard to see
- [x] The "End Turn" button should not be highlightable nor clickable while the "Build City" dialog is open, but is (this seems to apply to all the buttons as well, though only End Turn has a clear effect)
- [ ] The spacing between right-click-menu items is slightly larger
- [x] The font color on hovered right-click-menu items is orange, not black (likely font_color_hover -> font_hover_color in style overrides)
- [x] The advisor art above the Build City Popup has some odd transparency issues going on, maybe needs shadows turned off?
- [x] The lower-right info box font is slightly too large (or misaligned?) (most visible when you fortify all and "ENTER or SPACEBAR for next turn" starts too far to the left)

pcen:
thanks for helping to close a bunch of these out, the source of the regressions can be tricky to track down!

PR #446, opened 2024-09-28
[godot 4] make caret color black by default

pcen:
Fixes the caret color being white on the godot 4 branch, as described in https://github.com/C7-Game/Prototype/issues/441.

### DIFF
--- a/C7/UIElements/Theme.cs
+++ b/C7/UIElements/Theme.cs
@@ -1,0 +1,15 @@
+using Godot;
+
+public static class ThemeFactory {
+
+	static ThemeFactory() {
+		defaultTheme = new Theme();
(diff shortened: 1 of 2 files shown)


C7/UIElements/Theme.cs
```diff
@@ -0,0 +1,15 @@
+using Godot;
+
+public static class ThemeFactory {
+
+	static ThemeFactory() {
```
pcen, line 5: The idea is to start putting all of the hardcoded theme defaults here. In the future, ThemeFactory can read from a config file so it is customizable (ie. for modding).

WildWeazel: love it

WildWeazel (approved):

PR #447, opened 2024-09-29
Prevent mouse interactions with map and other UI elements when popup is open

pcen:
Fixes the issue with popups described in https://github.com/C7-Game/Prototype/issues/441. Also adjusts the spacing of the "ENTER or SPACEBAR for next turn" so it doesn't start too close to left edge.

### DIFF
--- a/C7/UIElements/Popups/PopupOverlay.cs
+++ b/C7/UIElements/Popups/PopupOverlay.cs
@@ -1,3 +1,4 @@
+using System.Net;
 using Godot;
 using Serilog;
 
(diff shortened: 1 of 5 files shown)


C7/UIElements/Popups/PopupOverlay.cs
```diff
@@ -64,6 +88,18 @@ public void ShowPopup(Popup child, PopupCategory category)
 			log.Error("Invalid popup category");
 		}
 		AudioStreamWav wav = Util.LoadWAVFromDisk(Util.Civ3MediaPath(soundFile));
+
+		// 1. prevent mouse interaction with non-UI elements (ie. the map)
```
pcen, line 92: Long term, need to learn how to use Godot UI properly. Couldn't find a more idiomatic solution on line for this particular problem though.

Short term, I plan to decouple UI from game logic as much as possible. This is kinda a prerequisite for decoupling game from server so that someday we could implement multiplayer.

PR #454, opened 2024-11-14
UnitButton.cs: Only display buttons for available actions

TomWerner:
This avoids the problem of the "road" button showing up when a worker is standing on a tile with a road.

Please let me know if there's a good way to test this.


TomWerner:
### Before

We can see the worker has completed the road, but the road button is still visible.

image


### After

When there is no road, the road button is visible.

image

But once the road is built, the button is gone.

image

WildWeazel (approved):

PR #459, opened 2024-11-20
Introduce a new layer for displaying the "goto" cursor 

TomWerner:
This is a nice "quality of life" feature that also helped me identify
bugs in the movement cost calculations, such as the one fixed in [this
PR](https://github.com/C7-Game/Prototype/pull/457).

This doesn't yet implement a line or some other mechanism to show the
actual planned path, but it both displays the tile that the user wants
to go to, and the cost of getting there in turns.

I moved the code for calculating the cost between two tiles into
`TilePath.cs`, since that's a lower dependency file, and this sort of
common utility should probably be in a low dependency file.

### Example screenshot

 (please excuse the potato quality, I had to take a picture of my screen because the screenshot tool deselects "goto" mode when the mouse loses focus)

Note that despite the destination tile being 2 away from the unit, the cost is only 1, because we can use the road.

image

### DIFF
--- a/C7GameData/AIData/TilePath.cs
+++ b/C7GameData/AIData/TilePath.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 
 namespace C7GameData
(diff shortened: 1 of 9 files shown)


C7GameData/AIData/TilePath.cs
```diff
@@ -29,6 +29,32 @@ public int PathLength() {
 			return path != null ? path.Count : -1;
 		}
 
+		public int PathCost(Tile from, float perTurnMovePoints, float remainingMovementPoints) {
```
WildWeazel, line 32: This should round up, as partial movement still uses a turn. In Civ3 "0" is only shown when movement is completely free.

TomWerner: Partially done. In my git noob-ness, I accidentally ended up with the change applied to https://github.com/C7-Game/Prototype/pull/464/ :(

I'm using to `hg update <this PR>; hg amend; hg evolve; hg update tip; hg uploadchain` to do something like this.

My `git checkout <this PR>; git switch -c <temp branch for this fix>; git switch <child PR>;  git merge <temp branch for this fix>` didn't work. Maybe I wanted `rebase` or something.

Let me know if you want me to mess with this more to get the fix in this PR, sorry about that.

WildWeazel: No problem. I'm reviewing and merging in chronological order to keep the history clean so I hadn't seen the rest of the changes yet.
I don't think there's a simple answer for when you're queuing up multiple linear merges. Well, other than to do it all in one PR, but that's harder to review.

Issue #211, opened 2022-04-08
NodePath Management


pcen:
In the current setup, nodes in every scene have the same name as their type, and are referenced in `GetNode` by relative path. We should define important node paths in constants, and possible only use absolute paths (in non-trivial cases) to make the code consistent and less error-prone.

esbudylin:
The issue was addressed in #463 and no longer seems relevant. I'm closing it.

PR #463, opened 2024-11-22
[godot4] Manage node paths with export properties

esbudylin:
This pull request addresses the issue with node path management described in #211. 

Instead of separating node paths into constants, as suggested in the issue, this PR takes a different approach. It uses Godot export properties [1] to separate node paths from the source code and move them into the scene editor. This allows nodes to be easily selected within the currently edited scene. This also ensures that the node paths will be automatically updated when relevant nodes are moved, deleted or renamed in the editor.

As a demo of this approach, I've started with refactoring the UnitButtons class. I'm ready to refactor other classes as well.

[1] https://docs.godotengine.org/en/4.2/tutorials/scripting/c_sharp/c_sharp_exports.html

### DIFF
--- a/C7/C7Game.tscn
+++ b/C7/C7Game.tscn
@@ -58,8 +58,12 @@ _data = {
 "SlideOutAnimation": SubResource("1")
 }
 
-[node name="C7Game" type="Node2D"]
+[node name="C7Game" type="Node2D" node_paths=PackedStringArray("Toolbar", "popupOverlay", "slider", "animationPlayer")]
 script = ExtResource("1")
+Toolbar = NodePath("CanvasLayer/Control/ToolBar/MarginContainer/HBoxContainer")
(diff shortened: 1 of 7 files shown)


WildWeazel:
@pcen @QuintillusCFC curious to get your thoughts on this, as it's kind of a design convention. The proposal is to do this project-wide.

pcen:
Nice!! This looks a lot better to me than using constants.

esbudylin:
I've refactored the rest of the project with export properties, I think this PR is ready for review and merging. There are still a few instances of GetNode in the project; however, I think there are fine since the remaining calls use paths that are only one or two levels deep.

WildWeazel:
This looks like a good step towards best practices. I'm familiar with a similar pattern in Unity. It should help keep things cleaner as we (as I would like to see) move towards a more scene-based UI.

PR #464, opened 2024-11-23
Draw dots on the path the unit will take in "goto" mode

TomWerner:
This gets us to something much more like civ3's "goto" mode, where we can see the entire path that will be taken.

image

Sometime this looks a little goofy, I assume due to quirks of the pathing algorithm, but this is a good thing - it allows the player to choose something else if the pathed version is bad.

image

### DIFF
--- a/C7GameData/AIData/TilePath.cs
+++ b/C7GameData/AIData/TilePath.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 
 namespace C7GameData
@@ -29,6 +30,26 @@ namespace C7GameData
 			return path != null ? path.Count : -1;
 		}
 
+		public int PathCost(Tile from, float unitMovementPoints) {
+			if (path == null) { return 0; }
+
+			int turns = 0;
+			float remainingMovementPoints = unitMovementPoints;
+			foreach (Tile tile in path) {
+				float cost = getMovementCost(from, from.directionTo(tile), tile);
+				Console.WriteLine("Cost from " + from.ToString() + " to " + tile.ToString() + " is " + cost);
+
+				remainingMovementPoints -= cost;
+				if (remainingMovementPoints <= 0) {
+					++turns;
+					remainingMovementPoints = unitMovementPoints;
+				}
+
+				from = tile;
+			}
+			return turns;
+		}
+
 		// Indicates no path was found to the requested destination.
 		public static TilePath NONE = new TilePath();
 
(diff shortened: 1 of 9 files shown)


C7GameData/AIData/TilePath.cs
```diff
@@ -37,5 +74,22 @@ public static TilePath EmptyPath(Tile destination) {
 			return new TilePath(destination, new Queue<Tile>());
 		}
 
+		public static float getMovementCost(Tile from, TileDirection dir, Tile newLocation) {
+			// River crossings disrupt roads, so check that first.
+			if (from.HasRiverCrossing(dir)) return newLocation.MovementCost();
+
+			// Travelling between two tiles with railroads is free.
```
Sean-Brown, line 92: shouldn't this logic come first?

WildWeazel: I had to test this and technically no, if you don't have the "allows bridges" tech, crossing a river on a railroad does use movement! The real omission is checking for that flag, but nether it nor railroads are implemented yet.
Good to see you back @Sean-Brown :+1:

C7GameData/AIData/TilePath.cs
```diff
@@ -52,9 +53,29 @@ public int PathCost(Tile from, float perTurnMovePoints, float remainingMovementP
...
 			}
+
+			// Special case: if we consumed part of our movement points (such as by
+			// walking along a road, consuming 1/3 of a point), return a cost of 1.
+			// This doesn't apply when railroads are used, since that wouldn't change
+			// movementPoints.remaining. If it wasn't for railroads we could just
+			// return Max(turns, 1).
+			if (turns == 0 && movementPoints.remaining != remainingMovementPoints) {
+				return 1;
+			}
```
WildWeazel: True, but it still needs to round up if any movement points are consumed on the final turn. For example if I goto 4 or 5 tiles along a road it shows 1 even though the movement won't complete until the second turn. The general case for this final check would be something like `if (movementPoints.remaining < (turns == 0 ? remainingMovementPoints : perTurnMovePoints) ) ++turns`

WildWeazel (changes requested):

C7GameData/AIData/TilePath.cs
TomWerner: Ah, good catch. I hadn't tested 4 moves along a road. Your suggestion worked perfectly, thanks.

C7GameData/AIData/TilePath.cs
Sean-Brown: interesting, I assumed that if you have railroads then you have a bridge crossing tech

PR #467, opened 2024-11-25
Settler.ini: Fix the path format for the death sound

TomWerner:
Prior to this the game would freeze, and the console would show errors like `System.ApplicationException: Media path not found: Art/Units/Settler/..\Worker\WorkerDeath.wav`.

The other paths in this file already use the correct format.


WildWeazel:
This led me down a bit of a rabbit hole. I reproduced the issue and then confirmed the fix, which was surprising to me because in either case I don't see the correct animations from this file! After chasing down the animation path resolution issue in the debugger, I have a very simple working solution to that by using "." instead of "" as the base case root directory, but again surprised because none of the relevant file search code seems to have changed in godot4 and it was working in the last release. There seems to be a different code path for resolving the ini for sound vs animation that I haven't quite sorted out yet. So I want to keep this open until we have a better understanding of what's going on and why, because I think we need to take a wider look at how and when file paths are resolved.

TomWerner:
> which was surprising to me because in either case I don't see the correct animations from this file!

That's very strange - the animations definitely play correctly for me. I took a screenshot mid-death animation.

image

If there are larger issues with path loading we should probably open a separate issue for that. This PR seems to fix the game crashing issue though, so it seems worthwhile on its own.

WildWeazel:
That's the regular Civ3 settler unit though, not the Carthaginian settler that we're supplying.  Here it is as of Carthage Preview 3: https://forums.civfanatics.com/attachments/1665107157229-png.641129/
Edit: To be more clear, the issue seems to be that the animation code finds the wrong ini file, but the audio code uses the correct one. My workaround was to change the root path in FileExistsIgnoringCase on line 148 of Util.cs, but it's not clear to me why this was working before or what changed.

TomWerner:
Emerging from the rabbit hole, https://github.com/C7-Game/Prototype/pull/475 fixes the issue of the wrong settler art being used.

I still think this PR is a good idea as a standalone PR, and is still needed even after fixing the art loading issue.

PR #468, opened 2024-11-25
Ensure that destroyed cities no longer show up in the UI  (no longer preferred)

TomWerner:
Now if a city is captured it is removed from the UI and the implicit
city road is removed.

Let me know if this is the preferred approach. I could also see a valid implementation where we exposed the `cityLayer` in `MapView.cs`, added a new `MessageToUI`, and had `Game.cs` do the modifications to the city layer. I'm not sure what's the stylistic choice; the latter is a bit more complex, but might work better for capturing cities in the future.

To make this easier to test I modified `c7-static-map-save.json` locally
to add this:

```
    {
      "id": "Warrior-1",
      "prototype": "Warrior",
      "owner": "player-2",
      "previousLocation": {
        "x": -1,
        "y": -1
      },
      "currentLocation": {
        "x": 57,
        "y": 43
      },
      "hitPointsRemaining": 5,
      "movePointsRemaining": 1,
      "facingDirection": "north",
      "experience": "Elite"
    },
```

and modified this (to change the x/y position so an AI spawns next to
us, so our added warrior can destroy the city)

```
    {
      "id": "Settler-2",
      "prototype": "Settler",
      "owner": "player-3",
      "previousLocation": {
        "x": -1,
        "y": -1
      },
      "currentLocation": {
        "x": 59,
        "y": 43
      },
      "hitPointsRemaining": 3,
      "movePointsRemaining": 1,
      "facingDirection": "north",
      "experience": "Regular"
    },
```

### DIFF
--- a/C7/Map/CityLayer.cs
+++ b/C7/Map/CityLayer.cs
@@ -7,8 +7,12 @@ namespace C7.Map {
 	public class CityLayer : LooseLayer {
 
 		private ILogger log = LogManager.ForContext<CityLayer>();
+
+		private ImageTexture cityTexture;
+		private Dictionary<string, ImageTexture> cityLabels = new Dictionary<string, ImageTexture>();
+
+		private List<City> citiesWithScenes = new List<City>();
 		private Dictionary<City, CityScene> citySceneLookup = new Dictionary<City, CityScene>();
-		private Dictionary<Tile, City> tileCityLookup = new();
 
 		public CityLayer()
 		{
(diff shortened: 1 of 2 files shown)


TomWerner:
I think I prefer https://github.com/C7-Game/Prototype/pull/469 to this, since it would allow for capturing cities. I'm not sure how that would be implemented in the other approach.

I'll leave this one open until the other is accepted though.

WildWeazel:
merged #469

PR #470, opened 2024-11-26
Make the retire button on the popup menu work

TomWerner:
This now takes you back to the main menu, where you can start a new game, retire, start a new game, etc.

I don't fully understand the godot4 magic at work here, but was able to cobble this together by copying the other button's behavior.


QuintillusCFC (approved):
The other part of the Godot magic is the `_on_QuitButton_pressed` method in Game.cs, and that C7Game.tscn is linked with Game.cs (how exactly I'm not sure, though I see a reference on line 2) and can thus call its methods.

Looks good, tested and working.  I fired up the Godot 3.5 version too, and Retire never actually worked there.  Instead we had a menu on the right side (from way back before the Aztec release) that had a return-to-menu button.  That menu got removed at some point in the Godot 4 switch - but thanks to this PR it's now possible to return to the main menu again!

PR #473, opened 2024-12-02
Fix scenario loading with no cities

TomWerner:
This is sufficient for scenarios 1 (which starts with no cities) and 6
(which starts with some units and cities, the cities just don't render
yet).

This doesn't yet tackle loading the cities, which requires some extra
logic around viable units to produce, city naming (see the added TODO),
and some annoying index differences between leaders and playable civs.

Scenario 1: 
image

Scenario 6: 
image


TomWerner:
I fixed the merge issue, this should be good to go now.

PR #475, opened 2024-12-04
Fix settler animation loading

TomWerner:
This ensures that the C7 settler art is actually used, instead of the civ3 settler.

### DIFF
--- a/C7/Util.cs
+++ b/C7/Util.cs
@@ -81,6 +81,14 @@ public partial class Util {
 	public static string FileExistsIgnoringCase(string exactCaseRoot, string ignoredCaseExtension) {
 		// First try the basic built-in File.Exists method since it's adequate in most cases.
 		string fullPath = System.IO.Path.Combine(exactCaseRoot, ignoredCaseExtension);
+
+		// In debug builds paths that are specified relative to the project directory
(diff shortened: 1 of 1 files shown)


C7/Util.cs
```diff
@@ -81,6 +81,14 @@ static public string Civ3PathFromRegistry() {
 	public static string FileExistsIgnoringCase(string exactCaseRoot, string ignoredCaseExtension) {
 		// First try the basic built-in File.Exists method since it's adequate in most cases.
 		string fullPath = System.IO.Path.Combine(exactCaseRoot, ignoredCaseExtension);
+
+		// In debug builds paths that are specified relative to the project directory
+		// will start with res://. We want to turn those into real paths, and then
+		// update our exactCaseRoot variable so that the search logic below works.
+		if (fullPath.Contains("res://")) {
```
WildWeazel: Would it not make more sense to check if exactCaseRoot contains (starts with?) "res://" and if so globalize it before combining it with the extension, rather than combining first and then extracting a substring? Or is "res://" maybe going to appear in the extension?

TomWerner: Much simpler - thanks!

I thought I had tried that, but must have goofed something up in my earlier hacking.

PR #478, opened 2024-12-08
Fix unresponsive unit buttons

esbudylin:
Prior to this commit, some of the unit buttons (e.g., Build City, Build Road) could become unresponsive. This issue was related to how the game handled button presses. When a button was pressed, the game emulated the press of an action hotkey, which should have been handled by the Game class. However, in some cases, the input simulation didn't work properly, and the game didn't react to the button presses.

To solve this issue, I decoupled the handling of button presses from the handling of hotkey input. Now, instead of imitating input, a unit button sends a signal to the Game class. In the Game class, the logic for handling action requests has been moved into a new method, which is now used for both handling signals from unit buttons and processing hotkey input.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -509,10 +509,10 @@ public partial class Game : Node2D {
 		}
 
 		if (this.HasCurrentlySelectedUnit()) {
-			(bool moveUnit, TileDirection dir) = C7Action.ToTileDirection(currentAction);
+			TileDirection? dir = C7Action.ToTileDirection(currentAction);
 			
-			if (moveUnit) {
-				new MsgMoveUnit(CurrentlySelectedUnit.id, dir).send();
+			if (dir.HasValue) {
+				new MsgMoveUnit(CurrentlySelectedUnit.id, dir.Value).send();
 				setSelectedUnit(CurrentlySelectedUnit); //also triggers updating the lower-left info box
 			}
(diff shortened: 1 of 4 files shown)


pcen:
Is this a bug in Godot or incorrect usage? Conceptually this makes sense regardless

esbudylin:
> Is this a bug in Godot or incorrect usage? Conceptually this makes sense regardless

I'm not sure, but my guess is that there were some subtle changes to the `Input` class between Godot versions that weren't properly documented. I couldn't find any records about that in the Godot bug tracker.

C7/Game.cs
```diff
@@ -504,21 +512,21 @@ private void processActions() {
 			// TODO: replace bool with an invalid TileDirection enum
 			TileDirection dir = TileDirection.NORTH;
 			bool moveUnit = true;
-			if (Input.IsActionJustPressed(C7Action.MoveUnitSouthwest)) {
+			if (currentAction == C7Action.MoveUnitSouthwest) {
```
Sean-Brown: possibly should be a switch statement?

WildWeazel:
This seems sensible to me. I'm a little unsure about handling overlapping inputs for different actions (eg, if we implement Ctrl-G for the grid, which also activates G for go-to) but it's no worse than before.

PR #479, opened 2024-12-10
Add civ destruction

TomWerner:
... and display a popup so the user knows they accomplished this.

This gets us a step closer to being able to "win" a game of C7, instead of just being able to produce units and fight.

To make the popup dialog read reasonably well I had to fill in the "noun" field for each civilization.

As with many of these city/settler related PRs, I used the procedure from https://github.com/C7-Game/Prototype/pull/469 to have the AI build a city right next to me on turn 2.

---

Some screenshots are below. In particular, note that the American worker is destroyed at the same time the city is.

image

image

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -171,6 +171,15 @@ public partial class Game : Node2D {
 					break;
 				case MsgCityDestroyed mCD:
 					mapView.cityLayer.UpdateAfterCityDestruction(mCD.city);
+
+					// If this was the last city of the civilization, display a popup
+					// noting that the civ is gone and destroy any remaining units.
+					if (mCD.city.owner.RemainingCities() == 0) {
+						popupOverlay.ShowPopup(new CivilizationDestroyed(mCD.city.owner.civilization), PopupOverlay.PopupCategory.Advisor);
+						for (int i =0; i < mCD.city.owner.units.Count; ++i) {
+							MapUnitExtensions.disband(mCD.city.owner.units[i]);
+						}
+					}
 					break;
 			}
 		}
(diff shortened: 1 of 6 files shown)


C7/Game.cs
```diff
@@ -171,6 +171,15 @@ public void processEngineMessages(GameData gameData) {
 					break;
 				case MsgCityDestroyed mCD:
 					mapView.cityLayer.UpdateAfterCityDestruction(mCD.city);
+
+					// If this was the last city of the civilization, display a popup
+					// noting that the civ is gone and destroy any remaining units.
+					if (mCD.city.owner.RemainingCities() == 0) {
```
esbudylin, line 181: If I recall correctly, in Civ 3 a civilization will stay alive as long as it has at least one settler, even if all its cities are destroyed.

Sean-Brown: I believe that's correct, and also the game conditions can decide this, e.g. 1 city elimination, regicide, etc. For now this may be a good starting place until we implement game rules.

TomWerner: Agreed. I've added a TODO to note this discrepancy, but I think this is still worth doing.

Adding the settler check means we also need to run this type of check every time a unit is destroyed, in case it was the last settler, and this at least gives us a starting point.

Sean-Brown: since "civilization destroyed" can occur due to several conditions, perhaps the logic should be its own event (or at least the logic moved to a common location)

WildWeazel: Agreed, this is good enough for a first pass but can be extracted to an event later with all the proper checks

PR #480, opened 2024-12-14
Refactor handling of Godot signals

esbudylin:
This commit addresses two issues with Godot signal handling, both of which relate to the transition between engine versions. In Godot 3, these were common practices for handling signals; however, the Godot 4 signal API offers more type-safe alternatives.

1) Use of raw strings for signal identification

Previously, raw strings were used to identify signals during emission. This has been improved by using the nested SignalName class provided by Godot signal API.

2) Use of the Connect() method for connecting signals

The Connect() method was used to connect methods to signals, a practice discouraged by the Godot documentation. This commit replaces Connect() with event mechanism.

[Relevant Godot documentation](https://docs.godotengine.org/en/stable/tutorials/scripting/c_sharp/c_sharp_signals.html)

### DIFF
--- a/C7/Credits.cs
+++ b/C7/Credits.cs
@@ -62,7 +62,7 @@ public partial class Credits : Node2D
 		goBackButton.TextureNormal = goBackTexture;
 		goBackButton.SetPosition(new Vector2(952, 720));
 		AddChild(goBackButton);
-		goBackButton.Connect("pressed",new Callable(this,"ReturnToMenu"));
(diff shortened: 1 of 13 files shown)


Sean-Brown:
lgtm, nice work and explanation

WildWeazel:
Now this looks like the C# I know. It's also more IDE friendly.

@pcen @QuintillusCFC for your reference- this is now the recommended pattern in Godot 4

PR #484, opened 2024-12-30
Code cleanup to address godot4 PR comments

WildWeazel:
Resolved some of the open comment threads by pruning dead or commented out code and renaming things. Civ3Map.LoadTileSet did not seem to serve any purpose anymore, as it did not return or modify anything outside of the method. I did not address the question of what purpose MovingSprite serves in Civ3Unit or what its values mean.

### DIFF
--- a/C7/PCXToGodot.cs
+++ b/C7/PCXToGodot.cs
@@ -4,6 +4,17 @@ using ConvertCiv3Media;
 using System.Collections.Generic;
 
 public partial class PCXToGodot : GodotObject {
+	private const byte SHADOW_START_INDEX = 224;
+	private const byte CIVCOLOR_START_INDEX = 239;
+	private const byte TRANSPARENCY_START_INDEX = 254;
+	private const byte SHADOW_INDEX_RANGE = CIVCOLOR_START_INDEX - SHADOW_START_INDEX;
+	private const ushort MAX_PALETTE_SIZE = 256;
+
+	private const byte GREEN_BITSHIFT = 8;
+	private const byte BLUE_BITSHIFT = 16;
+	private const byte ALPHA_BITSHIFT = 24;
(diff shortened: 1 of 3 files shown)


C7/Civ3Map/Civ3Map.cs
```diff
@@ -48,51 +48,9 @@ public void TerrainAsTileMap() {
 		{
 			foreach (Tile tile in Civ3Tiles)
 			{
-				// If tile media file not loaded yet
-				if (TileIDLookup[tile.ExtraInfo.BaseTerrainFileID, 1] == 0) { LoadTileSet(tile.ExtraInfo.BaseTerrainFileID); }
```
QuintillusCFC: Should the remaining references to TileIDLookup also be removed?  The lines that are now lines 51-52, just below this removed block, look odd now, setting the Map int[] to 0, and then looking it up via TileIDLookup which will now always be zero.

Granted I don't entirely remember the purpose of this section.  It looks like it was added back in March of '21, to implement a legacy map base terrain viewer.  But I suspect those lines can additionally be removed.  The changes are working for me in-game (and the much-decreased amount of magic numbers is nice).

QuintillusCFC (approved):

C7/Civ3Map/Civ3Map.cs
lexer-gamedev: +1 to this, it doesn't seem to be used

otherwise lgtm

lexer-gamedev (changes requested):

C7/Civ3Map/Civ3Map.cs
TomWerner: The Civ3Map class was removed a while back.

TomWerner:
I fixed the merge conflicts and am getting this merged in to clean up some of the open PRs.

PR #491, opened 2025-01-05
ImportCiv3.cs: Add support for loading cities in scenarios

TomWerner:
This still has some limitations, in that it will only work if there is a "Worker" unit. The `producible` field really should be loading from the list of unit prototyes, but that can be done later.

Some example screenshots:

Screenshot 2025-01-05 1 32 22 PM
Screenshot 2025-01-05 1 31 06 PM

The labels on the cities seems slightly amiss, and the unit health bars are somewhat obscured by the cities, but I don't think that's an issue with this PR.


TomWerner:
I'm going to merge this one, pcen approved 492 which had this PR as a diffbase.

PR #489, opened 2025-01-05
Run the code formatter over all the C# code

TomWerner:
I'd like to be able to have "format on save" turned on without it generating obnoxiously large diffs.

Doing this all at once makes it clearer this is safe.

To do this I configured VSCode [according to the wiki instructions](https://github.com/C7-Game/Prototype/wiki/Developing-and-Setting-Up-IDEs#code-formatting) and then used the [workspace formatter extension](https://marketplace.visualstudio.com/items?itemName=franneck94.workspace-formatter) to run the formatter on the relevant directories.

#406

PR #494, opened 2025-01-08
Run the formatter on the QueryCiv3 and C7GameData folders

TomWerner:
I used the same procedure as in https://github.com/C7-Game/Prototype/pull/489

Without this I can't have format on save turned on, it made a spotting a one line diff in a following PR almost impossible.

If these PRs are annoying I can just turn off autoformat.


pcen:
This are great, everyone should enable format on save moving forward imo.

pcen (approved):

PR #501, opened 2025-01-13
Fix the unit tests

TomWerner:
There were some minor formatting changes with the canned save causing failures.

I'll work on getting these tests to run on PRs to avoid this in the future.

```
[xUnit.net 00:00:07.19]     SaveTests.SimpleSave [FAIL]
  Failed SaveTests.SimpleSave [3 s]
  Error Message:
   Assert.Equal() Failure
Expected: Byte[] [123, 10, 32, 32, 34, ...]
Actual:   Byte[] [123, 10, 32, 32, 34, ...]
  Stack Trace:
     at SaveTests.SimpleSave() in /home/tomwer3/C7_real/Prototype/C7GameDataTests/SaveTest.cs:line 71
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

Failed!  - Failed:     1, Passed:     5, Skipped:     0, Total:     6, Duration: 7 s - C7GameDataTests.dll (net8.0)
```

### DIFF
--- a/C7/Text/c7-static-map-save.json
+++ b/C7/Text/c7-static-map-save.json
@@ -58661,7 +58661,7 @@
       "facingDirection": "north",
       "experience": "Regular"
     },
-	{
+    {
       "id": "Settler-2",
       "prototype": "Settler",
       "owner": "player-3",
@@ -61858,7 +61858,7 @@
   "civilizations": [
     {
       "name": "A Barbarian Chiefdom",
-	  "noun": "Barbarians",
+      "noun": "Barbarians",
       "colorIndex": 0,
       "leaderGender": "male",
       "cityNames": [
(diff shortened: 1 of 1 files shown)


pcen (approved):

pcen:
Thanks! The way I implemented this unit test is fairly brittle, but it's a bit simpler than comparing the json in a smarter way

TomWerner:
Thanks for the quick review :)

The test seems fine, it's easy to fix via `cp ../C7GameDataTests/data/output/static-save-never-game-data.json Text/c7-static-map-save.json`

I'll send a child PR of https://github.com/C7-Game/Prototype/pull/488 that runs the tests on PRs automatically so we don't hit this again.

PR #502, opened 2025-01-13
Run unit tests automatically on new PRs and commits to Development

TomWerner:
This required one slight change to a unit test that was assuming Civ 3 was installed on the machine running the test; the rest of the tests work fine.

Here's the latest run: https://github.com/C7-Game/Prototype/actions/runs/12739968114

### DIFF
--- a/C7GameDataTests/SaveTest.cs
+++ b/C7GameDataTests/SaveTest.cs
@@ -106,6 +106,14 @@ public class SaveTests {
 
 	[Fact]
 	public void LoadAllConquests() {
+		// When running the tests via github actions, civ3 isn't installed so we can't
+		// check the conquests directories.
+		//
+		// See https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
+		// for a full list of env vars.
+		string is_on_github = System.Environment.GetEnvironmentVariable("CI");
+		if (is_on_github != null) { return; }
+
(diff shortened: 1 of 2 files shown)


C7GameDataTests/SaveTest.cs
```diff
@@ -106,6 +106,11 @@ public async void LoadSampleSaves() {
 
 	[Fact]
 	public void LoadAllConquests() {
+		// When running the tests via github actions, civ3 isn't installed so we can't
+		// check the conquests directories.
+		string is_on_github = System.Environment.GetEnvironmentVariable("RUNNING_ON_GITHUB");
```
pcen: Instead of introducing the concept of being in ci (`RUNNING_ON_GITHUB`) to the codebase, perhaps LoadAllConquests should check if there's a folder at `conquests` and return if nothing is there?

TomWerner: I thought about that, but do we want the tests to silently pass if a dev hasn't set things up correctly?

pcen: Good point. In that case, let's check if CI is true or GITHUB_ACTIONS is in ENV instead of adding a new env variable? https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

TomWerner: Nice idea, done.

Issue #288, opened 2022-07-05
Draw territory borders on the map


QuintillusCFC:
Territory borders make it easier to see whose territory is whose.

We might need to add a method to calculate who owns the tiles too, as I don't think we have that.

Reference for drawing: https://hg.sr.ht/~adj/civ3_cross_platform_editor/browse/src/main/java/com/civfanatics/civ3/xplatformeditor/tabs/map/renderer/ClassicRenderer.java?rev=tip#L963
Reference for calculating tile ownership: https://hg.sr.ht/~adj/civ3_shared_components/browse/src/main/java/com/civfanatics/civ3/biqFile/IO.java?rev=tip#L5935

This is not really a Carthage goal, possibly a Dutch or Egypt one.

esbudylin:
I've started to work on this issue. I'll probably open a PR in a week or two.

esbudylin:
Drawing borders is now implemented, but the issue remains relevant. There are still minor graphical inconsistencies (see #504 for details) and, more importantly, no culture-based territory calculations. These seem like more distant goals to me.

TomWerner:
This can be closed now, culture-based border drawing is fully implemented.

PR #504, opened 2025-01-17
Draw civilization borders

esbudylin:
This PR implements drawing borders on the game map (#288). The border drawing logic is heavily based on the code from Quintilus' editor.

This PR has following limitations:
* Territory calculation based on culture is not implemented. A placeholder method is used instead, which marks territory as city tiles and tiles neighboring cities.
* In Civ3 two colors are used to color a border of each civ. The first color is the color of civilization, the second color is derived from it. I haven't figured out the exact calculations Civ3 uses to derive this color, so a rough approximation is used instead. 
* The original game draws border over hill and mountains in rather nuanced way, which is not replicated in this PR. 
* Also, there is no border rendering on the edge of the fog of war (see screenshot). I'm not sure whether the problem lies with the border or the fog of war layer.
image

### DIFF
--- a/C7/Map/BorderLayer.cs
+++ b/C7/Map/BorderLayer.cs
@@ -1,0 +1,78 @@
+using System.Collections.Generic;
+using C7GameData;
+using Godot;
+using ConvertCiv3Media;
+
+namespace C7.Map {
+	public class BorderLayer : LooseLayer {
+		private readonly string texturePath = "Art/Terrain/Territory.pcx";
+		private readonly ImageTexture[] borderGraphics = new ImageTexture[8];
+		private readonly Dictionary<(int, Color), ImageTexture> textureCache = new();
+
+		public BorderLayer() {
+			Pcx texturePcx = Util.LoadPCX(texturePath);
+
+			int textureWidth = 128;
+			int textureHeight = 72;
+
+			for (int j = 0; j < 4; j++) {
+				for (int k = 0; k < 2; k++) {
+					borderGraphics[j * 2 + k] = PCXToGodot.getImageTextureFromPCX(texturePcx, k * textureWidth, j * textureHeight, textureWidth, textureHeight, true);
+				}
+			}
+		}
+
+		// TODO: This method doesn't precisely mirror Civ3 coloring
+		private static Color CalcSecondaryColor(Color mainColor) {
+			return new Color(mainColor.R * 0.8f, mainColor.G * 0.8f, mainColor.B * 0.8f);
+		}
+
+		private ImageTexture GetBorderTexture(int textureIndex, Color borderColor) {
+			if (textureCache.TryGetValue((textureIndex, borderColor), out ImageTexture res)) {
+				return res;
+			}
+
+			ImageTexture texture = borderGraphics[textureIndex];
+
+			Color secondaryColor = CalcSecondaryColor(borderColor);
+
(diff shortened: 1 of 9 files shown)


C7/Map/BorderLayer.cs
```diff
@@ -0,0 +1,78 @@
+using System.Collections.Generic;
+using C7GameData;
+using Godot;
+using ConvertCiv3Media;
+
+namespace C7.Map {
+	public class BorderLayer : LooseLayer {
```
TomWerner, line 8: Could you leave a brief comment documenting this class?

C7/Map/BorderLayer.cs
```diff
@@ -0,0 +1,78 @@
...
+		private readonly string texturePath = "Art/Terrain/Territory.pcx";
+		private readonly ImageTexture[] borderGraphics = new ImageTexture[8];
+		private readonly Dictionary<(int, Color), ImageTexture> textureCache = new();
+
+		public BorderLayer() {
+			Pcx texturePcx = Util.LoadPCX(texturePath);
+
+			int textureWidth = 128;
```
TomWerner, line 16: Could you make maybe a small ascii art drawing of the layout of this pcx file? How do the indices work?

C7/PCXToGodot.cs
```diff
@@ -1,9 +1,10 @@
 using Godot;
 using System;
 using ConvertCiv3Media;
+using System.Collections.Generic;
 
 public partial class PCXToGodot : GodotObject {
-	private readonly static byte CIV3_TRANSPARENCY_START = 254;
+	private static readonly HashSet<int> transparents = new() { 1, 254, 255 };
```
TomWerner: Could you document this field? Are these the set of color indices that should be treated as transparent? Maybe "transparentColorIndices"/"transparentColorIndexes"?

C7GameData/Tile.cs
```diff
@@ -3,11 +3,13 @@ namespace C7GameData {
 	using System.Text.Json.Serialization;
 	using System.Collections.Generic;
 	using System.Linq;
+
 	public class Tile {
 		public ID Id { get; internal set; }
 		public Civ3ExtraInfo ExtraInfo;
 		public int xCoordinate;
 		public int yCoordinate;
+		public Player owner;
```
TomWerner: Could you document this field?

C7/Map/BorderLayer.cs
```diff
@@ -0,0 +1,78 @@
...
+
+			Color secondaryColor = CalcSecondaryColor(borderColor);
+
+			Dictionary<Color, Color> colorReplacements = new Dictionary<Color, Color>
+			{
+				{ Color.Color8(236, 255, 0), borderColor },
+				{ Color.Color8(255, 0, 0), secondaryColor}
+			};
+
+			Image image = Util.TransformColors(texture.GetImage(), colorReplacements);
```
TomWerner, line 57: Could you leave a brief comment explaining the broad idea of what's happening here?

Something like:

```
// The texture template uses two base colors (<foo> and <bar>), so to make this a civ-specific
// template we need to replace them with this civ's border color. Once we've done the
// replacement, stick it in the cache so we only have to do this relatively expensive pixel-by-pixel
// scan once per texture.
```

esbudylin:
@TomWerner thanks for review! I pushed commits with additional comments. Let me know what you think!

TomWerner (approved):

PR #497, opened 2025-01-11
UI improvements: save menu

esbudylin:
Menu for loading saves and scenarios has been made bigger and centered on the screen. The file dialog has been refactored into a separate scene to make it reusable.

PR #512, opened 2025-01-20
Implement the in-game load game menu

TomWerner:
I'm pretty sure this isn't the right way to do this, since the load game menu isn't centered or the right size, like it was after #497. However, I think the fact that the `GameMenu` class is an anonymous node (the path is something like `/root/C7Game/CanvasLayer/PopupOverlay/@TextureRect@350"`, where the number changes run-to-run) means that the approach from #497 doesn't work. Maybe more things need to become their own small scenes? I don't know enough about godot to say for sure.

This seems like an improvement though, since a new button works, but I'd welcome any advice on how to do this the "correct" way.


esbudylin:
My advice is to create an instance of the 'civ3_file_dialog.tscn' scene within the C7Game scene. This way, you will be able to reuse the file dialog's UI appearance. You can instantiate it either in the editor (I think this is the preferable option, as it makes the scene's structure more discoverable) or via code.

esbudylin:
I've centered the load menu on the screen and made some minor refactoring. I think this PR is ready for review and merge.

TomWerner:
Very nice, thanks! TIL about the '%' part of Godot node paths.

I just tested it locally and it looked good, but I can let someone else approve.

pcen (approved):

Issue #392, opened 2023-02-09
Display Tech Advisor Screen - Basic Layout


QuintillusCFC:
This is very much a stretch, optional goal for the tech milestone; advisors were originally planned for much later.

But if someone is hankering to do some graphics work, this would be a good chunk of graphics work.

Lay out the techs based on the x/y positions they have in the BIQ (may have to be imported).  Display their tech icons.

Showing buildings, potentially across a second row, and selecting the tech would come later.

PR #513, opened 2025-01-21
Load the Civ3 techs from save games and scenarios

TomWerner:
This doesn't yet add any support for rendering them or working on techs, but this is the first step towards doing that.

#392

### DIFF
--- a/C7GameData/Save/SaveTech.cs
+++ b/C7GameData/Save/SaveTech.cs
@@ -1,0 +1,37 @@
+using System.Collections.Generic;
+
+namespace C7GameData.Save {
+	// A class representing a single technology that can be researched.
+	//
+	// This is intended for use in save games, so it does not have recursive
+	// references for prereqs.
+	public class SaveTech {
+		public ID id;
+		public string Name { get; set; }
+		public string CivilopediaEntry { get; set; }
+		public int Cost;
+		public string Era { get; set; }
+		public int AdvanceIcon;
+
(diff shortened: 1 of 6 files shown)


C7GameData/Save/SaveTech.cs
```diff
@@ -0,0 +1,37 @@
...
+	// This is intended for use in save games, so it does not have recursive
+	// references for prereqs.
+	public class SaveTech {
+		public ID id;
+		public string Name { get; set; }
+		public string CivilopediaEntry { get; set; }
+		public int Cost;
+		public string Era { get; set; }
+		public int AdvanceIcon;
+		public int X;
```
esbudylin, line 17: What does X and Y fields represent?

C7GameData/Tech.cs
```diff
@@ -0,0 +1,65 @@
...
+	public class Tech {
+		public ID id;
+		public string Name { get; set; }
+		public string CivilopediaEntry { get; set; }
+		public int Cost;
+		public string Era { get; set; }
+		public int AdvanceIcon;
+		public int X;
+		public int Y;
+		public Tech Prerequisite1;
```
esbudylin: Maybe we should use a collection here to reduce code duplication and avoid hard-coding a maximum number of prerequisites.

TomWerner: Good call, done.

esbudylin (approved):

PR #520, opened 2025-01-25
Properly handle texture loading when `$CIV3_HOME` doesn't end in a slash

TomWerner:
Thanks to Yegor for catching this.


esbudylin (approved):
Can confirm the fix is working.

TomWerner:
Thanks!

PR #526, opened 2025-01-26
Import and track the science/luxury slider values

TomWerner:
I'll need this to determine how much gold goes to research.

#392

### DIFF
--- a/C7GameData/Player.cs
+++ b/C7GameData/Player.cs
@@ -37,6 +37,14 @@ namespace C7GameData {
 
 		public int turnsUntilPriorityReevaluation = 0;
 
+		// The values of the science/happiness/tax sliders (tax is implicit)
+		// A value of 1 => 10%, a value of 10 => 100%.
+		//
+		// INVARIANT: LuxuryRate + ScienceRate + TaxRate = 10
+		public int luxuryRate = 0;
+		public int scienceRate = 5;
+		public int taxRate = 5;
+
 		// The amount of gold this player has.
 		public int gold = 0;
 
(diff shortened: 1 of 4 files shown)


C7GameData/Player.cs
```diff
@@ -37,6 +37,14 @@ public class Player {
 
 		public int turnsUntilPriorityReevaluation = 0;
 
+		// The values of the science/happiness/tax sliders (tax is implicit)
+		// A value of 1 => 10%, a value of 10 => 100%.
+		//
+		// INVARIANT: LuxuryRate + ScienceRate + TaxRate = 10
+		public int luxuryRate = 0;
+		public int scienceRate = 50;
+		public int taxRate = 50;
```
esbudylin: Judging by the comment above, shouldn't these values be 5 and 5?

C7GameData/Save/SavePlayer.cs
```diff
@@ -28,6 +28,14 @@ public class SavePlayer {
 
 		public int turnsUntilPriorityReevaluation = 0;
 
+		// The values of the science/happiness/tax sliders (tax is implicit)
+		// A value of 1 => 10%, a value of 10 => 100%.
+		//
+		// INVARIANT: LuxuryRate + ScienceRate + TaxRate = 10
+		public int luxuryRate = 0;
+		public int scienceRate = 50;
+		public int taxRate = 50;
```
esbudylin: The same applies here.

C7GameData/Player.cs
TomWerner: Good catch, done.

C7GameData/Save/SavePlayer.cs
TomWerner: Done

PR #528, opened 2025-01-26
Track each player's beakers and turns researched

TomWerner:
#392

PR #527, opened 2025-01-26
Display the science/luxury rates and allow the player to change them 

TomWerner:
image

image

#392

PR #529, opened 2025-01-27
Award techs to the player when fully researched

TomWerner:
... and show progress in the UI.

This doesn't add a popup for the player to pick the tech, but as long as the player goes to the science advisor they can pick new techs.

Here's some screenshots showing I was able to research pottery and writing, and that I'm currently researching philosophy.

image

image

#392

### DIFF
--- a/C7/UIElements/GameStatus/LowerRightInfoBox.cs
+++ b/C7/UIElements/GameStatus/LowerRightInfoBox.cs
@@ -152,7 +152,7 @@ public partial class LowerRightInfoBox : TextureRect {
 	}
 
 	public void UpdateTechProgress(string techName, int turnsRemaining) {
-		if (turnsRemaining > 0) {
+		if (turnsRemaining >= int.MaxValue) {
 			SetTextAndCenterLabel(scienceProgress, $"{techName} (-- turns)");
 		} else {
 			SetTextAndCenterLabel(scienceProgress, $"{techName} ({turnsRemaining} turns)");
--- a/C7Engine/EntryPoints/TurnHandling.cs
+++ b/C7Engine/EntryPoints/TurnHandling.cs
@@ -54,6 +54,16 @@ namespace C7Engine {
 					// incremented if the player is actually spending money on
 					// research, or has a science specialist.
 					player.turnsResearched++;
+
+					// Check to see if the player has finished researching their
+					// tech, and if they have, add it to the list of known techs
+					if (player.currentlyResearchedTech != null) {
+						Tech tech = gameData.techs.Find(x => x.id == player.currentlyResearchedTech);
+						if (player.EstimateTurnsToResearch(tech) <= 0) {
+							player.knownTechs.Add(player.currentlyResearchedTech);
+							player.SetCurrentlyResearchedTech(null);
+						}
+					}
 				}
 				OnBeginTurn();
 			}
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -221,10 +221,14 @@ public partial class Game : Node2D {
 					// F6 is the science advisor.
 					// TODO: Move the F* key strings to a set of constants/enum.
 					EmitSignal(SignalName.ShowSpecificAdvisor, "F6");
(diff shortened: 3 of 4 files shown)


C7/UIElements/Advisors/Advisors.cs
```diff
@@ -44,13 +44,14 @@ private void OnShowSpecificAdvisor(string advisorType) {
 		}
 
 		if (advisorType.Equals("F1")) {
-			if (domesticAdvisor == null) {
-				domesticAdvisor = new DomesticAdvisor();
-				advisors.Add(domesticAdvisor);
-				AddChild(domesticAdvisor);
-			} else {
-				domesticAdvisor.Show();
+			if (scienceAdvisor != null) {
```
Sean-Brown: should this be `domesticAdvisor`?

C7/UIElements/GameStatus/LowerRightInfoBox.cs
```diff
@@ -152,7 +152,7 @@ public void SetTurnAndGold(int turnNumber, int gold, int goldPerTurn) {
 	}
 
 	public void UpdateTechProgress(string techName, int turnsRemaining) {
-		if (turnsRemaining > 0) {
+		if (turnsRemaining >= 1e9) {
```
Sean-Brown: is there some significance to `1e9` or is it just meant to be a big number? If just a big number, then maybe a constant like `Int32.MaxValue`

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -40,10 +41,27 @@ internal static void AdvanceTurn() {
 				UnitInteractions.ClearWaitQueue();
 
 				SpawnBarbarians(gameData);
+
+				// TODO: Should we be iterating over players, and then over each
```
Sean-Brown: For single player, the game behavior is to respect player order when iterating over cities. In multiplayer I'm unsure if it has to do with the game rules (e.g. simultaneous moves vs turn based, etc) or if it's the same as single player.  Perhaps that can be addressed in a separate ticket

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -40,10 +41,27 @@ internal static void AdvanceTurn() {
...
+				// player's city, instead of over all cities, irrespective of
+				// player order?
 				HandleCityResults(gameData);
 				gameData.UpdateTileOwners();
 
 				gameData.turn++;
+				foreach (Player player in gameData.players) {
+					player.turnsResearched++;
```
Sean-Brown, line 56: not always true but probably good enough for now. In the early game if you have 0% on the science slider, then this shouldn't increment. Later on it is possible to have research even with 0% on the science slider although I'm not sure what all contributes to that.

Sean-Brown (approved):

C7Engine/EntryPoints/TurnHandling.cs
N00byEdge: Researcher specialists provide science even with slider on 0%.

C7/UIElements/Advisors/Advisors.cs
TomWerner: Good catch, done. I fixed this in #527 (which was a diffbase of this PR)

C7Engine/EntryPoints/TurnHandling.cs
TomWerner: I've added a TODO for improving this behavior, thanks for the details here! The actual change was applied in #528 (a diffbase of this one).

C7Engine/EntryPoints/TurnHandling.cs
TomWerner: Thanks - I've added a link to this comment in the TODO, applied in #528.

C7/UIElements/GameStatus/LowerRightInfoBox.cs
TomWerner: Good call, done.

PR #530, opened 2025-01-27
chore: :memo: upgrade target .NET framework to 8.0

### LINKED ISSUE #483: Upgrade to .NET 8
.NET 8.0.11 is the current LTS release with support through 2026. This will also give us access to C# 12 features.

lt-brlara:
Closes #483 

This change modifies all .csproj to target .NET 8.0.

This change also upgrades to the latest Godot SDK version of 4.3.0.

### DIFF
--- a/.github/workflows/composite/buildSetup/action.yml
+++ b/.github/workflows/composite/buildSetup/action.yml
@@ -22,7 +22,7 @@ runs:
     if: github.event.inputs.include-datetime == 'true'
     run: |
       echo "BUILD_TIME=-$(date --iso-8601=minutes | sed -e 's/:/-/g' | sed -e 's/\+.*//')" >> $GITHUB_ENV
-  - name: Set .net7.0 SDK
+  - name: Set .net8.0 SDK
     shell: bash
     run: |
       wget https://packages.microsoft.com/config/debian/10/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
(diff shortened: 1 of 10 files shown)


WildWeazel:
There are a few more projects than this that target 7.0 but also looks like our CI environment needs an update

lt-brlara:
Apologies, malformed search string got the best of me, and I was deceived by the successful build log. Will address the remaining `.csproj` files.

pcen (approved):

WildWeazel:
Does anyone object to upgrading to Godot 4.3 at this time? That will mean updating the project file and CI image also. 4.4 should be just around the corner.

TomWerner:
> Does anyone object to upgrading to Godot 4.3 at this time? That will mean updating the project file and CI image also. 4.4 should be just around the corner.

@lt-brlara if it's not too much trouble could we save the Godot version change for a separate PR? Right now if we download the artifact from the CI runs it appears to be using `Godot Engine v4.2.2.stable.mono.official.15073afe3` based on the console output, so there's still some sort of mismatch there.

It might be nice to save the Godot version changing for a new PR, to make debugging any version-related issues easier.

Sorry that this PR has languished for so long!

WildWeazel:
I was leaning towards that too. Let's revert the engine change and have another ticket to import the project to 4.3 and check everything out.

PR #531, opened 2025-01-27
Retrieve capital city information from BIQ and SAV files

esbudylin:
This commit loads capital information from BIQ and SAV files, making it possible to mark the capital city in the game.

image
image

### DIFF
--- a/C7GameData/Save/SavePlayer.cs
+++ b/C7GameData/Save/SavePlayer.cs
@@ -1,5 +1,4 @@
 using System.Collections.Generic;
-using System.Text.Json.Serialization;
 
 namespace C7GameData.Save {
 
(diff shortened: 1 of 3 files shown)


C7GameData/Save/SavePlayer.cs
```diff
@@ -12,6 +13,13 @@ public class SavePlayer {
 		public string civilization;
 		public int cityNameIndex = 0;
 
+		// Index of the player's capital city.
```
TomWerner: Instead of having this behavior branch on whether things are loaded from a sav or bic file, why not always store the information in the SaveCity class?

esbudylin: This mimics the difference in the structure of BIQ and SAV files. SAV files store the information about a player's capital as part of player's data, while BIQ files store it as part of the city's data. So, when loading data from SAV we first need to load it as part of the player data and later record it into the SaveCity class. 

Maybe we should create a structure to store such temporary data inside ImportCiv3 instead of SavePlayer? I wonder if there are any other use cases for it.

TomWerner: Could it just go in `ImportBicCities()`? There's no rule that you can't access the player data in that function, and it would avoid this non-serialized field.

esbudylin: Thanks for the advice! For some reason it just didn't occur to me. Should be fixed now.

PR #532, opened 2025-01-29
Support a barebones "city screen" view

TomWerner:
This is triggered when building a city and when clicking "zoom to city"

For now I've just added the "exit" button, but hope to add tile yields soon.

I also fixed an issue with the advisor and input handling - previously if you exited the advisor via escape the game would ask if you wanted to exit. Now it works properly.

This will hopefully yield additional good starter PRs - there's lots of labels and icons to place on this screen.

image

### DIFF
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -420,21 +420,23 @@ namespace C7Engine {
 			return unit.location.neighbors.Values.All(tile => !tile.HasCity);
 		}
 
-		public static void buildCity(this MapUnit unit, string cityName) {
+		public static City? buildCity(this MapUnit unit, string cityName) {
 			if (!unit.canBuildCity()) {
 				log.Warning($"can't build city at {unit.location}");
-				return;
+				return null;
 			}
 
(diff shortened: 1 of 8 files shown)


C7Engine/MapUnitExtensions.cs
```diff
@@ -420,21 +420,23 @@ public static bool canBuildCity(this MapUnit unit) {
 			return unit.location.neighbors.Values.All(tile => !tile.HasCity);
 		}
 
-		public static void buildCity(this MapUnit unit, string cityName) {
+		public static City buildCity(this MapUnit unit, string cityName) {
 			if (!unit.canBuildCity()) {
 				log.Warning($"can't build city at {unit.location}");
-				return;
+				return null;
```
esbudylin, line 426: Perhaps we could return a nullable type from this method to enhance type safety.

esbudylin (approved):

C7Engine/MapUnitExtensions.cs
TomWerner: Done

TomWerner:
This is related to #541

PR #533, opened 2025-01-30
 Fix ID collisions on save and scenario loading

esbudylin:
This PR fixes ID collisions when loading existing game files by introducing a new constructor to the ID Factory class. The constructor accepts a SaveGame instance and tracks the highest ID count per ID type.

Currently, this is implemented only for units and cities, as they appear to be the only entity types created during gameplay.  In contrast, civilizations and technologies are assigned IDs only at the start of the game, so tracking them when loading a save file is unnecessary.

The PR includes unit tests for the ID Factory.

### DIFF
--- a/C7GameData/ID.cs
+++ b/C7GameData/ID.cs
@@ -1,9 +1,16 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using C7GameData.Save;
 
 namespace C7GameData {
+	public interface IHasID {
+		ID id { get; }
+	}
+
 	public class ID {
 		// at runtime, any ID parsed from a string as "<key>-none" will be
 		// compared to other IDs by -0xC7C7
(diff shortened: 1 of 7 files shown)


C7GameData/ID.cs
```diff
@@ -60,33 +67,45 @@ public override bool Equals(object obj) {
...
+			// When creating an ID Factory for a loaded save game, 
+			// this constructor finds the largest n-value for each ID key.
+			// ie. loading a save with units ["warrior-1", "warrior-3", "worker-2"]
+			//     should result in an id factory with
+			// keyCounter = {
+			//     "warrior": 4,
+			//     "worker": 3,
+			// }
+			public Factory(SaveGame save) {
+				IEnumerable<IHasID> entities = new List<IEnumerable<IHasID>>() {save.Units, save.Cities}
```
pcen, line 87: I understand the motivation of only doing this for units and cities, but I suspect it may be simpler and less error prone to have a way to get the ID factory used to create the SaveGame, and copy the internal state from that factory instead of iterating over the SaveGame contents themselves

esbudylin: Can you elaborate? I don't see how it would be viable for loading a JSON save that already has a set of predefined IDs.

TomWerner (approved):
This seems like a reasonable approach to fixing this problem, thank you for adding tests!

I'm going to go ahead and merge this, if pcen@ has followup thoughts they can always be addressed in future PRs.

PR #535, opened 2025-01-31
Fix border not disappearing after capturing a city

esbudylin:
A minor fix that adds border recalculation on city capturing. The logic for recalculating borders can definitely be made cleaner (mainly to avoid recalculating all borders after a capture), but I see no sense in doing so until we have proper culture/territory calculations.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -197,6 +197,8 @@ public partial class Game : Node2D {
 							MapUnitExtensions.disband(mCD.city.owner.units[i]);
 						}
 					}
+
+					gameData.UpdateTileOwnersOnCityDestruction(mCD.city);
 					break;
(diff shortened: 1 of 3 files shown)


esbudylin:
Here is a save I used to test it. It makes American settler spawn near the player and gives the player a warrior.

[c7-static-map-save.json](https://github.com/user-attachments/files/18624346/c7-static-map-save.json)

TomWerner (changes requested):
I'm not sure this game data interaction should go in `Game.cs` - could it go in `CityInteractions.cs` in the `DestroyCity` method, to pair with the similar logic in `BuildCity()`?

esbudylin:
@TomWerner good catch! Should be fixed now.

TomWerner (approved):

PR #537, opened 2025-02-02
Improve handling of variables and relative paths when reading paths from env vars

### LINKED ISSUE #460: Handle variables in file paths
It turns out that `$CIV3_HOME` is not evaluated correctly when using `~`. Ensure that we have a consistent method of handling any symbols, variables, relative paths, etc across the code base and update at least all non-hardcoded file path usage to comply. .NET surely has utility functions for this.

David1Socha:
Should resolve https://github.com/C7-Game/Prototype/issues/460 . 

Some notes on the approach taken: 

Dotnet has [Environment.ExpandEnvironmentVariables](https://learn.microsoft.com/en-us/dotnet/api/system.environment.expandenvironmentvariables?view=net-9.0) which seems like it'd be suitable for expanding any nested environment variables, but [by design](https://github.com/dotnet/runtime/issues/25792) that method only supports Windows-style `%VARIABLES%`. I could not find any ready-made .NET functions or util libraries that fully addressed the concerns flagged in issue 460, so I added some ad-hoc checks and updates to the `GetCiv3Path` method in `Civ3Location.cs`.

In `Util.cs`, I noticed that its `GetCiv3Path` method seems to be unused at the moment, with the `Civ3Root` property forwarding out to `Civ3Location.GetCiv3Path`. I thought it made sense to consolidate the similar `GetCiv3Path` methods in `Util.cs` and `Civ3Location.cs` to a single method, so I updated `Util.GetCiv3Path` to point to `Civ3Location.GetCiv3Path` (but first checking C7Settings for a saved civ3 location, since that seemed like a significant difference between the two methods). I'd be happy to either fully remove `Util.GetCiv3Path` or restore its original method body and then also add my changes to that method if either of those would be preferred. 

I updated the `GetHome` function from using `Environment.SpecialFolder.Personal` to using `Environment.SpecialFolder.UserProfile` , this is because .NET 8 (which I see is on the roadmap) has a breaking change to `SpecialFolder.Personal` changing behavior for Linux systems from returning `$HOME` to returning "`XDG_DOCUMENTS_DIR` if available; otherwise `$HOME/Documents`". `Environment.SpecialFolder.UserProfile` returns $HOME under both .NET 7 and .NET 8.

Tested relative directory handling and environment variable replacement on a local Windows game build with these changes, both worked successfully. I tested the logic for relative directory handling, Unix-style environment variable replacement, and `~` home path replacement in a Console app in WSL and that also worked as expected. Haven't tested the full game build with these changes on Linux or OSX because I don't currently have setups for those operating systems, but if that's needed before merge I can setup a Linux VM and try to validate there as well.

### DIFF
--- a/C7/Util.cs
+++ b/C7/Util.cs
@@ -1,57 +1,20 @@
 using System;
-using System.Linq;
 using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.InteropServices;
 using C7Engine;
-using Godot;
 using ConvertCiv3Media;
-using C7GameData;
+using Godot;
 using QueryCiv3;
 
 public partial class Util {
 	static public string Civ3Root = Civ3Location.GetCiv3Path();
 
-	static private string SteamCommonDir() {
-		string home = System.Environment.GetFolderPath(System.Environment.SpecialFolder.Personal);
-		if (home == null) { return null; }
-		return System.IO.Path.Combine(home, "Library/Application Support/Steam/steamapps/common");
-	}
-
-	static private bool FolderIsCiv3(System.IO.DirectoryInfo di) {
-		return di.EnumerateFiles().Any(f => f.Name == "civ3id.mb");
-	}
-
 	static public string GetCiv3Path() {
-		// Use CIV3_HOME env var if present
-		string path = System.Environment.GetEnvironmentVariable("CIV3_HOME");
-		if (path != null) { return path; }
-
-		// Use settings value if present
-		path = C7Settings.GetSettingValue("locations", "civ3InstallDir");
+		string path = C7Settings.GetSettingValue("locations", "civ3InstallDir");
 		if (path != null) return path;
 
-		if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
-			// Look up in Windows registry if present
-			path = Civ3PathFromRegistry();
-			if (path != null) { return path; }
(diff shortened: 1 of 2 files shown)


C7/Util.cs
```diff
@@ -1,57 +1,20 @@
...
 using System.Runtime.InteropServices;
 using C7Engine;
-using Godot;
 using ConvertCiv3Media;
-using C7GameData;
+using Godot;
 using QueryCiv3;
 
 public partial class Util {
 	static public string Civ3Root = Civ3Location.GetCiv3Path();
```
David1Socha, line 11: Not strictly relevant to this PR, but I was wondering if this property is truly meant to use `Civ3Location.GetCiv3Path()` and not the class method `Util.GetCiv3Path()`. Currently `Util.GetCiv3Path` has no refs, and the `C7Settings.GetSettingValue` check seems like something that might be expected when calling `Util.Civ3Root`. From git history it looks like this property was updated [2 years ago](https://github.com/C7-Game/Prototype/commit/fe80243324ba2fd59b65f13fe5efcb2335cf9988#diff-bff3453a67ebda4628f6ee324889d513104e16548be5c36d952595c4f02a5f22R12), but I wanted to double check if this setup was intentional

TomWerner: I can't say one way or another if it was intentional, but if it doesn't seem needed it might be worth a separate PR to change it.

TomWerner (approved):
Thanks! I confirmed this worked correctly on Linux when using a `CIV3_HOME` with `~` in it.

WildWeazel:
@David1Socha congrats on your first(?) PR and thanks for this fix! How would you prefer to be named in the game credits?

PR #538, opened 2025-02-04
Remove GodotSink and remaining GD.Print calls

### LINKED ISSUE #258: Migrate GD.Print and Console.Write to Serilog
Replace all (? I can't really think of a reason not to, other than to print an error from the logger) GD.Print and Console.Write* statements with logs as demonstrated in MainMenu.cs. Use appropriate log levels and try to be consistent with what we log and how it's phrased. Also remove the redundant "Hello/Goodbye logger" prints in LogManager now that logging is working.
Note that since LogManager is now a Node, **calls to ForContext should be made within Node callbacks** even though it is static, to avoid calling before LogManager is ready.

Requires #267

bmgxyz:
It seems like this is already mostly done. There were only two `GD.Print`s, and all of the `Console.WriteLine`s are in places where they should be (`BuildDevSave` and `C7GameDataTests`). It also seems like `GodotSink` isn't needed anymore, so I removed it.

Closes #258


WildWeazel:
It's been a long time since I looked at logging, but how did you conclude that GodotSink is unused? I thought it was providing some kind of global service.

bmgxyz:
>how did you conclude that GodotSink is unused?

My reasoning was that the `GodotSink` identifier doesn't appear anywhere except in its own file and [one comment](https://github.com/C7-Game/Prototype/blob/077be117d1370b43c18ce9a98066de277627ddc4/C7/LogManager.cs#L24) in `LogManager.cs`. But now that I read the comment more closely, it seems like it's supposed to be there and enabled when needed (?).

I guess the question here is whether we still need `GodotSink` to serve [its original purpose](https://github.com/C7-Game/Prototype/pull/255#issue-1238054970), which was to pipe Serilog messages to the Godot editor. If we do, then maybe it should be a configuration option somewhere instead of a comment. If we don't, then we should probably remove it.

TomWerner (approved):

PR #539, opened 2025-02-05
Add settler tile adjustments to Civilization class

### LINKED ISSUE #92: Settler AI - Refactor so weights are not constants
Right now, the weights given by the AI to where to settle are hard-coded - food yield, shields, commerce, and distance, as well as bonuses based on hills and coastline.  But it would be nice to make this configurable and available to mods.

I'd suggest making them configurable, but with default values that will be used in lieu of any being set.  There should also be some sort of a setup for AIWeights, which will be mod-configurable (in the JSON file?).

Longer-term, I'd like for it to be possible for different AIs to have different weights; maybe a Commercial AI puts extra weight on commerce yield, and an Industrious one on industrial output.  Whether this is changing the direct weight (+1 for shield output for industrious civs), or a percent bonus (+20% weight to shield yield) is to be determined.  Optionally, think about and add something to make it variable by civ.

bmgxyz:
- Add `SettlerTileAdjustments` as a child of `Civilization`
  - Encodes the previously hard-coded values as defaults for public fields
  - ~~Uses lambda functions to take values at runtime for adjustments that take arguments; this allows future users to supply arbitrary logic~~
- Replace hard-coded bonus values in `SettlerLocationAI.cs` with ~~calls and~~ references to `SettlerTileAdjustments` via the supplied `Player`
  - Notably, `DistancePenalty` is now a negative value by default, which I think is more clear because now adjustments are always added to the tile score
- Make settler tile score a `float` instead of `int`, which gives more precision for multiplicative adjustments
- ~~Apply lots of lints from Rider~~
  - ~~Identifier case tweaks (this caused most of the changes in other files)~~
  - ~~Using ternary operators to make code a little smaller when it's still clear~~
  - ~~Some grammar problems (Rider is picky about these for some reason)~~
  - ~~Lint ignores in cases where it makes sense (to me)~~
  - ~~I'm willing to undo these if they're too aggressive or unwanted~~

Closes #92


TomWerner (changes requested):
Thanks for the PR, and the detailed description!

Would it be possible to split these apart into a couple of PRs? In particular, I think a PR that only fixes linting issues would be ideal. The linting issues touch so many files that its hard to see the actually interesting changes, so reducing diff noise would make it much easier to review the PR confidently.

Also it looks like `C7/C7.sln.DotSettings` is a new file, was that intended to be included in this PR?

bmgxyz:
> Would it be possible to split these apart into a couple of PRs? In particular, I think a PR that only fixes linting issues would be ideal.

Makes sense to me. I'll pull out the lint changes into a new PR and then go through the (many) other ones on that separate branch. It will take some time but will make for a cleaner development experience because the UI will quiet down. Only the parts that are directly related to #92 will remain here.

> Also it looks like `C7/C7.sln.DotSettings` is a new file, was that intended to be included in this PR?

This is part of the linting changes. Rider applies [spelling and grammar linting](https://www.jetbrains.com/help/rider/Proofreading.html) in comments :roll_eyes: and some of the words that appear in comments aren't in its dictionary (e.g., "civilopedia"). It's possible to add the unrecognized words to a dictionary so Rider stops flagging them, either locally (untracked) or "team-shared" (tracked). The file you mentioned is where Rider stores this configuration. I *think* tracking this file is a good idea because it *should* prevent this kind of linter noise in all Rider installations for this project.

A better option would probably be to find a way to disable this kind of linting entirely, but that should be a discussion elsewhere. I hadn't used Rider before this project, so I'm sure there's a lot I don't know here.

----------

Do you have any thoughts about my approach related to #92? I'm still unfamiliar with the project internals and C# in general, so it wouldn't surprise me if it can be improved.

TomWerner:
> Do you have any thoughts about my approach related to https://github.com/C7-Game/Prototype/issues/92? I'm still unfamiliar with the project internals and C# in general, so it wouldn't surprise me if it can be improved.

If you're talking about the changes to `Civilization.cs` and `SettlerLocationAI.cs`, yeah, that seems good. Do you happen to know if the tests pass successfully? I'm curious as to how the new class of constants is handled with JSON serialization.

bmgxyz:
Linting changes removed. This PR now only contains the changes related to the settler AI. Ready for another pass @TomWerner 

I'm not sure whether I'm running the tests right, but on my system there are 12 of 46 failing on this branch, compared to 8 failing on `Development`. I'll look into what's broken. I also don't know of a good way to actually test the changes I've made, so if anyone has any insight on that please let me know.

bmgxyz:
The failing tests on `Development` were due to my not having set `CIV3_HOME`. Now all tests are passing for me on that branch. Six tests are still failing on this PR, and they do seem to be related to JSON serialization. I'll look into those more.

bmgxyz:
All tests pass locally. The problem was my use of `Func` to encode the yield bonuses. `Func` can't be serialized easily, so I replaced them with `float`s and inlined the multiplications. This reduces flexibility compared to an arbitrary lambda function, but it's not clear that we even needed that flexibility.

It might be cool to find a way to encode the difference between multipliers and constant bonuses in the type system. Right now everything is a `float`, which makes it easier to write logic bugs.

I also figured out that I needed to update the test saves for the tests to pass, so I overwrote them with the outputs of the failed runs. This resulted in a sensible diff that (I think) only adds a default block of settler location AI constants to each civ. If there's a better/accepted way to update these files in this situation, let me know.

Still not sure how to robustly test that the settler location AI logic is actually working right, but maybe that doesn't matter for this PR.

C7Engine/C7GameData/Civilization.cs
```diff
@@ -49,6 +50,24 @@ public Civilization(string name) {
...
+
+			// multipliers
+			public float CommerceYieldBonus = 2;
+			public float DistancePenalty = -2;
+			public float FoodYieldBonus = 5;
+			public float ProductionYieldBonus = 3;
+
+			// constants
+			public float HillsBonus = 10;
+			public float LuxuryResourceBonus = 15;
```
stavrosfa, line 67: Could you add the notes that are above the constants in the original SettlerLocationAI regarding weights as a TODO here so we won't miss/forget it in the future?

C7Engine/AI/UnitAI/SettlerLocationAI.cs
```diff
@@ -1,24 +1,18 @@
-using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using C7GameData;
 using System.Linq;
 using C7GameData.AIData;
 using Serilog;
+// ReSharper disable ForeachCanBeConvertedToQueryUsingAnotherGetEnumerator
```
stavrosfa: Do we need these comments here?

stavrosfa (commented):
Just some small stuff before Tom has a chance to review. I would wait for his review too, so you don't have to make multiple commits in case he wants some changes too

stavrosfa:
> Still not sure how to robustly test that the settler location AI logic is actually working right, but maybe that doesn't matter for this PR.

You could have a look at [EngineTests/AI/Pathing/EdgeWalkerTest.cs](https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/EngineTests/AI/Pathing/EdgeWalkerTest.cs) to see how you can construct a tiny map of sorts and see if the output of findSettlerLocation() is what you would expect, for example, a hill would win over a plains tile, or a hill next to the sea would win over a hilly tile with no sea neighbours, stuff like that I suppose

C7Engine/AI/UnitAI/SettlerLocationAI.cs
bmgxyz: Nope, good catch. These are left over from when I was trying to fix a lot of the lints that Rider was complaining about.

bmgxyz:
Added some tests

EngineTests/AI/UnitAI/SettlerLocationTest.cs
```diff
@@ -0,0 +1,137 @@
...
+using C7GameData;
+using C7GameData.AIData;
+using C7GameData.Save;
+using EngineTests.Utils;
+using System.Collections.Generic;
+using Xunit;
+
+namespace EngineTests.AI.UnitAI {
+	public sealed class SettlerLocationAITests : MapBase {
+		private Player MakeTestPlayer(List<Tile> knownTiles) {
```
stavrosfa: This method and SurroundTile() look like usefull methods that could be used in other places in the future, wouldn't it be better to move them in MapBase?

EngineTests/AI/UnitAI/SettlerLocationTest.cs
```diff
@@ -0,0 +1,137 @@
...
+		private Tile MakeFloodPlainTileWithDefaultYield() {
+			Tile tile = MakeFloodPlainTile();
+			tile.overlayTerrainType.baseFoodProduction = 3;
+			tile.overlayTerrainType.baseShieldProduction = 0;
+			tile.overlayTerrainType.baseCommerceProduction = 0;
+			return tile;
+		}
+
+		[Fact]
+		private void HillsOverPlains() {
```
stavrosfa, line 27: I was very confused after seeing the hills and plains tiles on the same location.
I was so sure that if I did
`Assert.True(hills.overlayTerrainType.isHilly());`
It would fail because it looks like it gets overriden by the plains tile later. But it passes, and I realized the reason is that the map contains all of the 18 tiles that are created with each pair being on the same location, because it's a list of tiles, not a hashset of some sort.
It works, and it should, given the API allows it, it's just confusing.
Perhaps put the tiles equally away from the Player, without having duplicate locations (at least the start tiles)? What do you think?

stavrosfa (commented):
Thanks for this!

EngineTests/AI/UnitAI/SettlerLocationTest.cs
bmgxyz: I was wondering if that would trip anyone up. It's a weird consequence of the design that this is possible, and I wonder if in the future we might try to eliminate it. My intention with the overlapping tiles in these tests was to be sure that distance wasn't contributing to the settler's decision at all. But I agree that it's not especially readable. I've separated the test locations and put the settler at the midpoint between them in the relevant tests. In my debugging I also found that there was a redundant distance penalty being applied, so I removed it.

stavrosfa: Yeah it did trip me up good haha
I guess in the game it doesn't matter because the map is generated once and that's that, I don't think tiles are getting added or removed at any point.

> But I agree that it's not especially readable

Maybe a comment would have been enough for me not to go looking for ghosts in the code, although since I didn't know it ws permitted, I probably would have anyway

>there was a redundant distance penalty being applied, so I removed it

I don't know what's the logic behind it tbh, Tom is probably gonna have a look at it when he does his review

Everything else looks good to me :)

TomWerner (commented):
Sorry for the delay, seems reasonable to me but I didn't have time to look in super close detail.

Seems like [stavrosfa](https://github.com/stavrosfa) has been doing a nicely thorough job - if they're happy with it, it sgtm

stavrosfa:
@TomWerner (sorry for the tags), does this warrant any further review or should it get merged?

TomWerner:
Looks like there’s a merge conflict to deal with but besides that feel free to merge!

PR #540, opened 2025-02-09
CityResident: Track the owning city

TomWerner:
Knowing the city makes it easier in a following PR to display the worked tiles for a specific city.

#541


esbudylin:
I think the "nationality" field in the CityResident class reflects the citizen's culture rather than the city's owner. For example, if Sumeria captures an English city, its residents remain English despite the change in ownership. So, we should definitely keep this field, even if we don't use it yet.

TomWerner:
Oh, very good point. Updated to just add the new city field, instead of replacing the nationality field.

esbudylin (approved):

Issue #541, opened 2025-02-09
Implement enough of the city screen to show and modify tile assignments


TomWerner:
This is a pretty core part of the early game, where making worker improvements and assigning citizens to those worked tiles is key to maximizing production.

TomWerner:
I'm going to close this now:
 - we can assign tiles
 - we can change citizens to specialists (though the specialists don't do anything yet)
 - we can change production
 - we can view food/commerce/shield levels

PR #542, opened 2025-02-09
Display tile yields when zooming to a city

TomWerner:
This allows us to see what tiles are being worked, and also makes it clearer that we are currently giving the wrong yield for city centers.

In order for this to render properly I had to fix some issues with how the map view layers worked. Previously all the layers we part of the same `Node2D`, and most things were draw as textures. However, cities and units are child nodes, and child nodes are drawn on top of textures. This usually isn't a problem, but it meant that drawing the city center tile yield was hidden by the city itself. This also is the reason health bars were partially hidden by the city - the health bar is a texture, the city is a node.

To fix this I grouped the layers into separate nodes, so that within each node the layer ordering was simple, and then the node ordering took effect after that. As a result health bars now render properly, and we can see the city center yield.

This also shows that the city center yield is wrong! Next steps will include fixing that tile yield, and implementing support for changing tile assignments.

Some screenshots:

Right after creating the city:

image

After mining+roading a tile:

image

After irrigating+roading a tile:

image

#541

### DIFF
--- a/C7/UIElements/CityScreen/CityScreen.cs
+++ b/C7/UIElements/CityScreen/CityScreen.cs
@@ -3,12 +3,14 @@ using System;
 using Serilog;
 using System.Collections.Generic;
 using C7GameData;
+using C7.Map;
 
 
 // Handles the city screen, where citizens can be assigned and other details of
 // the city can bee seen.
 public partial class CityScreen : CenterContainer {
 	private ILogger log = LogManager.ForContext<CityScreen>();
+	public TileAssignmentLayer tileAssignmentLayer;
 
 	// Called when the node enters the scene tree for the first time.
 	public override void _Ready() {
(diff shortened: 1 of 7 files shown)


C7/UIElements/CityScreen/CityScreen.cs
```diff
@@ -23,13 +25,19 @@ public override void _Ready() {
...
 		};
 		close.SetPosition(new Vector2(950, 20));
-		close.Pressed += () => { this.Hide(); };
+		close.Pressed += () => { HideScreen(); };
 		background.AddChild(close);
 
 		this.Hide();
 	}
 
+	public void HideScreen() {
```
esbudylin, line 34: Can this method be an override of the parent's Hide?

esbudylin (approved):

C7/UIElements/CityScreen/CityScreen.cs
TomWerner: I don't think so - `Hide` isn't virtual, so we can't override it.

PR #546, opened 2025-02-10
Import the tile assignments of citizens from save files

TomWerner:
This doesn't yet handle resistors or specialists, and doesn't handle scenarios.

The "spiral" logic was tested by creating a bunch of sav files with one city and one population, and then positioning that citizen in each of the spots. I then printed the tile worked variable in the `SaveTest` for each of them to confirm the ordering.

#541

Example screenshot from a random save:

image


KulkoBSW:
Reviewed the code and it looks good to me.

PR #547, opened 2025-02-10
Fix Error on Windows when CIV3_HOME was not set.

KulkoBSW:
I move the check for null one step up in the chain and now it works, but should also work for Linux. 

The original problem was that Environment.ExpandEnvironmentVariables(path); throws a nullpointerexception when fed a null value.

### DIFF
--- a/QueryCiv3/Civ3Location.cs
+++ b/QueryCiv3/Civ3Location.cs
@@ -31,6 +31,7 @@ namespace QueryCiv3 {
 		}
 
 		private static string GetExpandedPath(string path) {
+			// if (path == null) return null;
 			bool isUnixLike = !RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 			if (isUnixLike && path[0] == '~') path = GetHome() + path.Substring(1);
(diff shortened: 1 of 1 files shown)


QueryCiv3/Civ3Location.cs
```diff
@@ -31,6 +31,7 @@ private static string ConvertUnixVarsToWindowsVars(string input) {
 		}
 
 		private static string GetExpandedPath(string path) {
+			// if (path == null) return null;
```
TomWerner: Could you remove this commented out line?

TomWerner (approved):
I confirmed this still works on Linux with `~` in the path - thanks!

I can merge this once the comment is resolved.

KulkoBSW:
@TomWerner Comment is resolved.

TomWerner:
Thanks!

WildWeazel:
@KulkoBSW congrats on your first(?) PR! How would you prefer to be named in the game credits?

KulkoBSW:
Kulko is fine (without the BSW), unless you guys all choose real names.

PR #548, opened 2025-02-10
Import citizen types from the BIC file

TomWerner:
This allows us to know about things like entertainers, tax collectors, etc.

While here I noticed that one of the tests wasn't actually doing anything, so I fixed the path issue and added an assertion to catch similar issues in the future.

#541

### DIFF
--- a/C7/Text/c7-static-map-save.json
+++ b/C7/Text/c7-static-map-save.json
@@ -64533,5 +64533,87 @@
         "tech-44"
       ]
     }
+  ],
+  "citizenTypes": [
+    {
+      "id": "CitizenType-1",
+      "isDefaultCitizen": true,
+      "specialistIndex": 0,
+      "singularName": "Laborer",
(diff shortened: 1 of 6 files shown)


C7/Text/c7-static-map-save.json
KulkoBSW, line 1: The civilopedia entries for the policeman and the engineer seem of. or is this because the correct entries are not yet there?

KulkoBSW (commented):
See comments on Citizentypes.

C7/Text/c7-static-map-save.json
TomWerner: Amusingly they're actually wrong in the base game - you can see that if you open the civ3 editor.

TomWerner: So I decided to just import them as-is, we can always fix them up later if we actually write civilopedia entries for them.

PR #549, opened 2025-02-12
Add initial support for rendering citizens and specialists 

TomWerner:
This doesn't yet handle citizen moods (everyone is rendered as content), doesn't handle citizen gender, and doesn't handle the "region" of the civ. But it does allow us to see specialists, which will be important for tile assignment - we need to be able to turn a citizen into an entertainer to support some types of tile reassignment.

#541

Real game screenshot: 
image

C7 screenshot of the same save game: 
image


KulkoBSW (commented):
Looks good to me.

PR #550, opened 2025-02-12
Support turning citizens into specialists and changing their type 

TomWerner:
This doesn't yet implement the benefits of specialists, but it does allow us to force certain tiles to not be worked. This is important when passing tiles between cities, for example.

#541

Start: 
image

After turning everyone into entertainers: 
image

And after using a mix of specialists: 
image

### DIFF
--- a/C7/UIElements/CityScreen/CityScreen.cs
+++ b/C7/UIElements/CityScreen/CityScreen.cs
@@ -14,6 +14,7 @@ public partial class CityScreen : CenterContainer {
 	private ILogger log = LogManager.ForContext<CityScreen>();
 	public TileAssignmentLayer tileAssignmentLayer;
 	public MapView mapView;
+	public List<CitizenType> citizenTypes;
 	private TextureRect background;
 	private List<TextureButton> popHeads = new();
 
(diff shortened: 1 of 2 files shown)


C7/UIElements/CityScreen/CityScreen.cs
```diff
@@ -73,9 +81,15 @@ private void HandleReassignment(Tile tile, List<CitizenType> citizenTypes) {
 		}
 
 		// If we're already working a tile and click on it, turn the citizen
-		// into an entertainer.
+		// into a specialist.
 		if (tile.personWorkingTile != null && tile.personWorkingTile.city == city) {
-			// TODO: implement entertainers.
+			foreach (CitizenType citizenType in GetKnownSpecialists(city.owner)) {
```
KulkoBSW: What is the meaning of the For Each here?
I understand this happens when you click on a worked tile, so it shoudl turn the person into the first specialist type available? So why is there a need to cycle through all known types. why not just selecting the first entry?

TomWerner: Good question, I'm not sure what I was thinking! Switched to just the first entry.

TomWerner:
The one comment has been addressed and this PR has been open for a week, I'm going to merge it. Feel free to add post-submit comments, I'll happily address them.

PR #551, opened 2025-02-12
 Provide tile assignments for cities in scenarios

TomWerner:
Unlike with `.SAV` files, scenarios don't specify the tile assignments for each of the citizens in cities. To handle this we backfill them using the tile assignment AI. This is slightly complicated by fact that the `C7Engine` code can't depend directly on the AI logic, so we have to do some dependency injection via an `Action`.

#541


TomWerner:
Merging as this PR has been open a week. I'm always happy to address post-submit comments.

PR #552, opened 2025-02-12
Import the proper borders from save files

TomWerner:
It doesn't seem like scenarios store this information (it's unclear what the `Border` field is in the `TILE` struct, but it doesn't index into players or civs, and it seems like borders are calculated based on culture in the editor).

#288

Here's a random save I loaded:

image


TomWerner:
I realized this isn't the right way to do this, I'll drop the PR.

PR #553, opened 2025-02-12
ImportCiv3: Properly handle the "entire map visible" scenario option

TomWerner:
This is relevant for WWII, which now shows the entire map, as expected.

Zoomed out screenshot: 
image


KulkoBSW (commented):
Looks Good to me

PR #555, opened 2025-02-14
Basic multiturn irrigation

KulkoBSW:
@TomWerner 
I have created a first version and am looking for a bit of feedback before diving deeper in,

What it does for now:
- Make irrigation take 4 turns
- Set the worker busy while working
- - Multiple workers working on the same project

What it does not for now:
- animations
- use values from Terraform (instead every number is hardcoded variable)
- Difference between workers and slaves.
- Different update rules for Multiplayer
- Saving and loading

What I am not sure about:
I update the progress on the end of the current turn (Which seems to fit with the functionality described in the doc). But you handle other similar things (healing) in OnBeginTurn. 
How would I even begin to make a different behavior Multiplayer (update each Worker individually) vs. Singleplayer (Update all workers at one point and then check if the Job is done)


C7GameData/Tile.cs
```diff
@@ -251,6 +251,7 @@ public static Tuple<int, int> NeighborCoordinate(int X, int Y, TileDirection dir
 			}
 			return Tuple.Create(X, Y);
 		}
+
```
TomWerner: could you revert this extra line?

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -69,6 +70,7 @@ internal static void AdvanceTurn() {
 			}
 		}
 
+
```
TomWerner: could you revert this extra line?

C7Engine/TileExtensions.cs
```diff
@@ -40,6 +41,91 @@ public static void DisbandNonDefendingUnits(this Tile tile) {
...
+			int totalProgress = 0;
+			foreach (MapUnit unit in tile.unitsOnTile) {
+				if (unit.WorkerJob == null) {
+					continue;
+				}
+
+				if (currentWorkerJob.Equals(unit.WorkerJob)) {
+					totalProgress += unit.WorkerProgressTowardsJob;
+				} else {
+					// reset Unit working on other jobs
```
TomWerner, line 61: This needs to be reset if the worker is selected/moved while doing the work - otherwise you can start irrigating on a legal tile, then move to a different tile and let the irrigation finish.

Here I managed to irrigate a hill

image

C7Engine/TileExtensions.cs
```diff
@@ -40,6 +41,91 @@ public static void DisbandNonDefendingUnits(this Tile tile) {
...
+					unit.WorkerJob = null;
+					unit.WorkerProgressTowardsJob = 0;
+				}
+			}
+			return totalProgress;
+		}
+		/// <summary>
```
TomWerner, line 68: could you add a blank line between methods?

C7Engine/TileExtensions.cs
```diff
@@ -40,6 +41,93 @@ public static void DisbandNonDefendingUnits(this Tile tile) {
...
+		/// After a new WorkerJob has started, checks for workers working on different jobs and resets them
+		/// </summary>
+		/// <param name="tile">the current tile</param>
+		/// <param name="currentWorkerJob">the worker job currently started</param>
+		public static void UpdateAllWorkerJobs(this Tile tile, string currentWorkerJob) {
+			if (currentWorkerJob == null) {
+				Log.Error($"can't call method FinishWorkerJob without WorkerJob");
+				return;
+			}
```
TomWerner: I'd leave out this null check - there's already a null check in the caller.

C7Engine/TileExtensions.cs
```diff
@@ -40,6 +41,93 @@ public static void DisbandNonDefendingUnits(this Tile tile) {
...
+			int totalProgress = 0;
+			foreach (MapUnit unit in tile.unitsOnTile) {
+				if (unit.WorkerJob == null) {
+					continue;
+				}
+
+				if (currentWorkerJob.Equals(unit.WorkerJob)) {
```
TomWerner: Is there a reason to use `.Equals` instead of `==`? (Genuine question, I'm not super well versed in the nuances of C#)

C7Engine/TileExtensions.cs
```diff
@@ -40,6 +41,93 @@ public static void DisbandNonDefendingUnits(this Tile tile) {
...
+		/// <summary>
+		/// After a WorkerJob has finished, Clean up all the WorkerJobs and set the correct overlay
+		/// </summary>
+		/// <param name="tile">the current tile</param>
+		/// <param name="currentWorkerJob">the worker job currently finished</param>
+		public static void FinishWorkerJob(this Tile tile, string currentWorkerJob) {
+			if (currentWorkerJob == null) {
+				Log.Error($"can't call method FinishWorkerJob without WorkerJob");
+				return;
+			}
```
TomWerner: Similarly I'm not sure this null check is necessary

C7Engine/MapUnitExtensions.cs
```diff
@@ -527,8 +533,21 @@ public static void irrigate(this MapUnit unit) {
 			}
 
 			// TODO add animation and long process of building
-			unit.location.overlays.irrigation = true;
+			unit.WorkerJob = C7Action.UnitIrrigate;
+			unit.updateWorkerJob();
```
TomWerner: To double check my understanding here:
 - This won't cause double counting because the end of turn logic only looks at units with movement points left, and this consumes the movement points?
 - The check below ensures that if the cost was 1, we finish right away?

If both are true, it might be worth leaving a comment explaining some of the subtlety here.

TomWerner:
> How would I even begin to make a different behavior Multiplayer (update each Worker individually) vs. Singleplayer (Update all workers at one point and then check if the Job is done)

I wouldn't worry too much about multiplayer yet, just single player to start seems very reasonable. 

---

Overall I think this looks good! 

One other code pointer, I think we can set the button `TooltipText` field [here](https://github.com/C7-Game/Prototype/blob/Development/C7/UIElements/UnitButtons/UnitButtons.cs#L100) which would be a nice eventual improvement.

TomWerner (changes requested):

C7Engine/TileExtensions.cs
KulkoBSW: Okay I checked, you are right these are just my habits from other programming languages.

C7Engine/TileExtensions.cs
KulkoBSW: The method is designed to be reusable, which is why I added the line. Right now a nullpointer would be thrown when I compare with a nullvalue, but that won't be the case anymore once I use ==. But still if somebody would call the function with null it would behave rather unintuitive and might create hard to find bugs which is why I thought it better to add the line. If we are concerned Efficiency/Performance I can also take it back out.

C7Engine/TileExtensions.cs
KulkoBSW: Correct, I knew that but forgot to add under things not yet working.

C7Engine/TileExtensions.cs
TomWerner: I don't think it's a question of efficiency or performance, but more a question of why this method is special compared to all other methods - most other methods don't have explicit checks that the arguments are non-null.

You could probably get away with a comment like `// REQUIRED: currentWorkerJob != null`

PR #556, opened 2025-02-15
Properly calculate city borders based on culture

TomWerner:
We now have nearly tile-for-tile accuracy with the real game for loaded save files!  There is 1 tile that is different from the real game to C7 in the screenshot below, which I've circled.

The difference I suspect is due to rules 7/8 from https://forums.civfanatics.com/threads/the-eight-laws-of-border-dynamics.106882/:
 - > VII. If an unassigned tile finds a tile of a single civ directly adjacent to both the NW and SE, it is assigned to that civ.
 - > VII. If an unassigned tile finds a tile of a single civ directly adjacent to both the NW and SE, it is assigned to that civ.

I'm not sure the best way to implement this without doing a second pass over all the tiles, but that may be what we need to do.

image

image

---

This change required changing how we track tile owners, from a player owner to a city owner, so that we can correctly check rank and culture comparisons.

#288

### DIFF
--- a/C7/Map/BorderLayer.cs
+++ b/C7/Map/BorderLayer.cs
@@ -16,11 +16,11 @@ namespace C7.Map {
 			int textureWidth = 128;
 			int textureHeight = 72;
 
-			// This loops slices the PCX image into separate border textures. 
+			// This loops slices the PCX image into separate border textures.
 			// The PCX image contains 4 rows and 2 columns:
 			// - Each row corresponds to a border direction.
 			// - The first column contains border textures for regular terrain.
-			// - The second column contains border textures for hills and mountains 
+			// - The second column contains border textures for hills and mountains
 			//   (rendering logic for textures of the second column is not yet implemented).
 			for (int j = 0; j < 4; j++) {
 				for (int k = 0; k < 2; k++) {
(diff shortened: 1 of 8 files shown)


KulkoBSW:
IF you use an exponent of 1,8 it is working for all Culture size up to 20K (Which is a victory condition) but it breaks again slightly at 101K, no idea what formula they are using there.
So the better formula should be Power( (Power(abs(X1-x2),1,8) +  Power(abs(y1-y2),1,8)), 1/1,8)

TomWerner:
@KulkoBSW nice! I've updated the code and PR description. How'd you figure that out?

KulkoBSW:
@TomWerner Have you tested it out with your code. I realize you still have that additional Sqrt in DeltaX/DeltaY and I am not totally sure how that will interfere with my test data. So the exponnent might still be slightly off.

As for How I got this? Studying math in University. changing the exponent changes the roundness of the circle and there is basically no other valid Metric on the plane. So I just wrote an excel to try out different exponents until I found one that fit.

TomWerner:
Yeah, I tested it with a quick hack to exponentially increase culture, and it looked good, and I also loaded the save game I was using for testing and updated the screenshot.

The `Sqrt(2)` factor is needed because this grid is rotated, that factor cancels out the rotation.

KulkoBSW:
I also checked it out on 5 editor saves and with that addition it works even for Culture 100K, so its probably the correct formula.

TomWerner:
This PR has been open for a week and Kulko's very helpful suggestions have been integrated. I'm going to merge.

PR #557, opened 2025-02-15
Load basic building data

esbudylin:
This commit implements loading basic building data from save and scenario files. It loads the list of possible buildings; however, loading data on already existing buildings is not yet implemented.

Additionally, this commit implements loading of the currently produced units and buildings from save files.

Next, I intend to work on loading existing buildings and adding logic for creating new ones.

### DIFF
--- a/C7/Text/c7-static-map-save.json
+++ b/C7/Text/c7-static-map-save.json
@@ -61515,6 +61515,423 @@
       ]
     }
   ],
+  "buildings": [
+    {
+      "name": "Palace",
+      "shieldCost": 100,
+      "populationCost": 0
+    },
+    {
+      "name": "Barracks",
+      "shieldCost": 40,
+      "populationCost": 0
+    },
+    {
+      "name": "Granary",
+      "shieldCost": 60,
+      "populationCost": 0
+    },
+    {
+      "name": "Temple",
+      "shieldCost": 60,
+      "populationCost": 0
+    },
+    {
+      "name": "Marketplace",
+      "shieldCost": 100,
+      "populationCost": 0
+    },
+    {
+      "name": "Library",
+      "shieldCost": 80,
+      "populationCost": 0
+    },
+    {
+      "name": "Courthouse",
+      "shieldCost": 80,
+      "populationCost": 0
+    },
+    {
+      "name": "Walls",
+      "shieldCost": 20,
+      "populationCost": 0
+    },
+    {
+      "name": "Aqueduct",
+      "shieldCost": 100,
+      "populationCost": 0
+    },
+    {
+      "name": "Bank",
+      "shieldCost": 160,
+      "populationCost": 0
+    },
+    {
+      "name": "Cathedral",
+      "shieldCost": 160,
+      "populationCost": 0
+    },
+    {
+      "name": "University",
+      "shieldCost": 200,
+      "populationCost": 0
+    },
+    {
+      "name": "Colosseum",
+      "shieldCost": 120,
+      "populationCost": 0
+    },
+    {
+      "name": "Factory",
+      "shieldCost": 240,
+      "populationCost": 0
+    },
+    {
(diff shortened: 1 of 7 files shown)


C7GameData/Building.cs
```diff
@@ -0,0 +1,7 @@
+namespace C7GameData {
+	public class Building : IProducible {
+		public string name { get; set; }
+		public int shieldCost { get; set; }
+		public int populationCost { get; set; }
```
TomWerner: Could you leave a comment explaining what population cost is for buildings?

TomWerner: Ah, I see you left a comment in `ImportCiv3.cs` - could you add that here too? Is it possible to mark this as `const` with a value of 0 to make it clearer this is to just satisfy the interface?

C7GameData/Save/SaveGame.cs
```diff
@@ -189,6 +190,7 @@ public GameData ToGameData() {
 		public List<Resource> Resources = new List<Resource>();
 		public List<SaveUnit> Units = new List<SaveUnit>();
 		public List<UnitPrototype> UnitPrototypes = new List<UnitPrototype>();
+		public List<Building> Buildings = new();
```
TomWerner, line 196: If you run the tests locally then in `C7GameDataTests/data/output/gotm_save_0.json` you should be able to grab the list of buildings from the base game rules. Could you copy paste those into `c7-static-map-save.json`?

TomWerner (changes requested):
Really nice!

TomWerner:
I'll get #548 merged, I fixed the goofy test issue there.

C7GameData/Building.cs
esbudylin: I don't think an interface member can be declared as a constant. In general, I think the purpose of this interface is to allow buildings to have a population cost in modded rule sets, so I don't think this field needs to be restricted. I added a comment to clarify a situation with Civ3 rules.

C7GameData/Save/SaveGame.cs
esbudylin: Thanks for the pointer! I've added the building data to the save.

C7/Text/c7-static-map-save.json
```diff
@@ -61515,6 +61515,423 @@
       ]
     }
   ],
+  "buildings": [
+    {
+      "name": "Palace",
+      "shieldCost": 100,
```
TomWerner, line 61521: Not for this PR, but we'll need to figure out how this works, because the palace cost increases over time or per city, right?

C7/Text/c7-static-map-save.json
```diff
@@ -64533,87 +64950,5 @@
         "tech-44"
       ]
     }
-  ],
-  "citizenTypes": [
```
TomWerner: Looks like these got lost in a merge, could you restore them?

C7GameData/ImportCiv3.cs
```diff
@@ -541,6 +545,17 @@ private void ImportSavCities() {
 			}
 		}
 
+		private (string, ProducibleType) CityToProducible(QueryCiv3.Sav.CITY city) {
+			PRTO[] unitPrototypes = biq.Prto ?? defaultBiq.Prto;
+
+			return city.ConstructingType switch {
+				0 => ("Warrior", ProducibleType.UNIT), // TODO: Wealth production is not implemented yet
```
TomWerner: Maybe make this "Worker"? Warriors aren't always producible

TomWerner (approved):

C7/Text/c7-static-map-save.json
esbudylin: There is a write-up on the palace cost on Civfanatics https://forums.civfanatics.com/threads/cost-of-the-palace-for-c3c.107789/ It looks like the the palace cost depends on the number of cities.

Issue #391, opened 2023-02-09
AI - Very Basic Tech Chooser Logic


QuintillusCFC:
This is for a basic first pass of the AI having some strategy in selecting what it researches.

Many of the mechanics that will influence this are not present yet.  Victory is not possible.  Diplomacy is not an option.  Switching governments is not possible.

My current thinking is to add two basic elements that would cause some influence:

 - Import or set the "aggressiveness" setting for civilizations.  The more aggressive they are, the more they favor techs that unlock military units.
 - Import or set civilization traits.  They are more likely to research traits that unlock buildings that correspond to the trait.

This may be better to do a bit later in the milestone, as we're more likely to have things such as buildings at that point.  The "totally random" research strategy can work until then.

PR #561, opened 2025-02-16
Implement basic AI player research

TomWerner:
Right now they just randomly pick techs to research without any particular strategy, but this will get us closer to some niceties, like tech trading and era advancement.

#391


TomWerner:
This PR is just shy of a week old, and merging it now avoids merge conflicts on my two main chains (improving AI, adding diplomacy). 

As always, I'm happy to address postsubmit comments.

PR #566, opened 2025-02-18
Implement the A* path finding algorithm

TomWerner:
This gives much nicer results when doing path finding over railroads in particular, but also should involve much less searching in the general case.

Our path finding still isn't perfect though, because it only looks at the tile movement cost, without thinking about unit movement costs. So in some cases it routes around an obstacle like jungle tiles, when it would be faster to go through them. I implemented that logic [in TilePath.cs](https://github.com/C7-Game/Prototype/blob/Development/C7GameData/AIData/TilePath.cs#L33-L68) so the goto cursor is accurate, but it hasn't been backported to the pathing logic yet.

### DIFF
--- a/C7Engine/AI/Pathing/AStarAlgorithm.cs
+++ b/C7Engine/AI/Pathing/AStarAlgorithm.cs
@@ -1,0 +1,109 @@
+using System.Collections.Generic;
+using System.Linq;
+using System;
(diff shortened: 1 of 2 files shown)


KulkoBSW:
@TomWerner Looks good to me.

PR #567, opened 2025-02-18
Delete unused pathing algorithms and their tests

TomWerner:
These can always be brought back from version history if we need them.


TomWerner:
https://github.com/C7-Game/Prototype/pull/569 is a child of this one and was reviewed - merging this PR first.

PR #568, opened 2025-02-18
Tune the A* land distance cost heuristic to handle railroads better

TomWerner:
This ensures that if we have railroads in a large "C" shape, units will follow the "C", instead of taking the more direct (and expensive) path.


TomWerner:
https://github.com/C7-Game/Prototype/pull/569 is a child of this one and was reviewed - merging this PR first.

PR #569, opened 2025-02-18
Define an EdgeWalker that is aware of unit movement costs

TomWerner:
This avoids weird pathing where a warrior will walk the long way around a jungle, taking multiple moves to walk across grassland, because the movement cost of the jungle is 3.


Masemiliano:
Good work!

The pathfinder seems to give much better paths than before.

Two things remain to be implemented (at least): 
1. Making the pathfinder check all tiles for enemy units - those tiles should be considered impassable when pathfinding (except maybe the target square, since you should be able to pathfind an attack?).
2. Checking whether the tech engineering is researched, and if it is, units should be able to use roads that cross rivers.

TomWerner:
Thanks! I'm going to count this as a review and get this chain of PRs in. I definitely agree with both things you mentioned about needing to be implemented - if you're interested in either of them feel free to give it a shot. I'm happy to point you at the relevant code.

Issue #571, opened 2025-02-19
Implement some very rudimentary diplomacy logic


TomWerner:
Things I'd like:
 - meeting other civs
 - tracking war status
 - declaring war (peace deals may be out of scope, we'll see)
 - tech trading

TomWerner:
#572, #624, #621, and #573 are all part of this.

PR #572, opened 2025-02-19
Import known civs and war statuses from SAVs, and add a diplomacy popup

TomWerner:
We now have the lower right hand "D" button for opening the diplomacy popup, and the popup lists the civs we know and whether we're at war with them.

In the LEAD struct I changed some byte arrays to longs/ints to make it easier to dump them to the console when debugging.

In order for the "D" button to work I had to shuffle the "GameStatus" node in front of the unit buttons node, because the invisible HBox for the unit buttons was intercepting all the mouse events.

Obligatory screenshot:

image

#571

### DIFF
--- a/C7/C7Game.tscn
+++ b/C7/C7Game.tscn
@@ -97,17 +97,6 @@ grow_horizontal = 2
 grow_vertical = 2
 mouse_filter = 2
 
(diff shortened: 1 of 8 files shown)


TomWerner:
This is now a week old, merging. As always, I'm happy to address postsubmit comments.

PR #573, opened 2025-02-19
When entering a tile check to see if we've met a new civ

TomWerner:
We can now discover new civs in a pure C7 game!

#571


TomWerner:
This is now a week old, merging. As always, postsubmit comment are welcome.

PR #574, opened 2025-02-20
Fix a bug in the CityProductionAI to avoid wasting settlers

TomWerner:
Previously when running in observer mode the log would show multiple instances of building settlers and then destroying them via the `JOIN_CITY` decision. This was because we produced the settler without having any idea where it would go, and once the settler arrived, we still didn't have a place to put it.

By only producing settlers once we have a candidate location we avoid this problem, and it seems to significantly speed up the AI's expansion rate.


rchoffardet:
LGTM 👍🏻

rchoffardet (approved):

TomWerner:
Thanks!

PR #575, opened 2025-02-20
Remove LuaCiv3

esbudylin:
The code in the LuaCiv3 projects implemented a rough prototype of interoperating C7 code with Lua. However, it hasn't been updated in several years and no longer works against the current C7 codebase. I think there is no need to support it, as implementing the actual scripting system from scratch would be easier than modifying the stale code.


TomWerner (approved):
Hooray for removing dead code!

PR #577, opened 2025-02-21
Have the UnitAI instances hold their own state, rather than requiring casts

TomWerner:
This allows us to push more logic into the ai classes themselves, rather than having tons of it piling up in `PlayerAI.cs`. To do this I had to move `UnitAI.cs` over to `C7GameData` to avoid a dependency cycle, but declaring the interface there seems fine.

To test this I ran in observer mode for a while (which required fixing a small bug in Game.cs)


rchoffardet:
It's pretty hard to read as I'm unfamiliar with this part of the solution.
But it looks good to me, still

rchoffardet (approved):

TomWerner:
Thanks!

PR #578, opened 2025-02-22
MapUnitExtensions: Move can road/irrigate/mine into the Tile class

TomWerner:
This will let the worker AI check if a tile improvement can be performed without needing to be standing on the tile first.

### DIFF
--- a/C7GameData/Tile.cs
+++ b/C7GameData/Tile.cs
@@ -301,6 +301,56 @@ namespace C7GameData {
 			return riverNorth || riverNortheast || riverEast || riverSoutheast || riverSouth || riverSouthwest || riverWest || riverNorthwest;
 		}
 
+		public bool CanBeMined() {
+			return overlayTerrainType.miningBonus > 0 && !overlays.mine && !overlays.irrigation && cityAtTile == null;
+		}
+
+		public bool CanBeRoaded() {
+			return IsLand() && !overlays.road;
+		}
+
+		// TODO: This method doesn't handle two important irrigation cases:
+		//  - inland lakes/seas: we need to figure out what is fresh/salt water
+		//  - Electricity tech, to allow irrigating w/o fresh water access
+		public bool CanBeIrrigated(Player player) {
+			// Irrigation can't be done if there is no irrigation bonus for the 
+			// tile or if there's already an improvement or city on the tile.
+			if (overlayTerrainType.irrigationBonus == 0 ||
+				overlays.mine ||
+				overlays.irrigation ||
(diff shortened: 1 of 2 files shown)


C7GameData/Tile.cs
```diff
@@ -301,6 +301,56 @@ public bool BordersRiver() {
...
+
+			foreach (KeyValuePair<TileDirection, Tile> dirToTile in neighbors) {
+				// If a neighboring tile is irrigated, this tile has fresh water access.
+				if (dirToTile.Value.overlays.irrigation) {
+					return true;
+				}
+
+				// Special case, if we are neighboring the worker's city, check
+				// if the city can act as part of an irrigation chain.
+				if (dirToTile.Value.cityAtTile?.owner == player) {
```
rchoffardet: Is this true? Is it impossible to irrigate a tile next to city with fresh water if the city doesn't belong to you?

rchoffardet:
Shouldn't we put this logic to a dedicated class? I'm thinking about an abstract class that takes, the player, tile, and unit context and has two methods/properties `IsAllowed` and `Proceed`. Each action (mining, irrigating, whatever) would be inherited from this base class. This would allow unit testing and help people dive in the code :) 

What fo you think?

C7GameData/Tile.cs
TomWerner: Good catch, it is indeed possible.

image

TomWerner:
> Shouldn't we put this logic to a dedicated class? I'm thinking about an abstract class that takes, the player, tile, and unit context and has two methods/properties IsAllowed and Proceed. Each action (mining, irrigating, whatever) would be inherited from this base class. This would allow unit testing and help people dive in the code :)

That's probably a good idea, yeah. That would kind of mirror the TFRM struct (Tfrm.cs) that civ3 uses. That struct only has the tech/resources needed for an improvement, not the set of tiles that support it (that's in the Tile struct).

I think we'll want to see how https://github.com/C7-Game/Prototype/pull/585 shakes out first (as well as https://github.com/C7-Game/Prototype/pull/584, which fixes some of the multi-turn stuff as well) before implementing such a class though. Right now all the worker moves are trivial and take one turn, so once we have a working example of multi-turn worker moves we'll have a better idea of the data structure needed to model that.

So TL;DR, yes I agree, but probably not in this PR.

rchoffardet (approved):

rchoffardet:
LGTM 👍🏻

TomWerner:
Thanks!

PR #580, opened 2025-02-23
Load existing buildings

esbudylin:
This commit implements loading of produced buildings from save and scenario files. It introduces a new class, CityBuilding, which represents a produced building, and adds functionality to store the list of buildings within the City class.

Currently, the method for loading city buildings from Civ3 data handles destroyed and sold buildings incorrectly, as it loads them as existing buildings. The issue arises because Civ3 save files preserve data on both existing and destroyed/sold buildings, and I haven't yet found a way to distinguish between them.


C7GameData/City.cs
```diff
@@ -178,6 +187,15 @@ public int GetBorderExpansionLevel() {
 			return (int)Math.Floor(Math.Log10(culture)) + 1;
 		}
 
+		public void AddBuilding(Building building) {
+			buildings.Add(new CityBuilding {
+				building = building,
+				builtByPlayer = owner,
+				year = 1, // TODO: Implement in-game year tracking
+				culture = 0
```
TomWerner: Do we need a TODO to track building culture as well?

TomWerner (approved):

C7GameData/City.cs
esbudylin: No, I don't think so. The CityBuilding culture value is cumulative (it represents the total culture produced by the building), so it will always be zero for new buildings.

TomWerner: Ah, gotcha - maybe name it totalCulture then? To make it clear it isn't a per-turn culture value, which I assume we'll add at some point?

TomWerner: (I see the comment explains that but making the variable name clearer might help too)

esbudylin: Good point! I've renamed the relevant fields.

PR #582, opened 2025-02-23
Fixing the transition of the fog of war

rchoffardet:
This MR tries to fix the fog of war. 
before:
image

after:
image

~~it's not perfect, though...
When zooming in and out, artifacts may appear:
image
I don't think that's directly related to the fog of war; it's more like a floating-point error in tile placement.~~
It should be fixed now.

The end of aren't well handled, yet (north and south).
image
image
I don't know the reason 🤷🏻 

Known but not actively visible tiles aren't handled.

### DIFF
--- a/C7/Map/FogOfWarLayer.cs
+++ b/C7/Map/FogOfWarLayer.cs
@@ -15,7 +15,8 @@ namespace C7.Map {
 		}
 
 		public override void drawObject(LooseView looseView, GameData gameData, Tile tile, Vector2 tileCenter) {
-			Rect2 screenTarget = new Rect2(tileCenter - tileSize / 2, tileSize);
+			Vector2 fogOrigin = new(tileCenter.X - tileSize.X/2, tileCenter.Y);
+			Rect2 screenTarget = new Rect2(fogOrigin, tileSize);
 			TileKnowledge tileKnowledge = gameData.GetHumanPlayers()[0].tileKnowledge;
 			//N.B. FogOfWar.pcx handles both totally unknown and fogged tiles, indexed in the same file.
 			//Hence the trinary math rather than the more commonplace binary.
--- a/C7/Map/FogOfWarLayer.cs
+++ b/C7/Map/FogOfWarLayer.cs
@@ -19,7 +19,7 @@ namespace C7.Map {
 			TileKnowledge tileKnowledge = gameData.GetHumanPlayers()[0].tileKnowledge;
 			//N.B. FogOfWar.pcx handles both totally unknown and fogged tiles, indexed in the same file.
 			//Hence the trinary math rather than the more commonplace binary.
-			if (!tileKnowledge.isTileKnown(tile) || tileKnowledge.isBorderOfTileKnowleged(tile)) {
+			if (!tileKnowledge.isTileKnown(tile) || tileKnowledge.isBorderOfTileKnowlege(tile)) {
 				int sum = 0;
 				if (tileKnowledge.isTileKnown(tile.neighbors[TileDirection.NORTH]) || tileKnowledge.isTileKnown(tile.neighbors[TileDirection.NORTHWEST]) || tileKnowledge.isTileKnown(tile.neighbors[TileDirection.NORTHEAST]))
 					sum += 1 * 2;
--- a/C7/Map/FogOfWarLayer.cs
+++ b/C7/Map/FogOfWarLayer.cs
@@ -38,7 +38,8 @@ namespace C7.Map {
 		private Rect2 getRect(int sum) {
 			int row = sum / 9;
 			int col = sum % 9;
-			return new Rect2(col * tileSize.X, row * tileSize.Y, tileSize);
+			// The 0.999f scaling is a hack to prevent "seams".
+			return new Rect2(col * tileSize.X, row * tileSize.Y, tileSize * 0.999f);
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


TomWerner:
Could you provide a before and after screenshot of what this is fixing? It wasn't 100% clear to me which screenshots are before vs after.

C7/Map/FogOfWarLayer.cs
```diff
@@ -29,9 +29,8 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
 					sum += 9 * 2;
 				if (tileKnowledge.isTileKnown(tile.neighbors[TileDirection.SOUTH]) || tileKnowledge.isTileKnown(tile.neighbors[TileDirection.SOUTHWEST]) || tileKnowledge.isTileKnown(tile.neighbors[TileDirection.SOUTHEAST]))
 					sum += 27 * 2;
-				if (sum != 0) {
-					looseView.DrawTextureRectRegion(fogOfWarTexture, screenTarget, getRect(sum));
-				}
+
+				looseView.DrawTextureRectRegion(fogOfWarTexture, screenTarget, getRect(sum));
```
rchoffardet, line 34: This could lead to performance reduction, as black are drawn on every unknown tiles

rchoffardet:
> Could you provide a before and after screenshot of what this is fixing? It wasn't 100% clear to me which screenshots are before vs after.

They are all made afterward.
Current fog of war looks like this

image

C7/PCXToGodot.cs
```diff
@@ -46,11 +46,9 @@ public static ImageTexture getPureAlphaFromPCX(Pcx alphaPcx) {
 		for (int y = 0; y < alphaPcx.Height; y++) {
 			for (int x = 0; x < alphaPcx.Width; x++, dataIndex++) {
 				int index = alphaPcx.ColorIndexAt(x, y);
-				if (transparentColorIndexes.Contains(index)) {
-					bufferData[dataIndex] = 0;
-				} else {
-					bufferData[dataIndex] = alphaData[index] << 24;
-				}
+
+					bufferData[dataIndex] = (255 - alphaData[index]) << 24;
```
TomWerner: The formatting here seems off - could you remove the blank lines and run `dotnet format C7.sln whitespace` from the `C7` directory?

C7/Map/FogOfWarLayer.cs
TomWerner: Why not avoid doing the drawing like before then? How does the unconditional drawing fix the issue?

C7/project.godot
```diff
@@ -205,5 +205,5 @@ limits/debugger_stdout/max_chars_per_second=65535
 [rendering]
 
 textures/canvas_textures/default_texture_filter=0
-environment/defaults/default_clear_color=Color(0.301961, 0.301961, 0.301961, 1)
+environment/defaults/default_clear_color=Color(0, 0, 0, 1)
```
TomWerner, line 208: How is this change related to the fog of war?

TomWerner (changes requested):
So right now this seems like a visual regression - 
image


Do you need the trick applied [here](https://github.com/C7-Game/Prototype/blob/Development/C7/MapView.cs#L99-L102) to multiply by 100.1%?

C7/PCXToGodot.cs
rchoffardet: done

C7/project.godot
rchoffardet: It's a workaround about the "seams" in the full fog

rchoffardet:
I did my best to improve it without having to deal with a major refactor.
The difficulty is that the fog tile transition is drawn over an unknown tile and doesn't cover the full tile
The screenshot with only the transition (so without [removing the condition](https://github.com/C7-Game/Prototype/pull/582#discussion_r1966945506)), makes the issue, obvious:
image

The tile below the fog transition is a new concept of "border of knowledge" tiles. it's required because the map isn't fully drawn.

rchoffardet:
> So right now this seems like a visual regression -
> Do you need the trick applied [here](https://github.com/C7-Game/Prototype/blob/Development/C7/MapView.cs#L99-L102) to multiply by 100.1%?

Doesn't seem to work any better :( But I only have very slight seams, they seem bigger on your screenshot. Maybe it would help in your situation.
But that's not related to the fog, I think the issue was always here, just less visible

The goal of this PR was to fix the fog transition, only (I'll update the title)

rchoffardet:
> Do you need the trick applied [here](https://github.com/C7-Game/Prototype/blob/Development/C7/MapView.cs#L99-L102) to multiply by 100.1%?

The opposite: `* 0.999f ` it's a dirty hack anyway but "fixes" the issue

C7/Map/FogOfWarLayer.cs
rchoffardet: Can't easily do it (see [this comment](https://github.com/C7-Game/Prototype/pull/582#issuecomment-2677227067)). It would be possible if we stored the border of the border of the known tiles 😬

TomWerner:
Sorry to continue the back and forth, but this is a tricky issue and I want to make sure we get it right.

Here's a screenshot of the default "new game" in C7 with the grid turned on.

image

We can see the edges of the unknown tiles, but not what's on them (no idea if it has resources, etc).

This is pretty close to civ3's fog of war:

image


But here's a screenshot with this PR - we can now see extra information, specifically that the coast tiles have fish:

image

Do you know why this visibility changed?

TomWerner:
Also to be clear, thank you for working on this! Looking at those screenshots this will look so much cleaner when this is done.

rchoffardet:
@TomWerner nice catch 👍🏻 I pushed the fix

Here is the result, now:
image

TomWerner (approved):
Thanks!

PR #583, opened 2025-02-25
ImportCiv3: Fix building loading in scenarios

TomWerner:
The scenario loading tests caught this breakage, but they don't run as part of the CI because it requires a civ3 installation.


esbudylin:
Thanks for the fix! I'm going to set up a pre-push hook in Git to avoid such blunders in the future.

esbudylin (approved):

PR #584, opened 2025-02-26
Process multi-turn moves during the unit's order, not end-of-turn

TomWerner:
This better matches the base game, where if I have two units, a settler and a worker, and have the settler go on a multi-turn move, the next turn the worker will be selected. If I then press wait, the settler will do its movement phase.

This also is necessary for the single player "road-to"+"wait" hack, where cycling through the full list of units will result in a road being completed, which can then be used a turn early.


TomWerner:
Merging early to unblock Kulko on some worker move stuff.

PR #585, opened 2025-02-26
Kulkobsw/multiturn irrigation take2

KulkoBSW:
Since I had already to redo most of the work after some PR by tom, I would like to Merge my current state into the main branch.

The irrigation is working and animating
it is also stopping when you reactivate the worker, (but does not ask yet if you really want to.)
It also reads the state from the save, although quite brute forcy. I am also not sure yet where to add the animation in that case so it isn't animating a loaded save yet.
It does not yet read things from terraform (this would be the next step, which should also remove the ugly hardcoding)

This one replaces PR #555

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -347,6 +347,11 @@ public partial class Game : Node2D {
 			unit.path = TilePath.NONE;
 		}
 
+		// clear the current WorkerJob
+		if (unit.WorkerJob != null) {
+			unit.resetWorkerJob();
+		}
+
 		this.CurrentlySelectedUnit = unit;
 		this.KeepCSUWhenFortified = unit.isFortified; // If fortified, make sure the autoselector doesn't immediately skip past the unit
 
--- a/C7/Text/c7-static-map-save.json
+++ b/C7/Text/c7-static-map-save.json
@@ -58684,7 +58684,9 @@
       "hitPointsRemaining": 3,
       "movePointsRemaining": 1,
       "facingDirection": "north",
-      "experience": "Regular"
+      "experience": "Regular",
+      "workerProgressTowardsJob": 0,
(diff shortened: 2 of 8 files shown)


TomWerner:
I'll try to keep the round trips down this time :)

A couple of comments:
 - Could you run `dotnet format C7.sln whitespace`? from the C7 directory? (I should get this set up as a CI check)
 - The tests are currently broken. The quick fix is to do `dotnet test ../C7GameDataTests/C7GameDataTests.csproj` from the C7 directory, and then run `cp ../C7GameDataTests/data/output/static-save-was-game-data.json Text/c7-static-map-save.json` so that none of the workers start with a worker move of 0.
 - The feature itself seems a little broken right now - if you start an irrigation the game still lets you mine or road that tile.

KulkoBSW:
@TomWerner 
I fixed the Whitespaces and the Test.
As for the last part, that has never worked yet, and I am currently not sure how to best implement it.
- Remember on the tile that it is currently irrigated, and check against that? (Is a faster check, but more prone to errors I believe) 
- Check all units to see if they are currently Working on irrigation?  (Easier to implement, but ore ressource intensive to check)

And lastly how is it defined, that you cannot Mine and irrigate at the same time, but you can road and irrigate at the same time? Its not important for the current PR, but would be one of the next steps in making this more dynamic.

TomWerner:
Thanks Kulko - I'm going to merge this as-is to avoid more round trips for you. Thank you for the contribution!

One quick note, and I'll fix this myself, your edits to the JSON file seem to have added carriage returns, causing a very large diff. I'll strip those out in a quick PR after merging this.

Some of this work may get refactored a bit due to https://github.com/C7-Game/Prototype/pull/584, just FYI if you see merge issues in the future.

PR #586, opened 2025-02-26
Support cancelling multi-turn moves 

TomWerner:
This also adds the "This unit has already moved" popup.

image

Being able to cancel multi-turn moves is important, as players should be able to change their minds. This will also be important for breaking out of worker moves, automation, and exploring.


TomWerner:
Merging early to unblock Kulko on some worker move stuff.

PR #587, opened 2025-02-26
PlayerAI: Move ExplorerAIData creation to ExplorerAI

TomWerner:
This will be needed in a following PR that adds the explore action to player controlled units.

While here I reduced the [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88), which makes the code easier to read.


TomWerner:

PR #588, opened 2025-02-26
Fix city border alignment

rchoffardet:
Here is a very small PR to fix city border misalignment with the grid.

Before:
image

After:
image

### DIFF
--- a/C7/Map/BorderLayer.cs
+++ b/C7/Map/BorderLayer.cs
@@ -83,8 +83,8 @@ namespace C7.Map {
 					Vector2 size = texture.GetSize();
 					Vector2 offset = size/2;
 					// these values were found experimentally to alignment with the grid
-					offset.X += size.X * 0.01f;
-					offset.Y += size.Y * 0.05f;
+					offset.X += size.X * 0.00f;
+					offset.Y += size.Y * 0.055f;
 
 					looseView.DrawTexture(texture, tileCenter - offset);
 				}
--- a/C7/Map/BorderLayer.cs
+++ b/C7/Map/BorderLayer.cs
@@ -82,6 +82,7 @@ namespace C7.Map {
 					ImageTexture texture = GetBorderTexture(entry.Value, borderColor);
 					Vector2 size = texture.GetSize();
 					Vector2 offset = size/2;
+					// these values were found experimentally to alignment with the grid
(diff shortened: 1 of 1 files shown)


C7/Map/BorderLayer.cs
```diff
@@ -80,7 +80,10 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
 			foreach (var entry in directionToTextureIdx) {
 				if (tile.neighbors[entry.Key].owningCity?.owner != tile.owningCity?.owner) {
 					ImageTexture texture = GetBorderTexture(entry.Value, borderColor);
-					Vector2 offset = texture.GetSize() * 0.5f;
+					Vector2 size = texture.GetSize();
+					Vector2 offset = size/2;
+					offset.X += size.X * 0.01f;
```
TomWerner: were these constants just found experimentally or are they derived from something? Could you leave a comment either way?

TomWerner (approved):

C7/Map/BorderLayer.cs
rchoffardet: done 👍🏻

rchoffardet: hmm, wait a bit before merging, i need to make some adjustments still (just saw that when zooming in)

rchoffardet: done
image

C7/Map/BorderLayer.cs
```diff
@@ -80,7 +80,11 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
 			foreach (var entry in directionToTextureIdx) {
 				if (tile.neighbors[entry.Key].owningCity?.owner != tile.owningCity?.owner) {
 					ImageTexture texture = GetBorderTexture(entry.Value, borderColor);
-					Vector2 offset = texture.GetSize() * 0.5f;
+					Vector2 size = texture.GetSize();
+					Vector2 offset = size/2;
+					// these values were found experimentally to alignment with the grid
+					offset.X += size.X * 0.00f;
```
TomWerner: If this is multiplying by zero does this line do anything?

rchoffardet: my bad, forgot to remove it. Done.

TomWerner:
Thanks!

PR #589, opened 2025-02-27
Implement the "explore" and "automate worker" actions

TomWerner:
These use the explorer and worker AI classes, so in addition to giving us useful functionality, it gives us an easier way to test the explorer logic.

While these behaviors can be observed in observer mode (ctrl+alt+shift+O), observer mode doesn't show tile knowledge, so judging the usefulness of the explorer ai is hard. With this PR we can see that it's not that great (choosing moves that reveal none or only a single tile), and will allow us to improve it.

Both of these automated actions can be cancelled by clicking on the unit, just like in civ3.


TomWerner:

PR #590, opened 2025-02-27
ExplorerAI: Tune the "near city" exploration to be more useful

TomWerner:
The previous distance penalty was way too high. Here's what we got after 20 turns of exploring with one unit prior to this PR.

image

And here is the same 20 turns with this PR:

image

We have covered significantly more ground, and the AI player has managed to meet us - which is much more interesting for the player.


TomWerner:

PR #592, opened 2025-02-27
Fix black spots on loaded PCX textures

esbudylin:
The issue fixed by this commit was introduced in PR #504. It was caused by the addition of a new color index to the set of transparent colors. This new index was required to correctly display border textures, however, when applied to other textures it produced black spots in areas that should have remained opaque.

This commit modifies the methods for loading PCX textures, so they can accept a custom set of transparent color indexes. To improve readability, two new structs are introduced:

- ColorOptions: Encapsulates transparency and shadows settings.
- CropRegion: Defines a cropped section of a texture.

### DIFF
--- a/C7/Map/BorderLayer.cs
+++ b/C7/Map/BorderLayer.cs
@@ -24,7 +24,11 @@ namespace C7.Map {
 			//   (rendering logic for textures of the second column is not yet implemented).
 			for (int j = 0; j < 4; j++) {
 				for (int k = 0; k < 2; k++) {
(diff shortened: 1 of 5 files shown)


TomWerner (approved):

TomWerner:
Out of curiosity, is `shadows` ever false?

esbudylin:
> Out of curiosity, is `shadows` ever false?

Yes, at least within the `getImageFromPCXWithAlphaBlend` method.

PR #595, opened 2025-02-28
Move multi-turn worker moves to PerformBusyAction 

TomWerner:
This now puts it in the same place as multi-turn moves and multi-turn automation, allowing for some simplifications.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -712,7 +712,7 @@ public partial class Game : Node2D {
 			// unimplemented
 		}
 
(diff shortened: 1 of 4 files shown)


KulkoBSW:
I think you should also get rid of my old function AddTotalProgressAndResetOtherJobs in MapUnitExtensions.
Its not used anymore and I think the logic implemented there is also wrong. (Clicking on Irrigate doesn't cancel other units building mines, it is just not allowed.
Otheerwise It looks fine with me and works.

TomWerner:
Good call, done. Thanks for the review!

PR #597, opened 2025-03-01
Add a test that plays the game in "headless" mode

TomWerner:
This allows us to confirm the game continues to round trip through the save format even once we have started playing the game. The test (and loading the genered save files) caught a number of issues, and I suspect there are more to catch.


C7GameData/Save/SaveUnit.cs
```diff
@@ -14,6 +14,7 @@ public class SaveUnit : IHasID {
 		public float movePointsRemaining;
 		public string action; // "fortified"
 		public TileDirection facingDirection;
+		public int turnsUntilPriorityReevaluation = 0;
```
esbudylin: Can you elaborate on this change? I don't see how it's relevant to the new test.

C7GameDataTests/SaveTest.cs
```diff
@@ -80,6 +81,78 @@ public void SimpleSave() {
...
+		}
+		Assert.Equal(30, game.turn);
+
+		// Save the game.
+		string outputDirectSavePath = getDataPath("output/headless-game-direct-save.json");
+		SaveGame.FromGameData(game).Save(outputDirectSavePath);
+
+		// Load the saved game and save it again.
+		string roundTrippedSavePath = getDataPath("output/headless-game-round-tripped-save.json");
+		GameData roundTrippedGameData = SaveGame.Load(outputDirectSavePath).ToGameData((City c, CitizenType ct) => {
```
esbudylin: Is it possible to move the creation of the dummy delegate into a new overload of ToGameData method? The same applies to CreateGame.createGame method. This will make code easier to read and shorten method calls.

C7GameData/Save/SaveUnit.cs
TomWerner: This is one of the fields that wasn't being round-tripped correctly, that the test caught.

C7GameDataTests/SaveTest.cs
TomWerner: Good call, done

C7GameData/Save/SaveUnit.cs
TomWerner: Oops no, this one isn't needed. removed.

esbudylin (approved):

TomWerner:
Thanks!

PR #598, opened 2025-03-01
Load requirements for building and unit production

esbudylin:
This commits implement loading of several production requirements. For both buildings and units, it loads the required technology. For buildings it also loads the required building (like a Library for a University) and the wonder status.

Additionally, this commit refactors the ToGameData method of the SaveGame class, splitting it into multiple methods for better readability.

With these changes we are getting closer to creating a proper method for calculating a list of producible units and buildings, although we still need to load data on unique units and unit obsolescence.

### DIFF
--- a/C7GameData/ImportCiv3.cs
+++ b/C7GameData/ImportCiv3.cs
@@ -42,6 +42,7 @@ namespace C7GameData {
 		/// <param name="c7Save">Destination C7 in-memory structure</param>
 		private void ImportSharedBiqData() {
 			ImportRaces();
+			ImportTechs();
 			ImportUnitPrototypes();
 			ImportBuildings();
 			ImportCiv3TerrainTypes();
@@ -53,7 +54,6 @@ namespace C7GameData {
 			save.HealRates["city"] = 2;
 			// save.ScenarioSearchPath = biq?.Game[0].ScenarioSearchFolders;
 			ImportBarbarianInfo();
-			ImportTechs();
 			ImportCitizenTypes();
 		}
 
@@ -655,7 +655,7 @@ namespace C7GameData {
 		private void ImportUnitPrototypes() {
 			PRTO[] Prto = biq.Prto ?? defaultBiq.Prto;
 			foreach (PRTO prto in Prto) {
-				UnitPrototype prototype = new UnitPrototype();
+				SaveUnitPrototype prototype = new();
 				if (prto.Type == PRTO.TYPE_SEA) {
 					prototype.categories.Add("Sea");
 				} else if (prto.Type == PRTO.TYPE_LAND) {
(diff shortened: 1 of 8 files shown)


C7GameData/ImportCiv3.cs
```diff
@@ -42,6 +42,7 @@ private ImportCiv3() {
 		/// <param name="c7Save">Destination C7 in-memory structure</param>
 		private void ImportSharedBiqData() {
 			ImportRaces();
+			ImportTechs();
```
KulkoBSW, line 45: I think techs should be the first thing to import, as Races also include a reference to techs once we load everything.

KulkoBSW (commented):
Looks good to me. Just one minor comment Which is not really relevant at the current state.

C7GameData/ImportCiv3.cs
```diff
@@ -723,11 +727,22 @@ private void ImportBuildings() {
...
+					isSmallWonder=bldg.SmallWonder,
+					isGreatWonder=bldg.Wonder,
 				};
+
+				if (bldg.RequiredAdvance != -1) {
+					building.requiredTech = save.Techs[bldg.RequiredAdvance].id;
+				}
+
+				if (bldg.RequiredBuilding != -1) {
+					building.requiredBuilding = Bldg[bldg.RequiredBuilding].Name;
```
TomWerner, line 743: Is there a reason this is the building name and not the ID?

C7GameData/UnitPrototype.cs
```diff
@@ -13,6 +14,7 @@ public class UnitPrototype : IProducible {
 		public string artName { get; set; }
 		public int shieldCost { get; set; }
 		public int populationCost { get; set; }
+		public Tech requiredTech { get; set; }
```
TomWerner, line 17: If we made this an ID instead of a full tech would we need the SaveUnitPrototype? A lot of times just the ID is sufficient for the techs, because we're doing something like `player.knownTechs.Contains(<some tech id>);`

TomWerner (approved):

C7GameData/ImportCiv3.cs
esbudylin: Like UnitPrototype, the Building class doesn't have ID field. I think it is not needed since both entities can use their names as unique identifiers.

C7GameData/UnitPrototype.cs
esbudylin: It can be done for now, but eventually we'll need to store a UnitPrototype instance that the UnitPrototype upgrades to and, probably, a Civilization to which it belongs as a unique unit. So, I think having a separate SaveUnitPrototype is inevitable.

TomWerner: Ah, that makes sense - thanks!

TomWerner:
Do you want to merge this one to avoid conflicts with #599?

PR #599, opened 2025-03-01
Kulkobsw/multiturn irrigation take2

KulkoBSW:
Added the functionality to load the terraform data from the biq files


C7GameData/Terraform.cs
```diff
@@ -0,0 +1,19 @@
...
+
+	public string Name;
+
+	public string CivilopediaEntry;
+
+	public int TurnsToComplete;
+
+	public ID RequiredTech;
+
+	public ID RequiredResource1;
```
esbudylin: Can these two fields be converted into a collection to support an arbitrary number of required resources?

C7GameData/Terraform.cs
```diff
@@ -0,0 +1,19 @@
...
+
+	public ID RequiredResource2;
+
+	public string Order;
```
TomWerner: Do you happen to know what this "order" field means?

KulkoBSW: That should map to the C7_action.
    {
      "id": "Terraform-1",
      "name": "Mine",
      "civilopediaEntry": "TFRM_Mine",
      "turnsToComplete": 12,
      "requiredTech": "tech-26",
      "order": "Build Mine"
    },
    {
      "id": "Terraform-2",
      "name": "Irrigation",
      "civilopediaEntry": "TFRM_Irrigation",
      "turnsToComplete": 8,
      "requiredTech": "tech-27",
      "order": "Irrigate"
    },

C7GameData/Terraform.cs
KulkoBSW: I solved this in a similar way to what I found in other examples. Why do you think it should warant a different approach here?

TomWerner: We used a collection for tech prereqs, so we've used collections like this before - https://github.com/C7-Game/Prototype/blob/Development/C7GameData/Tech.cs#L24

It ensures more flexibility - what if some mod wants a worker move that needs 3 resources?

KulkoBSW: In tech it made most sense since that has this branching structure and already up to 4 prerequisites.

No problem doing this, but I think we should than soon refactor to handle all similar instances (buildings, Prototypes, etc)  this way.

C7GameData/Terraform.cs
KulkoBSW: @TomWerner I see in UnitPrototypes we converted this direct into out own action. 
I will add this to my code to and rename it action.
What I wonder, whats the advantage of having our own nomenclature, the original nomenclature looks quite fine to me.

I would also suggest implementing the check for blocking of orders directly into the C7Action class. That way we could later also decide to open things up if we want to expose this class to modders.

TomWerner: That sounds reasonable to me.

> What I wonder, whats the advantage of having our own nomenclature, the original nomenclature looks quite fine to me.

There's definitely some work to do to make some variables look more uniform. If you see opportunities to do so feel free.

>  I would also suggest implementing the check for blocking of orders directly into the C7Action class. That way we could later also decide to open things up if we want to expose this class to modders.

SGTM

KulkoBSW: Adapted everything.
If that looks fine I would say it is self-contained enough to be merged.

C7GameData/ImportCiv3.cs
```diff
@@ -924,6 +925,59 @@ private void ImportCitizenTypes() {
...
+				};
+				tf.Action = ConvertCiv3OrderToAction(t.Order);
+				if (t.Required > -1) {
+					tf.RequiredTech.Add(save.Techs[t.Required].id);
+				}
+				save.TerraForms.Add(tf);
+			}
+		}
+
+		private static string ConvertCiv3OrderToAction(string order) {
```
TomWerner, line 948: We might consider doing the reverse eventually, since these strings would be better tooltips for the buttons. But that's not super important for this PR.

C7GameData/Terraform.cs
```diff
@@ -0,0 +1,19 @@
...
+public class Terraform {
+	public ID Id;
+
+	public string Name;
+
+	public string CivilopediaEntry;
+
+	public int TurnsToComplete;
+
+	public List<ID> RequiredTech = new();
```
TomWerner: This one I think I'd leave as a single ID - I don't think there are any cases where a building/resource/worker action/unit/etc need more than one tech.

TomWerner (approved):

TomWerner:
Oh, if you run the tests locally (`dotnet test ../C7GameDataTests/C7GameDataTests.csproj` from the C7 directory), you can copy over the base game terraform information into the c7 static json save from the generated `gotm_save_0.json` that is generated in `C7GameDataTests/data/output`

KulkoBSW:
> Oh, if you run the tests locally (`dotnet test ../C7GameDataTests/C7GameDataTests.csproj` from the C7 directory), you can copy over the base game terraform information into the c7 static json save from the generated `gotm_save_0.json` that is generated in `C7GameDataTests/data/output`

What will be the effect of that?

TomWerner:
> > Oh, if you run the tests locally (`dotnet test ../C7GameDataTests/C7GameDataTests.csproj` from the C7 directory), you can copy over the base game terraform information into the c7 static json save from the generated `gotm_save_0.json` that is generated in `C7GameDataTests/data/output`
> 
> What will be the effect of that?

That ensures that when you click "new game" the `GameData` will have your new data.

TomWerner:
I can send you a PR for that if you want to just get this one merged

KulkoBSW:
> > > Oh, if you run the tests locally (`dotnet test ../C7GameDataTests/C7GameDataTests.csproj` from the C7 directory), you can copy over the base game terraform information into the c7 static json save from the generated `gotm_save_0.json` that is generated in `C7GameDataTests/data/output`
> > 
> > 
> > What will be the effect of that?
> 
> That ensures that when you click "new game" the `GameData` will have your new data.

Okay done that, it also worked for my game, but now I have broken the exports again. ;-(

TomWerner:
Thanks!

PR #600, opened 2025-03-03
Rewrite ExplorerAI and Return an enum from `PlayTurn`, not a bool

TomWerner:
The previous return value was confusing and conflated "I'm done with my goal" and "I hit an error trying to reach my goal". This resulted in stack overflows when using the automated explore functionality on a human unit, resulted in "goto" being used in the settler ai, and some other oddities that would pop up when running in observer mode.

The new enum allows better signaling of how the AI is progressing and better error handling.

I also consolidated the logic for "move the AI unit along a path", since it was repeated in a couple of places. The annoying C7Engine/C7GameData split required that I put this in an extension class rather than being part of the UnitAI class as a protected method, like I'd prefer.

This work also revealed that the explorer ai was overly complicated and different from the other unit AIs in that it didn't fit the pattern of the AI data being a plan, and PlayTurn executing the plan. To solve that I rewrote it, and the resulting behavior is much nicer:
 - we no longer get big clumps of explorers heading for the same tile at different speeds
 - debugging what it is doing is easier
 - we have to check far fewer tiles because we take advantage of the border tiles we track

Lastly, I had the AI evaluation skip fortified units, which sped things up a bunch and stopped log spam from units repeatedly fortifying themselves as city defenders.

From https://github.com/C7-Game/Prototype/pull/590, here's what 20 turns of one unit exploring looked like prior to this PR: 
image

Now we have this: 
image

which covers much more ground and finds another civ's cities. The new approach is also far more efficient with multiple explorers at play.

With building a city, getting 2 units and then exploring we used to have (20 turns):

image

We now have (20 turns):

image

### DIFF
--- a/C7Engine/AI/UnitAI/CombatAI.cs
+++ b/C7Engine/AI/UnitAI/CombatAI.cs
@@ -19,14 +19,16 @@ namespace C7Engine.AI.UnitAI {
 			combatAIData = d;
 		}
 
-		public bool PlayTurnImpl(Player player, MapUnit unit) {
+		public C7GameData.UnitAI.Result PlayTurnImpl(Player player, MapUnit unit) {
+			if (unit.location == combatAIData.destination) {
+				return C7GameData.UnitAI.Result.Done;
+			}
+
 			//Move along the path to the place where the unit plans to enter combat.
 			//This might initiate combat.
 			//Once there are more combat goals than run to a barbarian camp,
 			//we will need fancier algorithms
-			unit.path = combatAIData.path;
-			unit.moveAlongPath();
-			return true;
+			return this.TryToMoveAlongPath(unit, combatAIData.path, /*allowCombat=*/true);
 		}
 
 		public string SummarizePlan() {
--- a/C7Engine/AI/UnitAI/ExplorerAI.cs
+++ b/C7Engine/AI/UnitAI/ExplorerAI.cs
@@ -10,207 +10,132 @@ using Serilog;
 namespace C7Engine {
 	public class ExplorerAI : C7GameData.UnitAI {
 		private static ILogger log = Log.ForContext<ExplorerAI>();
-		public ExplorerAIData explorerData;
+		public ExplorerAIData data;
 
 		public ExplorerAI(ExplorerAIData d) {
-			explorerData = d;
+			data = d;
 		}
 
 		public static ExplorerAIData? MaybeMakeAiData(MapUnit unit, Player player) {
-			ExplorerAIData ai = new();
-			if (unit.unitType.categories.Contains("Sea")) {
-				ai.type = ExplorerAIData.ExplorationType.COASTLINE;
-				log.Information("Set coastline exploration AI for " + unit);
-				return ai;
+			HashSet<Tile> borderTiles = player.tileKnowledge.borderTiles;
+
+			IEnumerable<Tile> candidates = borderTiles.Where(x => (x.IsLand() && unit.IsLandUnit()) || (!x.IsLand() && !unit.IsLandUnit()));
+			ExplorerAIData? result = PickBestTileToExplore(unit, CalculateExplorationScores(player, unit, candidates));
+
+			if (result == null) {
+				return result;
 			}
 
-			if (unit.location.unitsOnTile.Exists((x) => x.unitType.categories.Contains("Sea"))) {
-				ai.type = ExplorerAIData.ExplorationType.ON_A_BOAT;
-				//TODO: Actually put the unit on the boat
-				log.Information("Set ON_A_BOAT exploration AI for " + unit);
-				return ai;
-			}
-
-			//Isn't a Settler.  If there's a city at the location, it's defended.  No boats involved.  What's our priority?
-			//If there is land to explore, we'll try to explore it.
-			//Long-term TODO: Should only send tiles on this landmass.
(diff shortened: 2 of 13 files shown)


C7Engine/AI/UnitAI/WorkerAI.cs
```diff
@@ -118,25 +109,34 @@ bool UnitAI.PlayTurnImpl(Player player, MapUnit unit) {
...
 		}
 
-		private bool PerformWorkerMove(MapUnit unit, string workerMove) {
+		private UnitAI.Result PerformWorkerMove(MapUnit unit, string workerMove) {
 			if (workerMove == C7Action.UnitBuildRoad) {
 				if (unit.canBuildRoad()) {
 					unit.buildRoad();
-					return true;
+					return UnitAI.Result.InProgress;
+				}
```
KulkoBSW, line 117: Is it a problem that the result might already be Done? The terraform action might actually only 1 turn in some scenarios, or due to smarter future AI the AI might use worker gangs.

C7Engine/AI/UnitAI/WorkerAI.cs
```diff
@@ -118,25 +109,34 @@ bool UnitAI.PlayTurnImpl(Player player, MapUnit unit) {
...
+				if (unit.location.overlays.road) {
+					return UnitAI.Result.Done;
 				}
-				return false;
+				return UnitAI.Result.Error;
 			} else if (workerMove == C7Action.UnitBuildMine) {
 				if (unit.canBuildMine()) {
 					unit.buildMine();
-					return true;
+					return UnitAI.Result.InProgress;
```
KulkoBSW, line 125: see above

C7Engine/AI/UnitAI/WorkerAI.cs
```diff
@@ -118,25 +109,34 @@ bool UnitAI.PlayTurnImpl(Player player, MapUnit unit) {
...
-				return false;
+				if (unit.location.overlays.mine) {
+					return UnitAI.Result.Done;
+				}
+				return UnitAI.Result.Error;
 			} else if (workerMove == C7Action.UnitIrrigate) {
 				if (unit.canIrrigate()) {
 					unit.irrigate();
-					return true;
+					return UnitAI.Result.InProgress;
```
KulkoBSW, line 134: See above

C7Engine/AI/UnitAI/ExplorerAI.cs
```diff
@@ -10,207 +10,132 @@
...
-					return MoveToNextTileOnPath(explorerData, unit);
-				}
+			// If we're at the destination we're done.
+			if (unit.location == data.destination) {
+				player.tileKnowledge.aiExplorationTargets.Remove(data.destination);
+				return UnitAI.Result.Done;
 			}
-			return false;
+
+			return this.TryToMoveAlongPath(unit, data.pathToDestination, /*allowCombat=*/false);
```
KulkoBSW, line 45: From what I understand the following is true:
If that move reaches the final target of the exploration, the result is still in progress, Then next turn we are only realizing that we are done and wasting a complete turn formulating the next goal. especially if we are exploring neighboring tiles, that would slow down the explorer massively. 
I think at that point we should also check if we reached our destination already and adjusting the result accordingly.

C7Engine/AI/UnitAI/CombatAI.cs
```diff
@@ -19,14 +19,16 @@ public CombatAI(CombatAIData d) {
...
+			}
+
 			//Move along the path to the place where the unit plans to enter combat.
 			//This might initiate combat.
 			//Once there are more combat goals than run to a barbarian camp,
 			//we will need fancier algorithms
-			unit.path = combatAIData.path;
-			unit.moveAlongPath();
-			return true;
+			return this.TryToMoveAlongPath(unit, combatAIData.path, /*allowCombat=*/true);
```
KulkoBSW, line 31: Not really in scope of that PR:
So the combat AI is mmoving to a location and fights there? What if the goal is to destroy a unit and that unit has moved?

C7Engine/AI/UnitAI/WorkerAI.cs
TomWerner: It shouldn't be, the next turn around we'd discover it was done and move on to the next action.

C7Engine/AI/UnitAI/WorkerAI.cs
TomWerner: Ditto

C7Engine/AI/UnitAI/WorkerAI.cs
TomWerner: Ditto

C7Engine/AI/UnitAI/ExplorerAI.cs
TomWerner: PlayTurnImpl is called in a loop until movement points are exhausted, so no turns are wasted. Good question though!

In other words, if we return "in progress" and still have movement points left, we keep exploring.

C7Engine/AI/UnitAI/CombatAI.cs
TomWerner: The combat AI doesn't seem to actually be used, and will definitely need to be rewritten.

TomWerner:
Thanks Kulko!

PR #601, opened 2025-03-03
Pathing: Ensure we don't try to path through enemies

TomWerner:
Previously we acted as though we could walk through enemies and enemy cities, even if we were at peace with them. This resulted in exploring units getting stuck trying to walk past each other.


TomWerner:

PR #602, opened 2025-03-03
Implement a "fast forward" mode for observer mode

TomWerner:
This makes testing things easier, as we can quickly run through the first N turns of the game without having to repeatedly press the space bar.

To make this faster, I reduced the frequency at which we flush the logs to disk.


TomWerner:

PR #604, opened 2025-03-04
Stop the AI from repeatedly attempting to walk across oceans

TomWerner:
While working to speed up the AI turns after turn 100, I found that the AI was repeatedly attempting to send settlers to other continents by foot. Attempting this pathing is the worst case scenario for the pathing algorithms, as it has to check every tile on the continent to discover there is no route.

To solve this, we just add an identifier to each land mass, to make it trivial to exit early if no land route is possible. I'm fairly certain the base game did something similar, since the base game had an existing "continents" field for each tile.

To backfill continent information for the static save, I used this code snipped in `SaveTest.cs`:

```
		GameData gameData = ToGameData(saveNeverGameData);

		int nextContinent = 0;
		HashSet<Tile> seen = new();

		foreach (Tile t in gameData.map.tiles) {
			if (!t.IsLand()) {
				t.continent = -1;
				continue;
			} 

			if (seen.Contains(t)) {
				continue;
			}
			
			t.continent = nextContinent;
			seen.Add(t);

			Queue<Tile> toCheck = new();
			toCheck.Enqueue(t);

			while (toCheck.Count > 0) {
				Tile x = toCheck.Dequeue();
				x.continent = nextContinent;

				foreach (Tile n in x.neighbors.Values) {
					if (!seen.Contains(n) && n.IsLand()) {
						seen.Add(n);
						toCheck.Enqueue(n);
					}
				}
			}
			++nextContinent;
		}

		SaveGame saveWasGameData = SaveGame.FromGameData(gameData);
```

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -591,9 +591,9 @@ public partial class Game : Node2D {
 						}
 						SetAnimationsEnabled(false);
 						popupOverlay.ShowPopup(
-							new TextDialog("How many turns to fast forward through?", 
+							new TextDialog("How many turns to fast forward through?",
 											"Turns: ", "100",
-											(string turns) => { turnsLeftToFastForward = int.Parse(turns); }), 
+											(string turns) => { turnsLeftToFastForward = int.Parse(turns); }),
 								PopupOverlay.PopupCategory.Advisor);
 					} else {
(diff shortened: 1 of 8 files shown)


TomWerner:

PR #606, opened 2025-03-04
Rework the logic for chosing which cities to defend 

TomWerner:
This allows us to put more of it in `DefenderAI`, making `PlayerAI` less of a monolith, and it also allows us to improve the logic. The previous logic would never converge and just keep shuffling new units around. This logic is much more flexible and similar to the other ranking algorithms used for the other AI types, and takes into account things like defenders on their way. It will also allow us to improve the logic incrementally once we add wars.


TomWerner:

PR #608, opened 2025-03-05
CityProductionAI: Allow cities to force building defense when unguarded

TomWerner:
The first defensive unit they build will always become a defender, so we might as well coordinate between PlayerAI and CityProductionAI.

I plan to also use this logic to force cities to build escorts for their settlers if they don't have one in a future PR.


TomWerner:
This has been open for a week, submitting.

PR #609, opened 2025-03-05
Workerjobs as terraforms

KulkoBSW:
Implemented Workerjobs as Terraforms and used it for irrigation, Mining and Roadbuilding.

There are still a few unfinished parts like but I would like to have some feedback on the solution for now.

Open Todos
- Check for blocking between Workerjobs
- Choose the correct animation (currently everybody is irrigating)
- Implement the generic version for saving and loading
- Find out how to best test the WorkerAI
- Remove some old method not in use anymore.
- Edit: The Tooltext is still missing

### DIFF
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -546,17 +546,6 @@ namespace C7Engine {
 			return unit.unitType.actions.Contains(C7Action.UnitIrrigate) && unit.location.CanBeIrrigated(unit.owner);
 		}
 
-		// public static void irrigate(this MapUnit unit) {
-		// 	if (!unit.canIrrigate()) {
-		// 		log.Warning($"can't build irrigate by {unit}");
-		// 		return;
-		// 	}
-		//
-		// 	unit.WorkerJob = null;
-		// 	unit.animate(MapUnit.AnimatedAction.IRRIGATE, false, AnimationEnding.Repeat);
-		// 	unit.PerformBusyAction();
-		// }
-
 		public static void PerformTerraformAction(this MapUnit unit, string action) {
 			Terraform workerJob = EngineStorage.gameData.GetTerraformByAction(action);
 			unit.WorkerJob = workerJob;
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -18,13 +18,8 @@ namespace C7Engine {
 		private const int JOB_PROGRESS_WORKER = 2;
 		private const int JOB_PROGRESS_SLAVE = 1;
 
-		private static int GetWorkerJobCost(string workerJob) {
-			// TODO: Read from scenario info, and probably store this in a
-			// worker job class.
-			if (workerJob == C7Action.UnitIrrigate) {
-				return 8;
-			}
-			return 1;
+		private static int GetWorkerJobCost(Terraform workerJob) {
+			return workerJob.TurnsToComplete;
 		}
 
 		public static void animate(this MapUnit unit, MapUnit.AnimatedAction action, bool wait, AnimationEnding ending = AnimationEnding.Stop) {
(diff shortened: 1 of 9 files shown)


TomWerner:
I'm taking a look at the rest of the code but I'll address one point quickly.

> Find out how to best test the WorkerAI

If you automate the worker by pressing "a" or the automate button, that uses the worker AI. So you can use that to quickly see how things are working. Otherwise you can press "Ctrl+Alt+shift+O" (that's the letter O, not the number 0) to turn on observer mode, where you can see all the civs. Carthage (the default civ) does lots of mining, to the south the Aztecs have land that results in irrigation.

C7Engine/MapUnitExtensions.cs
```diff
@@ -551,13 +546,20 @@ public static bool canIrrigate(this MapUnit unit) {
...
 		}
 
-		public static void irrigate(this MapUnit unit) {
-			if (!unit.canIrrigate()) {
-				log.Warning($"can't build irrigate by {unit}");
-				return;
-			}
-
-			unit.WorkerJob = C7Action.UnitIrrigate;
+		// public static void irrigate(this MapUnit unit) {
```
TomWerner: Could you remove this commented out code?

C7Engine/MapUnitExtensions.cs
```diff
@@ -551,13 +546,20 @@ public static bool canIrrigate(this MapUnit unit) {
...
+		// 		return;
+		// 	}
+		//
+		// 	unit.WorkerJob = null;
+		// 	unit.animate(MapUnit.AnimatedAction.IRRIGATE, false, AnimationEnding.Repeat);
+		// 	unit.PerformBusyAction();
+		// }
+
+		public static void PerformTerraformAction(this MapUnit unit, string action) {
+			Terraform workerJob = EngineStorage.gameData.GetTerraformByAction(action);
```
TomWerner, line 545: Could you add the checks for "canIrrigate/canRoad/canMine"?

C7GameData/Save/SaveUnit.cs
```diff
@@ -42,11 +42,11 @@ public SaveUnit(MapUnit unit, GameMap map) {
 			experience = unit.experienceLevelKey;
 			movePointsRemaining = unit.movementPoints.remaining;
 			WorkerProgressTowardsJob = unit.WorkerProgressTowardsJob;
-			WorkerJob = unit.WorkerJobAsInt();
+			WorkerJob = WorkerJobAsInt(unit.WorkerJob);
```
TomWerner, line 45: Rather than an int, maybe make this an ID, then it just becomes `if (unit.WorkerJob != null) { WorkerJob = unit.WorkerJob.id; }`

TomWerner (changes requested):
Very cool!

C7Engine/MapUnitExtensions.cs
KulkoBSW: This checks are actually always done before that function is called, so I did not feel like we should do them twice.
This one replaces the irrigate() function, which also did not include the check.
But I think I should add a comment to point out that fact.

C7GameData/Save/SaveUnit.cs
KulkoBSW: Yes, Thtas the part I plan on doing next, but the day was long enough. This will later be the TerraformID.

TomWerner (approved):
This LGTM. There is a bug with WorkerAI, but it'll be fixed by the changes to `UnitAI.cs` in https://github.com/C7-Game/Prototype/pull/600/files.

I'm happy to have this go in in the meantime, since the bug is just that the AI workers finish things in one turn, just like before this PR. So it isn't a regression.

If you had time to glance over https://github.com/C7-Game/Prototype/pull/600/files I'd appreciate it, but if you don't have time, no worries. I'll ensure your PR goes in first so you don't have to deal with the merge.

TomWerner:
Let me know if you had anything else you wanted to push, otherwise I can merge this one now.

KulkoBSW:
I added the check if you can irrigate etc once more and now it is in a stable state again. Given the risk of future merge problems, as this touches a lot of files, I would appreciate if you would merge it.

TomWerner:
Thanks!

PR #610, opened 2025-03-05
Implement settler escorting for the AI

TomWerner:
The AI will now always move its settlers around with escorting defensive units, which become the initial garrison of the new city.


TomWerner:

PR #612, opened 2025-03-07
Remove the duplicate logic for confirm/cancel buttons in popups

TomWerner:
There's no need to keep copy-pasting the logic to new popups.


TomWerner:

PR #614, opened 2025-03-07
Kulkobsw/use terraform data for workerjobs

KulkoBSW:
The Workerjobs are now read dynamically from the save and I have switched to the ID as the value for our own save format.

I also set the correct animations for each Action.

### DIFF
--- a/C7/Text/c7-static-map-save.json
+++ b/C7/Text/c7-static-map-save.json
@@ -58686,7 +58686,6 @@
       "facingDirection": "north",
       "experience": "Regular",
       "workerProgressTowardsJob": 0,
-      "workerJob": -1,
       "isAutomated": false
     },
     {
@@ -58706,7 +58705,6 @@
       "facingDirection": "north",
       "experience": "Regular",
       "workerProgressTowardsJob": 0,
-      "workerJob": -1,
       "isAutomated": false
     },
     {
@@ -58726,7 +58724,6 @@
       "facingDirection": "north",
       "experience": "Regular",
       "workerProgressTowardsJob": 0,
-      "workerJob": -1,
       "isAutomated": false
     },
     {
@@ -58746,7 +58743,6 @@
       "facingDirection": "north",
       "experience": "Regular",
       "workerProgressTowardsJob": 0,
-      "workerJob": -1,
       "isAutomated": false
     },
     {
@@ -58766,7 +58762,6 @@
       "facingDirection": "north",
       "experience": "Regular",
       "workerProgressTowardsJob": 0,
-      "workerJob": -1,
       "isAutomated": false
     },
     {
@@ -58786,7 +58781,6 @@
       "facingDirection": "north",
       "experience": "Regular",
       "workerProgressTowardsJob": 0,
-      "workerJob": -1,
       "isAutomated": false
     },
     {
@@ -58806,7 +58800,6 @@
       "facingDirection": "north",
       "experience": "Regular",
       "workerProgressTowardsJob": 0,
-      "workerJob": -1,
       "isAutomated": false
     },
     {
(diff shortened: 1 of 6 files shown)


TomWerner:
Very nice, thank you! Could you add the new animations to this list so that the worker doesn't remain incorrectly selected once it starts the move? https://github.com/C7-Game/Prototype/blob/7eb595356c5ad1340d920937c5e922296677fa15/C7GameData/MapUnit.cs#L123

TomWerner (approved):

KulkoBSW:
> Very nice, thank you! Could you add the new animations to this list so that the worker doesn't remain incorrectly selected once it starts the move?
> 
> https://github.com/C7-Game/Prototype/blob/7eb595356c5ad1340d920937c5e922296677fa15/C7GameData/MapUnit.cs#L123

Why do we control this via the animation? I was wondering why it is not working correctly but assumed I need to send a new mesage to the ui to Select the next unit.

TomWerner:
> > Very nice, thank you! Could you add the new animations to this list so that the worker doesn't remain incorrectly selected once it starts the move?
> > https://github.com/C7-Game/Prototype/blob/7eb595356c5ad1340d920937c5e922296677fa15/C7GameData/MapUnit.cs#L123
> 
> Why do we control this via the animation? I was wondering why it is not working correctly but assumed I need to send a new mesage to the ui to Select the next unit.

It's only kind of controlled by animation. I think this is specifically so that we can observe the results of a unit fighting another unit - we wouldn't want to switch to a new unit while the battle animation was still going on.

KulkoBSW:
Assuming I want to doku what all is to be chaged for the otehr 10 terraform actions, Where could I do this? In the Wiki? Or in a discussion? or in the Civfanatics board somewhere? Because I will not remember all this locations when I add the next once in half a year.

TomWerner:
I agree this isn't great. You could also look into some way to make it branch based on whether we're using `AnimationEnding.Repeat` - that's probably more maintainable long term.

KulkoBSW:
@TomWerner Two more things I found:
- I checked in Civ III and they do not stop you from sending one unit to mine and the other to irrigate. Whoever finishes first wins and then both workers stop working on there project.
They do however stop you from starting an irrigation once a mine is already finished, asking if you really want to go ahead.

And second Your refactoring with the performbusyAction broke that strange behavior how of CIV III concerning multiple workers on a job. When I have 2 workers starting a road at the same turn, only 1 of them is needed to finish it. However the game (in Single Player) first calculates the progress for each unit spending all 4 actions. And then checks if it is finished. So both workers are spend in that scenario. 
However in C8 1 of the workers becomes available after the other finished. 
MY ID to FIX this (or better to recreate that Bug) would be to Take the check for finish out of performbusy action again and check after performing all busyactions.
 
Anyway I think the current state of the PR could be merged again, when everything is fine.

TomWerner:
> I checked in Civ III and they do not stop you from sending one unit to mine and the other to irrigate. Whoever finishes first wins and then both workers stop working on there project.

Ah, I had forgotten about that. Good catch.

> broke that strange behavior how of CIV III concerning multiple workers on a job. When I have 2 workers starting a road at the same turn, only 1 of them is needed to finish it. However the game (in Single Player) first calculates the progress for each unit spending all 4 actions. And then checks if it is finished. So both workers are spend in that scenario.

This is actually the same behavior as civ3 *if* you use the "road to" command ("Ctrl+r"). Its always been a weird bit of trivia in the base game that using "road to" was more efficient. I think the behavior we have right now is pretty reasonable, and less weird than the base game.

> Anyway I think the current state of the PR could be merged again, when everything is fine.

Agreed, I'll merge it!

PR #615, opened 2025-03-07
Simplify the process for ignoring uninteresting animations

TomWerner:
This avoids the need to keep adding animations to the list of conditions we `||` on.


KulkoBSW (approved):
Looks good

PR #616, opened 2025-03-11
WorkerAI: Stop wasting worker moves on unowned tiles

TomWerner:
We still allow roading unowned tiles (for road networks) but this prevents mining other civ's tiles for them.

### DIFF
--- a/C7Engine/AI/UnitAI/WorkerAI.cs
+++ b/C7Engine/AI/UnitAI/WorkerAI.cs
@@ -78,6 +78,15 @@ namespace C7Engine {
 				return null;
 			}
 
+			// Don't waste worker moves on unowned tiles. We do allow roading
+			// unowned tiles though.
+			if (t.owningCity == null || t.owningCity.owner != player) {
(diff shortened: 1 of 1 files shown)


C7Engine/AI/UnitAI/WorkerAI.cs
```diff
@@ -78,6 +78,15 @@ UnitAI.Result UnitAI.PlayTurnImpl(Player player, MapUnit unit) {
 				return null;
 			}
 
+			// Don't waste worker moves on unowned tiles. We do allow roading
+			// unowned tiles though.
+			if (t.owningCity == null || t.owningCity.owner != player) {
```
KulkoBSW, line 83: So the AI would road my cities for me? Is that intentional?

TomWerner: Only if an enemy city was in between two of your cities. We could (and probably will) eventually need a custom pathing approach for worker paths that would route around other territory when roading, since right now the worker also doesn’t know to take the least cost roading route.

PR #617, opened 2025-03-12
Fix the problem of excessive boat production for exploring 

TomWerner:
Right now the only thing we use boats for is exploring the oceans, so once we've finished exploring the oceans we should stop sending out more boats. I noticed this while using observer mode on some random saves I had laying around that had more coastal cities.

I also updated the explorer logic a bit, to track whether we've finished exploring the oceans, and also to check whether the tile we're headed to has already been explored. If it has, we reroute.


esbudylin (approved):

TomWerner:
Thanks!

PR #618, opened 2025-03-12
StrategicPriority: Strip out some unused code and make WarPriority aware of wars

TomWerner:
The war priority logic doesn't do much right now, so let's make that clearer by removing unused fields like `properties`. We don't need to encourage more stringly-typed fields.


esbudylin (approved):

TomWerner:
Thanks!

PR #619, opened 2025-03-12
Implement a basic handling of wars, offensively and defensively 

TomWerner:
We now achieve a minimal defensive position and then allow offensive units to attack. The attackers have some basic prioritization (intercept other attackers, take cities, focus on closer targets, etc).

In practice this results in two AIs at war smashing into each other with a large stack, and then singleton attacks afterwards, but it's a start.

We also haven't implemented peace treaties yet, so once at war civs are always at war, and will stop building settlers and such. I'll add that in a new PR.

This was tested by running in observer mode for ~100 turns, declaring war by doing a manual attack, and then running observer mode again. We don't yet have AIs declaring war themselves.

### DIFF
--- a/C7Engine/AI/PlayerAI.cs
+++ b/C7Engine/AI/PlayerAI.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using C7Engine.Pathing;
 using C7GameData;
+using C7GameData.Save;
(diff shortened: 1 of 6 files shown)


esbudylin (approved):

TomWerner:
Thanks!

PR #620, opened 2025-03-12
Enable complete fog of war (unknown & active tiles)

rchoffardet:
This PR implements the full fog of war. This supports both unknown and inactive tiles.
image

### DIFF
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -513,7 +513,6 @@ public partial class LooseView : Node2D {
 
 			if (!gD.observerMode) {
 				foreach (FogOfWarLayer layer in layers.Where(layer => layer is FogOfWarLayer).Cast<FogOfWarLayer>()) {
-					layer.ComputeTileKnowledge(gD);
 					for (int Y = visRegion.upperLeftY; Y < visRegion.lowerRightY; Y++)
 						if (gD.map.isRowAt(Y))
 							for (int X = visRegion.getRowStartX(Y); X < visRegion.lowerRightX; X += 2) {
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -531,7 +531,7 @@ public partial class LooseView : Node2D {
 			return true;
 		}
 		TileKnowledge knowledge = gameDataAccess.gameData.GetHumanPlayers()[0].tileKnowledge;
-		return tile != Tile.NONE && (knowledge.isTileKnown(tile));
+		return tile != Tile.NONE && knowledge.isTileKnown(tile);
 	}
 }
 
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -512,7 +512,8 @@ public partial class LooseView : Node2D {
 			}
 
 			if (!gD.observerMode) {
-				foreach (LooseLayer layer in layers.FindAll(layer => layer is FogOfWarLayer)) {
+				foreach (FogOfWarLayer layer in layers.Where(layer => layer is FogOfWarLayer).Cast<FogOfWarLayer>()) {
+					layer.ComputeTileKnowledge(gD);
 					for (int Y = visRegion.upperLeftY; Y < visRegion.lowerRightY; Y++)
 						if (gD.map.isRowAt(Y))
 							for (int X = visRegion.getRowStartX(Y); X < visRegion.lowerRightX; X += 2) {
@@ -531,7 +532,7 @@ public partial class LooseView : Node2D {
 			return true;
 		}
 		TileKnowledge knowledge = gameDataAccess.gameData.GetHumanPlayers()[0].tileKnowledge;
-		return tile != Tile.NONE && (knowledge.isTileKnown(tile) || knowledge.isBorderOfTileKnowlege(tile));
+		return tile != Tile.NONE && (knowledge.isTileKnown(tile));
 	}
 }
 
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -413,9 +413,9 @@ namespace C7Engine {
 				float movementCost = TilePath.getMovementCost(unit.location, dir, newLoc);
 				if (!unit.location.unitsOnTile.Remove(unit))
 					throw new System.Exception("Failed to remove unit from tile it's supposed to be on");
-				unit.OnEnterTile(newLoc);
 				newLoc.unitsOnTile.Add(unit);
 				unit.location = newLoc;
+				unit.OnEnterTile(newLoc);
 				unit.movementPoints.onUnitMove(movementCost);
 				unit.animate(MapUnit.AnimatedAction.RUN, wait);
 			}
--- a/C7Engine/MapUnitExtensions.cs
+++ b/C7Engine/MapUnitExtensions.cs
@@ -413,6 +413,7 @@ namespace C7Engine {
 				float movementCost = TilePath.getMovementCost(unit.location, dir, newLoc);
 				if (!unit.location.unitsOnTile.Remove(unit))
 					throw new System.Exception("Failed to remove unit from tile it's supposed to be on");
+				// Make sure the unit is on the new location before claiming we have entered the tile
 				newLoc.unitsOnTile.Add(unit);
 				unit.location = newLoc;
 				unit.OnEnterTile(newLoc);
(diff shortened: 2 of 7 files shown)


C7GameData/AIData/TileKnowledge.cs
```diff
@@ -59,5 +68,12 @@ public List<Tile> AllKnownTiles() {
 			}
 			return list;
 		}
+
+		private void RecomputeActiveTiles() {
+			activeTiles = _player.cities
+				.SelectMany(x => x.GetTilesWithinBorders().SelectMany(y => y.neighbors.Values).Append(x.location))
+				.Concat(_player.units.SelectMany(x => x.location.neighbors.Values.Append(x.location)))
+				.ToHashSet();
+		}
```
rchoffardet: This method is fairly unoptimized but should be okay for a first version.

C7/MapView.cs
```diff
@@ -518,7 +518,7 @@ private static bool IsTileKnown(Tile tile, UIGameDataAccess gameDataAccess) {
 			return true;
 		}
 		TileKnowledge knowledge = gameDataAccess.gameData.GetHumanPlayers()[0].tileKnowledge;
-		return tile != Tile.NONE && (knowledge.isTileKnown(tile) || knowledge.isBorderOfTileKnowlege(tile));
+		return tile != Tile.NONE && (knowledge.isTileKnown(tile));
```
TomWerner: Could you remove the unneeded parenthesis here?

C7Engine/MapUnitExtensions.cs
```diff
@@ -394,9 +394,9 @@ public static bool move(this MapUnit unit, TileDirection dir, bool wait = false)
 				float movementCost = TilePath.getMovementCost(unit.location, dir, newLoc);
 				if (!unit.location.unitsOnTile.Remove(unit))
 					throw new System.Exception("Failed to remove unit from tile it's supposed to be on");
-				unit.OnEnterTile(newLoc);
 				newLoc.unitsOnTile.Add(unit);
 				unit.location = newLoc;
+				unit.OnEnterTile(newLoc);
```
TomWerner, line 419: Is this reordering important?

C7GameData/AIData/TileKnowledge.cs
TomWerner: Unfortunately it seems like this introduces a pretty significant regression to the unit test runtime, when run locally (the tests that get slow are the scenario loading ones, and those require a civ3 installation).

At least on my machine, `dotnet test ../C7GameDataTests/C7GameDataTests.csproj` (from the C7 directory) went from taking around 8 seconds to 70 seconds.

Could this be computed more easily in `isActiveTile`? I think if the logic lived there it would look like

```
foreach (Tile neighbor in t.neighbors.Values) {
  // A tile is active if it's next to a tile owned by a city.
  if (neighbor.owningCity != null && neighbor.owningCity.owner == _player) {
    return true;
  }

  // A tile is active if it's next to a tile with an owner on it.
  if (neighbor.unitsOnTile.Count > 0 && neighbor.unitsOnTile[0].owner == _player) {
    return true;
  }
}
return false;
```

Failing that we'd probably have to do something gross with OnEnterTile (and a corresponding OnLeaveTile or something)

TomWerner (changes requested):
This looks great! Visually it's a huge improvement.

I do think we'll need to fix the performance regression though, that big of a test slowdown is unfortunate.

C7Engine/MapUnitExtensions.cs
rchoffardet: Yes, the unit needs to be at the new location when the event is called so that the active tiles are correctly updated

TomWerner: It doesn't look like `unit.location` is used in that method (weirdly...). Unless it's some indirect method call I'm missing I think only the tile argument matters - the call is `unit.owner.tileKnowledge.AddTilesToKnown(tile)`.

That said, I think the previous ordering was weird and confusing. Could you leave a comment, something like `// Make sure the unit is on the new location before claiming we have entered the tile`.

rchoffardet: The recompute is made in `AddTilesToKnown` method.

I will add the comment 👍

rchoffardet:
@TomWerner, I cannot reply in the thread. So here is my reply about the performance issue during save load.

This is the tradeoff. Either it's made while moving the unit or while drawing the map. The map is drawn multiple times per second. So I thought the choice was obvious. I guess the best way to deal with that is to only update active tiles at the end of save loading. 

There are also more optimized methods such as using and combining bitmap indexes. But I'd use it as a last resort.

TomWerner:
> @TomWerner, I cannot reply in the thread. So here is my reply about the performance issue during save load.
> 
> This is the tradeoff. Either it's made while moving the unit or while drawing the map. The map is drawn multiple times per second. So I thought the choice was obvious. I guess the best way to deal with that is to only update active tiles at the end of save loading.
> 
> There are also more optimized methods such as using and combining bitmap indexes. But I'd use it as a last resort.

Yeah, github's support for threads leaves a lot to be desired.

So I personally care about the unit test being relatively quick. I'm already a bit sad that I had to slow it down by running a headless game for 30 turns, but that's already caught a couple of bugs.

I realized we could combine the approaches - this version of the method gets the test runtime down to ~13 seconds.

```
		private void RecomputeActiveTiles() {
			activeTiles.Clear();
			foreach (Tile t in knownTiles) {
				if (activeTiles.Contains(t)) {
					continue;
				}

				// A tile within a city's borders and all of its neighbors are active.
				if (t.owningCity != null && t.owningCity.owner == _player) {
					activeTiles.Add(t);

					foreach (Tile neighbor in t.neighbors.Values) {	
						activeTiles.Add(neighbor);
					}
					continue;
				}

				// A tile with a unit on it and all of its neighbors are active.
				if (t.unitsOnTile.Count > 0 && t.unitsOnTile[0].owner == _player) {
					activeTiles.Add(t);

					foreach (Tile neighbor in t.neighbors.Values) {	
						activeTiles.Add(neighbor);
					}
					continue;
				}
			}
		}
```

There's still a bug with the "new game" method, the tiles around the player aren't properly active.

image

I think that's because the tile knowledge is present in the canned save, so we never recompute the active tiles on load.

rchoffardet:
@TomWerner, sorry for the delay 😬 

>There's still a bug with the "new game" method, the tiles around the player aren't properly active.

I fixed it this evening :)

> I realized we could combine the approaches - this version of the method gets the test runtime down to ~13 seconds.

While fixing the bug above, this method proved to be buggy. Sometimes when moving on the edge on the active tiles, the tile activity does not always update. Is the slow down still bearable?

C7Engine/MapUnitExtensions.cs
rchoffardet: fixed :)

C7/MapView.cs
rchoffardet: fixed

TomWerner:
> While fixing the bug above, this method proved to be buggy. Sometimes when moving on the edge on the active tiles, the tile activity does not always update. Is the slow down still bearable?

Do you have a way to reproduce the issue? I clicked around for a bit with this PR patched in and everything looked ok to me - I'm happy to merge it as-is.

TomWerner:
(thanks for the fixes!)

rchoffardet:
> Do you have a way to reproduce the issue? I clicked around for a bit with this PR patched in and everything looked ok to me - I'm happy to merge it as-is.
it's fixed now, but with the original version of the method you provided:
```
private void RecomputeActiveTiles() {
	activeTiles.Clear();
	foreach (Tile t in knownTiles) {
		if (activeTiles.Contains(t)) {
			continue;
		}

		// A tile within a city's borders and all of its neighbors are active.
		if (t.owningCity != null && t.owningCity.owner == _player) {
			activeTiles.Add(t);

			foreach (Tile neighbor in t.neighbors.Values) {	
				activeTiles.Add(neighbor);
			}
			continue;
		}

		// A tile with a unit on it and all of its neighbors are active.
		if (t.unitsOnTile.Count > 0 && t.unitsOnTile[0].owner == _player) {
			activeTiles.Add(t);

			foreach (Tile neighbor in t.neighbors.Values) {	
				activeTiles.Add(neighbor);
			}
			continue;
		}
	}
}
```
this part caused the issue:
```
		if (activeTiles.Contains(t)) {
			continue;
		}
```
So I removed it. Sadly it was also a great way to improve test execution speed

TomWerner:
Ah, the tests were still pretty quick even without that - somewhere around 17 seconds on my rather slow machine. So if you're happy with this I'll merge!

TomWerner (approved):

rchoffardet:
fine by me :)

PR #623, opened 2025-03-16
Determine producible units

esbudylin:
This PR implements logic to determine producible units. The new method considers required technology, unit obsolescence (e.g., Warrior cannot be produced if Swordsman is already available), and whether a unit is unique.

**CIV3 approach**

The approach to storing data on unique units and upgrades in this PR is quite different from how CIV3 handles it.

In CIV3, unit availability is tracked separately for each civilization. So, making a unit unique requires modifying both the unit itself (e.g., Mounted Warrior) and the unit it replaces (e.g., Horseman). Additionally, CIV3 includes unique units in upgrade data, leading to convoluted upgrade chains. For example, the game represents the Horseman's upgrade path as follows:

Horseman->Mounted Warrior->Three-Man Chariot->Knight->Keshik->Ansar Warrior->Rider->Samurai->War Elephant->Cavalry

**This PR's approach**

CIV3’s method makes unit data complex and difficult to modify. Instead of replicating it, this PR introduces a nested `Unique` class within `UnitPrototype`. This class contains two fields:

- civilization: The civilization that has access to the unique unit.
- replace: The standard unit that this unique unit replaces.

Each UnitPrototype has a `unique` field of type `Unique`. It directly links unique units to their replacements. This structure also allows filtering out unique units from upgrade chains.

**Limitations**

This approach has some limitations:

- It generally works well for the base ruleset of the game, with the exception of Three-Man Chariot (Hittite UU). In CIV3, Hittites can't build either Horseman or Chariot and can only build their UU instead. This appears to be the only case where a UU replaces not one, but two units.
- It doesn't cover ruleset of scenarios, where certain units are accessible by multiple civilizations (e.g., in the Middle Ages scenario only Byzantines and Rus can build Cataphract).

### DIFF
--- a/C7GameData/City.cs
+++ b/C7GameData/City.cs
@@ -60,13 +60,34 @@ namespace C7GameData {
 		}
 
 		public bool CanBuildUnit(UnitPrototype proto) {
-			List<string> allowedUnits = new List<string> {"Warrior", "Chariot", "Settler", "Worker", "Catapult", "Galley"};
-			if (!allowedUnits.Contains(proto.name))
+			return MeetsProductionRequirements(proto) && !IsUnitObsolete(proto);
+		}
+
+		private bool IsUnitObsolete(UnitPrototype proto) {
+			var upgradeChain = owner.civilization.GetUpgradeChain(proto);
+
+			return upgradeChain.Any(MeetsProductionRequirements);
+		}
+
+		private bool HasRequiredTechnology(IProducible producible) {
+			return producible.requiredTech == null ||
+				   owner.knownTechs.Contains(producible.requiredTech.id);
+		}
+
+		private bool MeetsProductionRequirements(UnitPrototype proto) {
+			if (!owner.civilization.IsUnitAvailable(proto)) {
 				return false;
-			if (proto.categories.Contains("Sea"))
-				return location.NeighborsWater();
-			else
-				return true;
+			}
+
+			if (!HasRequiredTechnology(proto)) {
+				return false;
+			}
+
+			if (proto.categories.Contains("Sea") && !location.NeighborsWater()) {
+				return false;
+			}
+
+			return true;
 		}
 
 		public int TurnsUntilGrowth() {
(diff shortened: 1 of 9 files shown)


C7GameData/City.cs
```diff
@@ -60,13 +60,34 @@ public bool IsCapital() {
...
 		public bool CanBuildUnit(UnitPrototype proto) {
-			List<string> allowedUnits = new List<string> {"Warrior", "Chariot", "Settler", "Worker", "Catapult", "Galley"};
-			if (!allowedUnits.Contains(proto.name))
+			return MeetsProductionRequirements(proto) && !IsUnitObsolete(proto);
+		}
+
+		private bool IsUnitObsolete(UnitPrototype proto) {
+			var upgradeChain = owner.civilization.GetUpgradeChain(proto);
+
+			return upgradeChain.Any(MeetsProductionRequirements);
```
TomWerner, line 72: It might be worth leaving a TODO that we also need to consider golden ages here - until a civ has had its golden age it can always produce the unique unit, even if it would otherwise be obsolete.

C7GameData/City.cs
```diff
@@ -60,13 +60,34 @@ public bool IsCapital() {
 		}
 
 		public bool CanBuildUnit(UnitPrototype proto) {
-			List<string> allowedUnits = new List<string> {"Warrior", "Chariot", "Settler", "Worker", "Catapult", "Galley"};
-			if (!allowedUnits.Contains(proto.name))
+			return MeetsProductionRequirements(proto) && !IsUnitObsolete(proto);
+		}
+
+		private bool IsUnitObsolete(UnitPrototype proto) {
+			var upgradeChain = owner.civilization.GetUpgradeChain(proto);
```
TomWerner: Could you spell this type out?

TomWerner (approved):
Very nice! I'm curious how this will interact with the combat AI stuff I've been playing with once the AI has more options for building units. Should be interesting!

esbudylin:
@TomWerner Thanks for the review! I've addressed your comments in the latest commit.

PR #626, opened 2025-03-20
RightClickMenu: Implement the "contact" button

TomWerner:


C7/UIElements/RightClickMenu.cs
```diff
@@ -178,13 +178,21 @@ public void ResetItems(Tile tile, Dictionary<ID, bool> uiUpdatedUnitStates = nul
...
+					game.controller.EnsureRelationshipExists(nonPlayerUnits[0].owner);
+					game.OnDiplomacySelected(new ParameterWrapper<ID>(nonPlayerUnits[0].owner.id));
+				});
 			} else {
 				// TODO: This isn't necessarily the top unit, get that code to an accessible
 				// location and then use it here.
 				MapUnit unit = nonPlayerUnits[0];
 				AddItem($"{unit.owner.civilization.noun} {unit.Describe()}", null);
-				AddItem($"Contact {unit.owner.civilization.name}", null);
+				AddItem($"Contact {unit.owner.civilization.name}",  () => {
```
esbudylin: The logic for the contact button is the same for both conditional branches. Could it be moved out of if/else to eliminate code duplication?

esbudylin (approved):

C7/UIElements/RightClickMenu.cs
TomWerner: Good call, done.

TomWerner:
Thanks for the review!

PR #627, opened 2025-03-20
Add peace treaties to the list of things we can do in deals 

TomWerner:
This doesn't yet assign a cost in gold to peace treaties, but now there is away to end player-started wars.

#571

image

image

image

image


esbudylin (approved):

TomWerner:
Thanks!

PR #628, opened 2025-03-22
Optimize city drawing

esbudylin:
This commit removes unnecessary redrawing of city scenes on each frame. Instead, it redraws them only when production or citizen assignment changes, or at the beginning of the turn.

This change improves the frame rate when scrolling the map with a large number of cities.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -213,6 +213,7 @@ public partial class Game : Node2D {
 					}
 					break;
 				case MsgStartTurn mST:
+					mapView.cityLayer.RedrawCities();
 					OnPlayerStartTurn();
 					break;
 				case MsgCityCreated mCC:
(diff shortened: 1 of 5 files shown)


TomWerner:
Hmm, I'm not sure how I feel about this.

On the one hand, yay better performance. On the other, we introduce some inaccuracy to the UI. We'd need to be sure to redraw cities for all sorts of other events, like any worker actions completing (right now a completed mine or irrigation will sometimes result in the city scene updating, but after this PR it doesn't). Eventually that set also includes:
 - pillaging
 - enemies standing on terrain
 - pollution
 - bombardment
 - government changes
 - selling buildings (like factories)
 - etc

Do you know why cities in particular are so expensive to draw?

Should these lines be running every frame?
 - https://github.com/C7-Game/Prototype/blob/b0129138cfff1320eb8599bac17d9de28cad755b/C7/Map/CityLabelScene.cs#L113
 - https://github.com/C7-Game/Prototype/blob/b0129138cfff1320eb8599bac17d9de28cad755b/C7/Map/CityLabelScene.cs#L126-L149

Or should we store the last value of those strings the way we do the [label width](https://github.com/C7-Game/Prototype/blob/b0129138cfff1320eb8599bac17d9de28cad755b/C7/Map/CityLabelScene.cs#L103) and only update them if needed?

esbudylin:
@TomWerner Thanks for the comment! I think you are right, there are a lot of edge cases which I haven't considered, and it would be better to optimize the rendering of the city scene instead.  I'll look into it.

esbudylin:
continued in #633

PR #633, opened 2025-03-26
Rewrite CityLabelScene with Godot UI nodes

esbudylin:
This PR ports the custom drawing logic in the CityLabelScene class to Godot UI nodes. The new approach improves city rendering performance and makes code easier to maintain. It also solves some minor visual issues with text centering.

Before: 
image_2025-03-24_23-36-40

After:
image


TomWerner (approved):
Very nice, thank you!

PR #638, opened 2025-03-28
Load initial civilization goverments from the scenario files

esbudylin:
Recent changes to the government system broke local scenario-related tests because players weren't assigned governments when loading a scenario file. This PR fixes that.


TomWerner (approved):
Whoops, thanks!

PR #639, opened 2025-03-29
Open city screen on double-click

esbudylin:
This PR makes it possible to open the city screen by double-clicking on a city. 

It introduces a new class for detecting double-clicks. This new class was necessary because, although Godot provides built-in double-click detection, it does not allow handling single clicks separately. With Godot’s built-in detection, clicking on a city that contains a unit would both zoom to the city and select the unit.

To differentiate between single and double clicks the new class uses a timer-based detection. When a click occurs, the class waits for a short period (0.2 seconds) to check if another click follows. If a second click happens, it emits a double-click signal; otherwise, it emits a single-click signal. The only downside of using it is a small delay when single-clicking on a city to select a unit.

Additionally, this PR includes a refactor of the `_UnhandledInput` method in the `Game` class, splitting it into multiple smaller methods.

### DIFF
--- a/C7/DoubleClickHandler.cs
+++ b/C7/DoubleClickHandler.cs
@@ -1,5 +1,19 @@
 using Godot;
 
+/// <summary>
+/// A class for detecting double-clicks.
+///
+/// To differentiate between single and double clicks, this class uses
+/// a timer-based detection mechanism. When a click occurs, it waits
+/// for a short period (DOUBLE_CLICK_DELAY) to check if another click
+/// follows. If a second click happens, it emits a double-click
+/// signal; otherwise, it emits a single-click signal.
+///
+/// Although Godot provides built-in double-click detection, it does
+/// not allow handling single clicks separately. For example, with
+/// Godot’s built-in detection, clicking on a city that contains a
+/// unit would both zoom to the city and select the unit.
+/// </summary>
 [GlobalClass]
 public partial class DoubleClickHandler : Node {
 	int leftMouseButtonClickCount = 0;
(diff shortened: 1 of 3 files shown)


C7/DoubleClickHandler.cs
```diff
@@ -0,0 +1,38 @@
+using Godot;
+
+[GlobalClass]
+public partial class DoubleClickHandler : Node {
```
TomWerner, line 18: Thanks for the great details on why this class is necessary in the PR description, that was really informative. Could you summarize that as a class comment as well? That'll help with new devs in the future that are less familiar with Godot.

C7/Game.cs
```diff
@@ -479,135 +469,228 @@ public override void _UnhandledInput(InputEvent @event) {
...
+		}
+	}
+
+	private void HandleRightClickOnTile(Tile tile, InputEventMouseButton eventMouseButton) {
+		bool shiftDown = Input.IsKeyPressed(Godot.Key.Shift);
+
+		// Handle the shortcut of shift+right clicking a city to get the change production menu.
+		if (shiftDown && tile.cityAtTile?.owner == controller)
+			new RightClickChooseProductionMenu(this, tile.cityAtTile).Open(eventMouseButton.Position);
+		else if ((!shiftDown) && tile.unitsOnTile.Count > 0)
```
TomWerner: Are the extra parenthesis around `!shiftDown` needed, here and on the next `else if` branch?

TomWerner (approved):
This is great, thank you! The improvements to `Game.cs` are particularly nice, that method was a giant mess.

esbudylin:

PR #641, opened 2025-03-30
Make the war priority semi-useful, so the AI declares war

TomWerner:
Still no peace treaties, but the AI will now start wars, preferring to do so against civs it is near, and against civs it is stronger than.


esbudylin (approved):

TomWerner:
Thanks!

PR #644, opened 2025-04-01
Handle era advancement and allow viewing other era in the advisor

TomWerner:
Screenshots!

image

image

image

image

### DIFF
--- a/C7/UIElements/Advisors/ScienceAdvisor.cs
+++ b/C7/UIElements/Advisors/ScienceAdvisor.cs
@@ -9,17 +9,25 @@ public partial class ScienceAdvisor : TextureRect {
 	private ImageTexture MiddleBackground;
 	private ImageTexture IndustrialBackground;
 	private ImageTexture ModernBackground;
+	private TextureButton nextEra;
+	private TextureButton previousEra;
+	private List<TechBox> techBoxes = new();
+
+	// Stored separately so we can modify this without mutating the player.
+	private string eraName;
 
 	// Called when the node enters the scene tree for the first time.
 	public override void _Ready() {
 		this.CreateUI();
-		this.DrawTechTree();
+
+		using UIGameDataAccess gameDataAccess = new();
+		List<Tech> techs = gameDataAccess.gameData.techs;
+		Player player = gameDataAccess.gameData.GetHumanPlayers()[0];
+		eraName = player.eraCivilopediaName;
+		this.DrawTechTree(eraName, player, techs);
 	}
 
 	private void CreateUI() {
-		// TODO: support other eras. Amusingly the dependency arrows are drawn
-		// on this texture, so to render prereqs we need to update this per era.
-		//
 		// science_industrial_new is used as the industrial tech tree is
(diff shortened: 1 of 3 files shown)


esbudylin:
I've found an issue with the tech layout. If you first load a regular save, then return to the menu and load a scenario (for example, the WWII one), it doesn't update arrows in the tech tree.
image

C7GameData/Player.cs
```diff
@@ -281,15 +297,13 @@ public int EstimateTurnsToResearch(Tech tech) {
 			int remainingCost = tech.TechCostFor(this) - beakers;
 			int turnsRemaining = (int)Math.Ceiling((double)remainingCost / beakersPerTurn);
 
-			// We never spend more than 50 turns per tech.
+			// We never spend more than 50 turns per tech or less than 4.
 			int maxTurnsRemaining = 50 - turnsResearched;
+			int minTurnsRemaining = 4 - turnsResearched;
```
esbudylin, line 302: Could you declare these values (50 and 4) as constants?

C7/UIElements/Advisors/ScienceAdvisor.cs
```diff
@@ -59,56 +67,107 @@ private void CreateUI() {
...
+		HashSet<ID> knownTechs = player.knownTechs;
+		previousEra.Show();
+		nextEra.Show();
 
-			foreach (Tech tech in techs) {
-				if (tech.EraCivilopediaName != player.eraCivilopediaName) {
-					continue;
-				}
+		// Set the tech background based on the player's era.
+		if (eraName == "ERAS_Ancient_Times") {
```
esbudylin, line 114: Do Civ3 files always have the same set of 4 eras? If so, could we convert them to an Enum during loading to avoid string-to-index conversions in the game logic? Also, as a future TODO, we might want to support an arbitrary number of eras.

TomWerner:
> I've found an issue with the tech layout. If you first load a regular save, then return to the menu and load a scenario (for example, the WWII one), it doesn't update arrows in the tech tree. image

Take a look at https://github.com/C7-Game/Prototype/pull/653, it should fix it.

I'll try to address the other comments today.

C7GameData/Player.cs
TomWerner: I realized these should really be imported from the BIQ, so I did that in https://github.com/C7-Game/Prototype/pull/654.

 Sorry for sticking it at the end of the stack, I didn't want to mess with merge conflicts on the other importing I did.

C7/UIElements/Advisors/ScienceAdvisor.cs
TomWerner: > Do Civ3 files always have the same set of 4 eras?

As far as I can tell, yes. But I agree about handling an arbitrary number of eras. I opened https://github.com/C7-Game/Prototype/issues/655 to track this idea - in particular it might be nice to have the era->pcx mapping in the JSON file rather than hardcoded.

esbudylin (approved):

PR #650, opened 2025-04-04
Show built improvements, refactor the city scene

esbudylin:
This PR adds a list of built improvements to the city screen.

It also overhauls the city screen scene, utilizing [Godot's ability](https://docs.godotengine.org/en/stable/tutorials/plugins/running_code_in_the_editor.html) to execute code in the editor. Now, all Civ3 assets necessary for the city screen are loaded directly into the editor, and logic for positioning the nodes is moved from the C# class to the scene file.

Additionally, this PR fixes two bugs:
- A graphical bug where black sidebars in the city scene sometimes didn't cover the entire screen.
- An issue with the trading screen where it didn't load properly when there were no tradable techs.

image
image

### DIFF
--- a/C7/UIElements/CityScreen/CityScreen.cs
+++ b/C7/UIElements/CityScreen/CityScreen.cs
@@ -26,7 +26,7 @@ public partial class CityScreen : Control {
 	[Export] private TextureButton productionButton;
 	[Export] private TextureButton close;
 
-	private ProductionMenu productionMenu;
+	[Export] private ProductionMenu productionMenu;
 
 	Theme yieldDetailsFontTheme = new();
 	FontFile yieldDetailsFont = new();
@@ -225,11 +225,7 @@ public partial class CityScreen : Control {
 	}
 
 	private void OnExit() {
-		if (productionMenu != null) {
-			productionMenu.QueueFree();
-			productionMenu = null;
-		}
-
+		productionMenu.Hide();
 		tileAssignmentLayer.city = null;
 	}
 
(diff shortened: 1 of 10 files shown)


C7/UIElements/CityScreen/CityScreen.cs
```diff
@@ -256,6 +221,25 @@ private void OnShowCityScreen(ParameterWrapper<City> city) {
...
+		tileAssignmentLayer.city = null;
+	}
+
+	private void RenderExistingBuildings(List<CityBuilding> buildings) {
+		foreach (var node in existingBuildings.GetChildren()) {
+			existingBuildings.RemoveChild(node);
+		}
+
+		foreach (CityBuilding building in buildings) {
+			Label label = new() {
```
TomWerner, line 238: Out of curiosity, why a plain `Label` here? If you used a tree like the production menu that would make it easy to display the icons as well, right? Or is there a better way to do that in Godot?

TomWerner (approved):
Looks great! I don't fully understand some of the Godot stuff, but I'm sure I'll be trying to copy some of it in the future :)

C7/UIElements/CityScreen/CityScreen.cs
esbudylin: I'm not sure if Tree would work since, besides text and the building icon, Civ3 also shows icons for culture output, maintenance, and happiness. I think we'll have to implement something custom here.

image

TomWerner: Ah, that's a good point

PR #651, opened 2025-04-05
Import the civ3 difficulty levels and make the default game regent 

TomWerner:
This is the "fair" difficulty, where the player and AI play by roughly the same rules.

I need the difficulty levels to properly implement corruption.


C7GameData/Difficulty.cs
```diff
@@ -0,0 +1,41 @@
...
+
+		// Bonuses for unit support.
+		public int AdditionalFreeUnitSupport;
+		public int UnitSupportBonusForEachSettlement;
+
+		public int AttackBonusAgainstBarbarians;
+
+		// The cost factor for techs, growth, and production, 10 is a neutral
+		// value.
+		public int CostFactor;
```
esbudylin, line 31: Wouldn't it be more logical to make 1 a neutral value and store the cost factor as a float instead?

C7GameData/Difficulty.cs
```diff
@@ -0,0 +1,41 @@
...
+
+		public int PercentageOfOptimalCities;
+
+		public int AIToAITradeRate;
+		public int CorruptionPercentage;
+
+		// Number of citizens quelled by military.
+		public int MilitaryLaw;
```
esbudylin, line 43: How is this field used in the original game? It stays the same across all difficulty levels, and I guess the actual military police limit is set by the government type.

C7GameData/Difficulty.cs
TomWerner: Some of the formulas, like the [tech cost formula](https://forums.civfanatics.com/threads/research-cost-formula-v1-29f.29485/) assume the cost factor is this style of integer value. I agree it's a bit weird - if it helps a PR down the line explicitly defines this as the AiCostFactor and has a fixed HumanCostFactor of 10.

C7GameData/Difficulty.cs
TomWerner: From https://www.civfanatics.com/civ3/strategy/game-mechanics/the-inner-workings-of-resistance-revealed/ this is related to putting down resistance. It doesn't seem to be anything other than 1 in any of the scenarios either though. Want me to remove it?

esbudylin: Thanks for the link! I think we should keep this field. Can you add a comment to clarify its connection to the resistance mechanic?

esbudylin (approved):

C7GameData/Difficulty.cs
TomWerner: Good call, done.

PR #653, opened 2025-04-05
Clear the texture cache between games

TomWerner:
This fixes the issue where scenarios would display the wrong science arrows, since scenario and regular games use the same names for textures.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -96,6 +96,10 @@ public partial class Game : Node2D {
 		Global = GetNode<GlobalSingleton>("/root/GlobalSingleton");
 
 		try {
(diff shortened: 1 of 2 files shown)


esbudylin:
Looking good! The changes to the code make me think that cache storage should be refactored into a separate object created when loading a save/scenario file. But we can make these changes later.

esbudylin (approved):

TomWerner:
> Looking good! The changes to the code make me think that cache storage should be refactored into a separate object created when loading a save/scenario file. But we can make these changes later.

Yeah, in general I think there's a bit too much reliance on global variables, but haven't yet gotten enough motivation to try and address that.

PR #656, opened 2025-04-05
Fix scenario save loading

TomWerner:
Previously while we could load a scenario just fine, if we then saved the game in c7's format, we would be unable to load that information again in a new game. I noticed this while dealing with some texture cache issues.


esbudylin (approved):

TomWerner:
Thanks!

PR #659, opened 2025-04-06
Rudimentary palace screen

esbudylin:
- Adds the button panel to the lower right menu. For now, most of the buttons (except for Diplomacy and Palace buttons) are hidden during gameplay.
image

- Adds a basic UI for the palace screen (for now, it just shows the background picture).
image

- Creates a new reusable scene for the black sidebars, separating it from the city screen scene.  

- Makes more UI nodes visible in the editor by marking them with the [Tool] annotation.


C7/UIElements/GameStatus/StatusMenu.cs
```diff
@@ -0,0 +1,42 @@
...
+	[Export] PopupOverlay popupOverlay;
+	[Export] PalaceScreen palaceScreen;
+
+	public override void _Ready() {
+		openDiplomacy.Pressed += OpenDiplomacyPopup;
+		openPalaceScreen.Pressed += palaceScreen.Show;
+	}
+
+	public override void _Process(double delta) {
+		using (UIGameDataAccess gameDataAccess = new()) {
```
TomWerner: Is there an advantage to scoping the `UIGameDataAccess` like this vs doing `using UIGameDataAccess gameDataAccess = new();`?

TomWerner (approved):
Very cool!!

C7/UIElements/GameStatus/StatusMenu.cs
esbudylin: Good catch! Yes, it can be simplified.

PR #660, opened 2025-04-07
Implement corruption of shields and commerce

TomWerner:
This is taken from https://forums.civfanatics.com/threads/everything-about-corruption-c3c-edition.76619/, which goes into detail about the formulas involved. We don't yet have all the pieces (like trade networks and courthouses) but we have enough for this to be usefully implemented.

Where possible I've gone for strong types (`CorruptableValue`) to make it impossible to forget about corruption.

One oddity is that corrupt commerce is always displayed on the science row of the city screen, regardless of whether the science slider is turned on. After chatting in the discord this seems fine; corruption is applied and then that result has the sliders applied, so this is just a graphical choice by the game. We can match it pretty easily.

----

Here are some diff screenshots from a save I created for this purpose (so I've avoided reading between cities, and built no courthouses).

Capitol (food difference is because we haven't implemented the despotism penalty yet):

image

image

City 1:

image

image

City 2:

image

image

### DIFF
--- a/C7/UIElements/CityScreen/CityScreen.cs
+++ b/C7/UIElements/CityScreen/CityScreen.cs
@@ -262,13 +262,14 @@ public partial class CityScreen : Control {
 
 	private void RenderCommerceDetails(City city) {
 		CommerceBreakdown breakdown = city.CurrentCommerceYield();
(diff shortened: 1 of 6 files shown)


esbudylin (approved):

TomWerner:
Thanks!

PR #666, opened 2025-04-08
Implement the despotism penalty and show it visually

TomWerner:
We can make this more obvious to players by "X"ing out the penalized yield and labeling it "Desp". 

image


esbudylin (approved):

TomWerner:
Thanks!

PR #668, opened 2025-04-09
Make the domestic advisor a godot scene

TomWerner:
I'd like to add some additional features without having to guess+check positions.

I based this off of https://github.com/C7-Game/Prototype/pull/659/files, but had to add the `no_sidebars` scene for reasons I don't fully understand to get it to work.

### DIFF
--- a/C7/C7Game.tscn
+++ b/C7/C7Game.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=21 format=3 uid="uid://cldl5nk4n61m2"]
+[gd_scene load_steps=22 format=3 uid="uid://cldl5nk4n61m2"]
 
 [ext_resource type="Script" path="res://Game.cs" id="1"]
 [ext_resource type="Script" path="res://DoubleClickHandler.cs" id="2_t735q"]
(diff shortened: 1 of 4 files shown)


esbudylin:
Nice work! I see just a few issues with this approach.

1) This PR leaves the "Last Advisor" button in the top left corner in a non-functional state.

2) The usage of an additional "no_sidebars" scene just to center the scene seems excessive. Why can't the new domestic advisor scene be centered with the use of the CenterContainer, as was done before?

3) I think we should keep the logic for showing advisors in a specific class (like it's done now in the Advisors class) instead of moving it into the more general Game class. This way we can manage the last advisor display and screen centering separately.

TomWerner:
Thanks for the review! I'll address points 1 and 3 a bit later when I have time, but I gave point 2 a shot just now.

It still feels a bit fancy though, but I'm not sure how to simplify it further. Various things I tried when initially working on this either led to unparseable scenes, or errors like `WARNING: Parent path './Screen' for node 'Background' has vanished when instantiating: 'res://UIElements/Advisors/domestic_advisor.tscn'.`.

Is this approach of just using a `CenterContainer` in `no_sidebars` sufficient?

TomWerner:
> This PR leaves the "Last Advisor" button in the top left corner in a non-functional state.

Good catch, fixed.

> The usage of an additional "no_sidebars" scene just to center the scene seems excessive. Why can't the new domestic advisor scene be centered with the use of the CenterContainer, as was done before?

I think this is better, but I still had problems with "nodes vanishing" without the no_sidebars scene. This is almost certainly due to my complete lack of Godot knowledge - I was only able to get the PR this far by copying your palace PR.

> I think we should keep the logic for showing advisors in a specific class (like it's done now in the Advisors class) instead of moving it into the more general Game class. This way we can manage the last advisor display and screen centering separately.

Good call, done.

esbudylin:
@TomWerner You can remove the dependency on "no_sidebars" by opening the "domestic_advisor" scene, right-clicking on the root node, and selecting "Clear Inheritance". Also, since the HBoxContainer in the "domestic_advisor" scene has only one child, it can be removed. Otherwise, LGTM.

TomWerner:
Ah, that did the trick. Thank you!!

PR #669, opened 2025-04-10
Add more labels to the domestic advisor

TomWerner:
Now that this is in the godot editor this is much easier to do, though I still don't know how to hook up the `[Export]` variables without editing the `tscn` text file.

image


esbudylin (approved):
Looks good! Just in case you didn't know, there is a menu in the engine for setting export properties. You can access it by opening a scene and clicking on a node. The menu will appear on the right side in the "Inspector" tab. You'll need to rebuild the project for new properties to appear on this list.

TomWerner:
> Just in case you didn't know, there is a menu in the engine for setting export properties. You can access it by opening a scene and clicking on a node. The menu will appear on the right side in the "Inspector" tab. You'll need to rebuild the project for new properties to appear on this list.

I didn't know that - thank you!

PR #674, opened 2025-04-12
Add logic to check for resource access

esbudylin:
This PR implements basic logic to check for resource access. 

For now, the algorithm only considers regular roads since building effects (harbors, airports) have not yet been implemented. Also, the current implementation doesn't take into account wars (in the original game, trade networks cannot pass through civilizations you're at war with).

The PR also adds accessible resources to the city screen. For now, only strategic resources are displayed with icons. Showing luxury resources would require reading their icons from a separate PCX file ("Art/city screen/luxuryicons_small.pcx").

Screenshots:
image_2025-04-11_19-25-53
image_2025-04-11_19-26-27


C7Engine/AI/Pathing/EdgeWalker.cs
```diff
@@ -59,4 +59,19 @@ public override IEnumerable<Edge<Tile>> getEdges(Tile node) {
...
 		}
 	}
+
+	public class TradeNetworkWalker : EdgeWalker<Tile> {
+		public override IEnumerable<Edge<Tile>> getEdges(Tile node) {
+			List<Edge<Tile>> result = [];
+
+			// TODO: When buildings are implemented, the logic for airports and harbors should also be included.
+			foreach (Tile neighbor in node.neighbors.Values) {
+				if (neighbor.HasCity || neighbor.overlays.road) {
```
TomWerner: Cities should have an implicit road already - https://github.com/C7-Game/Prototype/blob/Development/C7Engine/EntryPoints/CityInteractions.cs#L33

C7/UIElements/CityScreen/CityScreen.cs
```diff
@@ -236,13 +239,70 @@ private void OnShowCityScreen(ParameterWrapper<City> city) {
...
+
+		foreach (var child in strategicResources.GetChildren()) {
+			strategicResources.RemoveChild(child);
+		}
+
+		foreach ((C7GameData.Resource resource, int count) in resourceCounter) {
+			VBoxContainer resourceContainer = new();
+			resourceContainer.AddThemeConstantOverride("separation", 0);
+
+			var texture = Util.GetResourceTexture(resource);
```
TomWerner, line 262: Why doesn't this work for Luxuries too? They're present on the same pcx as the strategic resources.

Even if the base game uses a separate set of icons in the city screen for luxuries, it seems like we don't necessarily need to

C7Engine/CityExtensions.cs
```diff
@@ -3,11 +3,47 @@ namespace C7Engine {
...
+			if (!city.owner.KnowsAboutResource(resource)) {
+				return false;
+			}
+
+			if (city.owner.resourcesInBorders.TryGetValue(resource, out var resTiles)) {
+				return resTiles.Where(t => city.HasTradeAccess(t, pathing)).Any();
+			}
+
+			return false;
+		}
```
TomWerner: Is this function still used?

TomWerner (approved):
Super cool! We may need to do some caching of some of the pathfinding in the future, especially for harbor-to-harbor routes (given that c3x did this for large maps) but this should be perfect for now.

C7/UIElements/CityScreen/CityScreen.cs
esbudylin: The luxury icons for the city screen just look different from the ones used on the game map. We can use the game map textures as placeholders, but they just look off to me in this position.

esbudylin:

PR #685, opened 2025-04-19
Introduce a custom C7-specific terrain pack

TomWerner:
This includes custom terrain, custom mines, custom resources, custom irrigation, custom rivers, and custom forts/barb camps.

It does not yet include fog of war, roads, or railroads.

Since I have very little artistic talent, the bulk of this was done via the following steps:

1. Using [WildWeazel's terrain templates](https://forums.civfanatics.com/threads/templates-for-modding.130897/post-3181736) and picking the terrain color.

2. Removing the pink and replacing it with transparency.

3. In paint.net, duplicating the layer.

4. In one layer copy, applying the "Dent" effect, adding noise, applying the "oil painting" effect, and adding noise again. This shinks the tiles somewhat, so leaving the base layer unchanged is important.

5. Flattening the labels and exporting as a png. I've included the raw `.pdn` files to make this easier to edit if we want to change it in the future.

The resources are the same as what we had before, but because it is now a `.png` file, I was able to use a better resizing algorithm that looks better, and real transparency and a shadow added.

The rivers were solid thick lines with the dent+noise+oil painting effects added.

The mountain and hills were crude isomorphic cube things with the oil painting and noise added. To make them stand out a bit the "inner shadow" plugin was used as well.

----

To make using our own art configurable, I made "Ctrl+Alt+Shift+T" the shortcut for toggling the new art.

----

Some screenshots: 

image

Showing mines/irrigation

image

Large random size:

image


esbudylin:
Nice work! I think in the future we'll need to create a specialized class to manage access to assets. Ideally, it should encapsulate all asset paths, caches, and cropping.

Also, I've noticed a bug with this PR. Try loading this save file with the new texture pack turned on. Then open the city screen for Delhi. After this, the game becomes unable to load any resource images, and the resource layer disappears from the map.

[Gandhi of the Indians, 150 BC.SAV.zip](https://github.com/user-attachments/files/19824623/Gandhi.of.the.Indians.150.BC.SAV.zip)

TomWerner:
Ah nice catch. There was a bug in the multi-arg version of `LoadTextureFromPCX`, I was doing `textureCache[relPath] = ...` instead of `textureCache[key] = ...`. Fixed.

esbudylin (approved):

PR #686, opened 2025-04-20
Merge C7GameData into C7Engine & Get rid of extension classes

esbudylin:
In February, there was a discussion on code organization in our Discord server ([link](https://discord.com/channels/909680526498082837/909857733388406844/1344198864487452693)). There were some good arguments for merging C7GameData into C7Engine. It was pointed out that splitting the code between two projects makes it harder to write and understand. Also, it was mentioned that the Separation of Concerns principle doesn't apply in this case since, despite the separation of the code, there is no actual logical division between the projects.

Since then, I've started to work on code related to city access to resources. This code requires both adding new methods to the City class (a part of C7GameData) and having access to path-finding code (a part of C7Engine), so my only option was to add new methods indirectly via the CityExtension class. This way I was able to add new logic for checking if a city has access to a resource. However, to add resource constraints for building and unit production I'll have to move a significant part of already existing City methods into the extension class.

This experience supports the arguments made during the February discussion. The current separation between C7GameData and C7Engine is creating unnecessary complications in development, forcing the use of extension classes which essentially are workarounds. Merging these projects would eliminate an artificial barrier and create a code organization that better reflects the logic of the game.


TomWerner (approved):
Agreed on your writeup, I think this will make things a lot easier to understand.

For the record I think we should still keep the separation between the engine/gamedata and the UI (it's useful to be able to find all the places we modify the game state quickly) but the separation between the engine and gamedata was just a pain.

PR #688, opened 2025-04-21
Add custom small heads artwork

TomWerner:
This doesn't attempt to have separate genders or races, but does have different eras and specialists.

We can always improve this in the future.

I've included the original paint.net file to make editing in the future easier, as the smiley faces are in a separate layer.

image


TomWerner:
I'm merging this one now because I'm sick of squinting at tiny faces.

PR #689, opened 2025-04-22
Add a list of all cities to the domestic advisor

TomWerner:
This allows zooming to a particular city by clicking on the name, but doesn't yet allow sorting.

I've kept the placeholder nodes in the scene since they make it easier to align things.

Pop heads and icons are next.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Godot-specific ignores
 .import/
+*.import
 export.cfg
 logs/
 .godot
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 # Godot-specific ignores
 .import/
+*.import
```
esbudylin, line 3: Note: [Godot documentation](https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/import_process.html#files-generated) strongly recommends adding import files to version control. For now, it may be not relevant to us, since we copy assets into the file system instead of distributing them as a pck file. But if we ever change this, the import files should be committed.

esbudylin (approved):

PR #692, opened 2025-04-22
Domestic advisor: render pop heads for each city

TomWerner:
image

### DIFF
--- a/C7/UIElements/Advisors/DomesticAdvisor.cs
+++ b/C7/UIElements/Advisors/DomesticAdvisor.cs
@@ -1,6 +1,7 @@
 using C7Engine;
 using C7GameData;
 using Godot;
+using System.Collections.Generic;
 using System;
 
 [GlobalClass]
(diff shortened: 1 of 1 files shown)


C7/UIElements/Advisors/DomesticAdvisor.cs
```diff
@@ -328,9 +329,9 @@ private HBoxContainer MakeCityRow(City city) {
 		hSeparator3.AddThemeStyleboxOverride("panel", new StyleBoxEmpty());
 		hboxContainer.AddChild(hSeparator3);
 
-		PanelContainer popuplationContainer = new();
+		TextureRect popuplationContainer = new();
```
esbudylin: Is there a reason to make this node a TexureRect? Could it be implemented as a Control instead?

esbudylin:
I think I've noticed a bug with the Domestic advisor panel. If I click on a city name inside the panel, then close the city screen, the input on the main screen will be blocked until I press "ESC".

esbudylin (approved):

C7/UIElements/Advisors/DomesticAdvisor.cs
TomWerner: Whoops, that was meant to be a hack while I looked up the proper node type and then I got distracted. Done.

TomWerner:
> I think I've noticed a bug with the Domestic advisor panel. If I click on a city name inside the panel, then close the city screen, the input on the main screen will be blocked until I press "ESC".

Nice catch - that was a pre-existing bug with the "exit" button, but this PR duplicated that incorrect logic. I've fixed the exit button issue and this one.

PR #694, opened 2025-04-23
Include corruption-related buildings in corruption calculations 

TomWerner:
This gets us closer to full accuracy - we still don't have the trade network bit hooked up, and don't have civ traits hooked up.


esbudylin (approved):

TomWerner:
Thanks!

PR #696, opened 2025-04-25
Add building-level mood modifications in a particular city

TomWerner:
This only applies to city-level effects, not global or continent-level effects, but means that temples will start to be functional.

Interestingly there's an "unhappiness" modifier for buildings available in the editor, though the base game doesn't actually use this. I did some testing with a modified scenario and found that it moves happy faces back to content.

### DIFF
--- a/C7Engine/C7GameData/Building.cs
+++ b/C7Engine/C7GameData/Building.cs
@@ -82,6 +82,16 @@ namespace C7GameData {
 
 		public int culturePerTurn = 0;
 
+		// The number of unhappy faces that become content in the city with this
+		// building.
+		public int contentFacesInCity = 0;
+
+		// The number of happy faces that become content in the city with this
+		// building. Note that this is less powerful than other sources of
+		// unhappiness, like drafting or poprushing, which converts happy faces
+		// to sad faces.
+		public int unhappyFacesInCity = 0;
+
 		public HashSet<Resource> requiredResources { get; set; } = [];
 
 		SaveBuilding dataSource;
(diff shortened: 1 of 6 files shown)


C7Engine/C7GameData/Building.cs
```diff
@@ -82,6 +82,16 @@ public class Building : IProducible {
 
 		public int culturePerTurn = 0;
 
+		// The number of unhappy faces that become content in the city with this
+		// building.
+		public int contentFacesInCity = 0;
+
+		// The numbe of happy faces that become content in the city with this
```
esbudylin: Nit: a typo ("numbe")

esbudylin (approved):
Looking good!

I'm wondering how relevant mood recalculations are during the turn. It seems mood doesn't affect gameplay during the turn itself. Sure, it has to be recalculated at the end of the turn and when showing City and Domestic Advisor UI to the player. But is it necessary to recalculate mood after each action that can potentially alter it? If so, should building/destroying roads also be taken into account, since it can alter access to luxuries?

C7Engine/C7GameData/Building.cs

TomWerner:
> It seems mood doesn't affect gameplay during the turn itself. 

Yeah, that's a good point. I'll send you a PR to simplify some of this.

TomWerner:
> > It seems mood doesn't affect gameplay during the turn itself.
> 
> Yeah, that's a good point. I'll send you a PR to simplify some of this.

Send https://github.com/C7-Game/Prototype/pull/698

PR #695, opened 2025-04-23
Basic production of buildings

esbudylin:
This PR:
- Adds resource requirements for unit production
- Adds requirements for building production, including:
  - basic checks (mainly resources and technology) 
  - building-specific requirements (currently implemented: river and coastal placement requirements)
- Implements promotion effects of barracks and harbors a a proof of concept

**Notes on the implementation**
The code is written with future moddability in mind. Instead of hard-coding specific requirements and effects as parts of the Building class, this PR introduces a delegate-based approach. Requirements are stored as a collection of predicate functions, and the promotion effects are attached to the generic "onUnitProduction" action.

Currently, these delegates are generated from SaveBuildings tags, but once we implement user scripting, the functional behavior of buildings could be moved into moddable scripts.

PR #697, opened 2025-04-25
Allow buildings to modify tile yields

esbudylin:
This PR:
- Allows buildings to modify tile yields. Implements the effects of Harbor, Commercial Dock, and Offshore Platform.
- Implements the Republic commerce bonus.
- Improve the save test. Now it compares the JSON files semantically. It will prevent test fails on reordering of the fields.

The implementation of tile modifiers is similar to the #695. Yield bonuses and penalties are implemented via action fields in Government and Building classes. These actions are invoked during yield calculation in the Tile.Yield class. In the future, the same approach can be used for the civilization traits (yield bonuses of seafaring, commercial, agricultural civs).

### DIFF
--- a/C7/Map/TileAssignmentLayer.cs
+++ b/C7/Map/TileAssignmentLayer.cs
@@ -51,9 +51,9 @@ namespace C7.Map {
 				return;
 			}
 
-			Tile.TileYield food = tile.foodYield(city.owner);
-			Tile.TileYield shields = tile.productionYield(city.owner);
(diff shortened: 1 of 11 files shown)


TomWerner (approved):
Very nice! Out of curiosity, what was the issue that needed the JSON comparison changes?

esbudylin:
@TomWerner thank you! For some reason, adding getters and setters to the Government class changed the order of the fields in the serialization.

PR #698, opened 2025-04-26
Only do mood calculations right before they are needed

TomWerner:
This is easier than trying to do them after everything that could affect moods (changing improvements, changing assignments, changing buildings, changing unit positions, etc)

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -241,10 +241,10 @@ public partial class Game : Node2D {
 					OnPlayerStartTurn();
 					break;
 				case MsgShowCityScreen mSCS:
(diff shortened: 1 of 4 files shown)


esbudylin:
Shouldn't the calculations happen also on showing the Domestic Advisor screen?

TomWerner:
> Shouldn't the calculations happen also on showing the Domestic Advisor screen?

Whoops yes, good catch. They're already done in the engine message when the sliders are changed, and on the start of the turn, so it's a little hard to catch, but we should definitely being doing that.

esbudylin (approved):

TomWerner:
Thanks!

PR #699, opened 2025-04-26
Fix the import of what building is being constructed in sav files

TomWerner:
This was fixed for loading building [here](https://github.com/C7-Game/Prototype/pull/697/files#diff-5b1d272ad07723413554fc5a64f60c1032e66dfaae37a3e72929dd9d08b2f2a8R620) when wealth was removed from the list of buildings but missed here.


esbudylin (approved):
Thanks!

PR #700, opened 2025-04-27
Refactor: replace comment annotations with named arguments

esbudylin:
I've recently learned that C# has support for [named arguments](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments), so I went through the codebase and replaced comment annotations (/\*argument=\*/1) with named arguments.


TomWerner (approved):
C# just gets all the cool language goodies!

PR #701, opened 2025-04-27
Store city population in only one way, not 2

TomWerner:
This avoids weird mistakes in accounting, as I accidentally introduced while trying to add pop rushing. Simplifying this will make adding pop rushing easier.

While here I also fixed the issue with scenario loading where none of the tile assignments were valid.


esbudylin (approved):

TomWerner:
Thanks!

PR #702, opened 2025-04-27
Implement hurrying production

TomWerner:
Right now this is only via the right click menu, not via the city screen, but it supports pop rushing and gold rushing, depending on the form of government.

I haven't yet added the happiness penalty for pop rushing; I'll do that in a following PR since it'll require thinking about negative happiness and weird things.

Pop rushing a barracks:

image

image

Cash rushing a temple:

image

image

### DIFF
--- a/C7Engine/C7GameData/City.cs
+++ b/C7Engine/C7GameData/City.cs
@@ -137,6 +137,78 @@ namespace C7GameData {
 			return TurnsToProduce(itemBeingProduced);
 		}
 
+		private int ShieldCostForHurrying() {
+			// If there are no shields in the box, hurrying costs double.
+			if (shieldsStored == 0) {
+				return itemBeingProduced.shieldCost * 2;
+			}
+			return itemBeingProduced.shieldCost - shieldsStored;
+		}
+
+		// Returns the feasibility of hurrying production
+		public class HurryProductionDetails {
+			public string? errorMessage;
+			public string? costMessage;
+
+			public int popCost = -1;
+			public int goldCost = -1;
+		}
+		public HurryProductionDetails GetHurryProductionDetails() {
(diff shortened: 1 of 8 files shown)


C7Engine/C7GameData/City.cs
```diff
@@ -137,6 +137,72 @@ public int TurnsUntilProductionFinished() {
...
+			}
+			return itemBeingProduced.shieldCost - shieldsStored;
+		}
+
+		// Returns the feasibility of hurrying production
+		public struct HurryProductionDetails {
+			public string? errorMessage;
+			public string? costMessage;
+		}
+		public HurryProductionDetails GetHurryProductionDetails() {
```
esbudylin, line 156: Currently there is code duplication between GetHurryProductionDetails and HurryProduction. I think ideally the logic for calculating the availability and cost of hurrying production should be separated into a new method.

TomWerner: Whoops, I somehow missed the email for this review. How's this? Now everything comes from the details struct.

esbudylin: Thank you, that looks good! A minor request: if you make non-trivial changes after a review, could you please add them as a new commit instead of force-pushing? For me that would make them easier to review.

esbudylin (approved):

C7Engine/C7GameData/City.cs
TomWerner: Sure, I can try to do that. I'm a big user of jj (https://github.com/jj-vcs/jj, highly recommended!) and it tends towards the one change=one commit model of things like Gerrit, and which GitHub doesn't have great review support for. But JJ supports multiple commits per PR too, and it makes reviewing easier I can certainly do that.

Thanks for the review!

PR #703, opened 2025-04-28
Add the unhappiness penalty for pop rushing 

TomWerner:
This rounds out the 4 combinations of happiness +/- signs, so we can now add the additional factors pretty trivially.

### DIFF
--- a/C7/Text/c7-static-map-save.json
+++ b/C7/Text/c7-static-map-save.json
@@ -70220,7 +70220,8 @@
     "maximumResearchTime": 50,
     "minimumResearchTime": 4,
     "shieldValueInGold": 4,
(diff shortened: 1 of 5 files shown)


esbudylin (approved):
I think this PR includes a file that was committed by a mistake (C7/Text/c7-autosave-turn-0.json). Otherwise, LGTM.

TomWerner:
> I think this PR includes a file that was committed by a mistake (C7/Text/c7-autosave-turn-0.json). Otherwise, LGTM.

Whoops, yes. That gets added to the gitignore in the map gen PR. Deleted.

TomWerner:
Thanks for the review!

PR #705, opened 2025-05-04
Implement clearing forests and wetlands

esbudylin:
This PR allows player-controlled workers to clear wetlands and forests. It also refactors the Terraform system by implementing the requirements and effects as delegates (similar to #697 and #695).

### DIFF
--- a/C7Engine/C7GameData/Actions.cs
+++ b/C7Engine/C7GameData/Actions.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Collections.Generic;
+
 namespace C7GameData {
 	// The strings for each action correspond to values in project.godot for keyboard shortcuts
 	public static class C7Action {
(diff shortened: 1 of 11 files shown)


C7Engine/C7GameData/Actions.cs
```diff
@@ -39,6 +42,21 @@ public static class C7Action {
...
+			UnitClearForest,
+			UnitClearWetlands,
+			UnitPlantForest,
+			UnitClearDamage,
+			UnitBuildBarricade,
+			UnitBuildFortress,
+			UnitBuildOutpost,
+			UnitBuildAirfield,
+			UnitClearDamage,
+			UnitBuildRadarTower];
```
TomWerner: Super minor nit, could you put each action on its own line, rather than having UnitBuildMine and UnitBuildRadarTower be on the same line as braces?

C7Engine/C7GameData/Terraform.cs
```diff
@@ -1,7 +1,27 @@
...
 
 namespace C7GameData;
 
+public static class TerraformRules {
+	public static readonly Dictionary<string, Action<Tile>> OnCompleteActions = new() {
+		{C7Action.UnitBuildMine, tile => tile.overlays.mine = true},
+		{C7Action.UnitIrrigate, tile => tile.overlays.irrigation = true},
+		{C7Action.UnitBuildRoad, tile => tile.overlays.road = true},
+		{C7Action.UnitClearWetlands, tile => tile.ClearTerrainOverlay()},
+		{C7Action.UnitClearForest, tile => tile.ClearTerrainOverlay()}, // TODO: add the bonus shields to the nearest city
```
TomWerner: It might be worth noting in the Todo that we'll need to keep track of whether the tile has had a forest cleared already, since shields are only awarded the first time.

TomWerner (approved):
Nice!

esbudylin:

PR #709, opened 2025-05-08
Generate random maps for new games!

TomWerner:
This is a big step towards making C7 an actually playable game - now there is something to explore, and each game can be different.

I've left the existing static map available via the "tutorial" button, since it is useful for testing to have the same map.

This is unfortunately a large PR, so here are some assorted notes:
 - We don't yet have a notion of `conquests.biq`, so I've cannibalized the static save to use as a rules base and then swapped out the map and starting locations. A future PR will make an actual new game flow where the user can pick their civ and map settings - until then we use a standard size, pangaea (since there are no functional boats yet, and the AI can't do water stuff), and otherwise random settings.
 - As with the heightmap PR, this PR includes a test that generates PNG files showing a high level view of generated maps. This is very useful for tweaking parameters, since it allows for generating multiple maps at once. It just uses a collection of color constants for each biome - better than nothing. A sample of those images is included below.
 - The river logic is pretty crude, and leaves some false starts that should be cleaned up. It also doesn't properly assign the N/S/E/W flags, just the NE/SE/NW/SW flags
 - The biome assignment is pretty crude, and could be made smarter by considering things like whether there is moisture near a biome.
 - The thresholds for the various things are pretty arbitrary - I played with several of them to get things that "looked right", and others were tweaked to make sure all the resources got placed (not enough mountains => hard to place gems, for example)
 - The logic for figuring out which graphics to assign to a tile are a bit primitive, I think there's something fancier for grassland in the original game that avoids repeating the same tile, but I figured this was "good enough".
 - Sometimes one of the luxuries isn't placed due to the constraints on how close they can be together, so I might play with that a bit more.

# Screenshots!

Here are some samples from hitting "new game".

image

image

image

----

And here are the test images.

image

image

image


C7Engine/C7GameData/MapGenerator.cs
```diff
@@ -0,0 +1,1681 @@
+namespace C7GameData {
```
esbudylin: Nit: Is the C7GameData namespace the right place for the map generator? I think C7Engine would fit it better.

C7Engine/C7GameData/MapGenerator.cs
```diff
@@ -0,0 +1,1681 @@
...
+			int aridPlainsBoost = wc.climate == WorldCharacteristics.Climate.Arid ? 8 : 0;
+			int aridDesertBoost = wc.climate == WorldCharacteristics.Climate.Arid ? 5 : 0;
+
+			switch (wc.temperature) {
+				case WorldCharacteristics.Temperature.Cool:
+					// About where Minneapolis is - in the last ice age glaciers
+					// went lower, but we don't want all tundra.
+					tundraLatitudeThreshold = 45;
+
+					tundraLowerBound = hm.FindSeaLevel(70);
```
esbudylin: Could you explain the usage of the FindSeaLevel here? I don't understand how the sea level relates to the bounds of terrain, and what the argument to this function represents.

C7Engine/C7GameData/MapGenerator.cs
```diff
@@ -0,0 +1,1681 @@
...
+			xdgc = 2,
+			xdpc = 3,
+			xdgp = 4,
+			xggc = 5,
+			wCSO = 6,
+			wSSS = 7,
+			wOOO = 8,
+		}
+
+		private static void AssignTextureDetails(WorldCharacteristics wc, GameMap m) {
```
esbudylin: It's a bit surprising to see logic for assigning textures as a part of the map generator class. I think logic related to the visuals should be separated into a different class.

C7Engine/C7GameData/MapGenerator.cs
```diff
@@ -0,0 +1,1681 @@
...
+				}
+
+				PlaceStrategicResourceType(rand, wc, m, r, tileIndicies);
+			}
+
+			// Bonus resources.
+			PlaceBonusResources(rand, wc, m, resourcesToPlace.Where(x => x.Category == ResourceCategory.BONUS).ToList(), tileIndicies);
+		}
+
+		private static int GetAppearance(WorldCharacteristics wc, Random rand, Resource r, int minCount) {
```
esbudylin: Could you add a comment explaining calculations in this method?

esbudylin (commented):
Very impressive work!

C7Engine/C7GameData/MapGenerator.cs
TomWerner: Does this help?

C7Engine/C7GameData/MapGenerator.cs
TomWerner: Done

C7Engine/C7GameData/MapGenerator.cs
TomWerner: Done

esbudylin (approved):
Thanks! A minor nit: I would also move MapGenerator and TerrainTextureFiles up a directory, so their directory would match their namespace (C7Engine). Otherwise, LGTM.

TomWerner:
> Thanks! A minor nit: I would also move MapGenerator and TerrainTextureFiles up a directory, so their directory would match their namespace (C7Engine). Otherwise, LGTM.

Done - thanks for the review! Sorry this PR ended up being so massive.

PR #710, opened 2025-05-10
Refactor terrain improvement system

esbudylin:
This PR introduces a new class, TerrainImprovement, which describes player-made improvements (roads, mines, irrigation, outposts...) Currently, it encapsulates the movement cost, the defense bonus (for barricades and fortresses), and the upgrade relationship (for example, the fact that the railroad is an upgrade from the regular road).

This PR groups the terrain improvements by layers
- Roads
- ResourceDevelopment - mines and irrigation
- Holdings - outpost, radar tower, fortress, barricade. 

The layer systems plays a limiting role when adding new improvements to the tile via the TileOverlays class: each tile can have only one type of improvement per layer.

Some of the Civ3 improvements doesn't fit well into this system: in the Civ3, there are irregularities with Colony (it can coexist with Fortress but not with Outpost or Radar Tower) and Airfield (it can't coexist with anything except roads). I think this can be addressed later by adding custom logic.

In the future, I want to move tile yield effects into the TerrainImprovement class. Also, it would be ideal to allow defining new improvements via scripting/JSON, but it would be tricky since it requires decoupling the improvement drawing logic from the TileOverlayLayer class.

### DIFF
--- a/C7/Map/TileOverlayLayer.cs
+++ b/C7/Map/TileOverlayLayer.cs
@@ -166,7 +166,7 @@ namespace C7.Map {
 		}
 
 		private static bool hasRoad(Tile tile) {
-			return tile.overlays.HasImprovement(TerrainImprovement.road);
+			return tile.overlays.HasImprovement(TerrainImprovement.road) || tile.HasCity;
 		}
 
 		private static bool hasRailRoad(Tile tile) {
--- a/C7/Map/TileOverlayLayer.cs
+++ b/C7/Map/TileOverlayLayer.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using C7GameData;
 using Godot;
 
@@ -45,12 +46,12 @@ namespace C7.Map {
 			Rect2 screenTarget = new Rect2(tileCenter - tileSize / 2, tileSize);
 
 			// Irrigation shows up under roads, so draw that first.
-			if (tile.overlays.irrigation) {
+			if (tile.overlays.HasImprovement(TerrainImprovement.irrigation)) {
 				// Figure out which index into the irrigation texture to use for
 				// this tile.
 				int irrigationIndex = 0;
 				foreach (KeyValuePair<TileDirection, Tile> dirToTile in tile.neighbors) {
-					if (dirToTile.Value.overlays.irrigation) {
+					if (dirToTile.Value.overlays.HasImprovement(TerrainImprovement.irrigation)) {
 						irrigationIndex |= getIrrigationFlag(dirToTile.Key);
 					}
 				}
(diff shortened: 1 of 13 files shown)


C7Engine/EntryPoints/CityInteractions.cs
```diff
@@ -32,12 +32,6 @@ public static City BuildCity(int X, int Y, ID playerID, string name) {
 
 			newCity.SetItemBeingProduced(CityProductionAI.GetNextItemToBeProduced(newCity, null));
 
-			// Cities are treated as though they have a road, but if
-			// a city is build on a mine, the mine should be removed.
-			tileWithNewCity.overlays.road = true;
-			tileWithNewCity.overlays.mine = false;
-			tileWithNewCity.overlays.irrigation = false;
```
TomWerner, line 39: This seems to break road connections to cities when adding new roads.

TomWerner (changes requested):
I think there are some issues with city creation/destruction here, otherwise looks good

C7Engine/EntryPoints/CityInteractions.cs
esbudylin: Thanks for the review! I've tested it, and it works as before. The logic for that was moved into the TileOverlays class (methods HasRoad and MovementCost). Now these methods explicitly handle the presence of a city on a tile, so there is no longer a need for adding/removing roads on city creation/destruction.

esbudylin: Hm, but I really forgot about removing mines and irrigation on city creation, I'll fix that.

TomWerner: Ah, the movement costs are correct, but the graphics are wrong. I think the road layer needs to be updated to have that check too. Here's what I see if I build a road next to a city - there's no connection.

image

esbudylin: Thanks, I see it now!

esbudylin:
@TomWerner I pushed the commit with fixes, let me know what you think!

C7/Map/TileOverlayLayer.cs
```diff
@@ -166,7 +166,7 @@ private static bool HasAnyOverlays(Tile tile) {
 		}
 
 		private static bool hasRoad(Tile tile) {
-			return tile.overlays.HasImprovement(TerrainImprovement.road);
+			return tile.overlays.HasImprovement(TerrainImprovement.road) || tile.HasCity;
 		}
 
 		private static bool hasRailRoad(Tile tile) {
```
TomWerner, line 172: I think this needs `|| tile.HasCity` too.

TomWerner (approved):
LGTM once the railroad thing is handled!

C7/Map/TileOverlayLayer.cs
TomWerner: Oh, actually the railroad issue is movement-related too, not just visual.

esbudylin: Thanks! The issue with movement from a city to a railroad is also present on the Development brunch. Should be fixed now!

PR #711, opened 2025-05-10
Display building icons in the production menu

TomWerner:
image

### DIFF
--- a/C7Engine/C7GameData/Building.cs
+++ b/C7Engine/C7GameData/Building.cs
@@ -94,7 +94,9 @@ namespace C7GameData {
 
 		public HashSet<Resource> requiredResources { get; set; } = [];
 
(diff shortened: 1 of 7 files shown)


C7Engine/C7GameData/Building.cs
```diff
@@ -94,7 +94,7 @@ public class Building : IProducible {
 
 		public HashSet<Resource> requiredResources { get; set; } = [];
 
-		SaveBuilding dataSource;
+		public SaveBuilding dataSource;
```
esbudylin: I think we should keep this field private. The privacy ensures that the SaveBuilding object doesn't change and can be reused for JSON serialization. Otherwise, LGTM.

TomWerner: Good call, done.

PR #714, opened 2025-05-12
Update tile knowledge to account for terrain heights

TomWerner:
#103 helpfully had the details here.

This makes early game exploring much more interesting.

Example of seeing mountains at a distance: 
image


esbudylin (approved):

TomWerner:
Thanks!

PR #717, opened 2025-05-15
 Separate texture paths and crop regions into a Lua config 

esbudylin:
This PR separates texture paths and crop regions from the C# codebase and moves them into a new configuration file. The idea behind this PR is to centralize the definitions of assets in a single file. Also, it helps to lessen the project's coupling with Civ3: with the new system custom art packs could come in any form, not necessarily mimicking the Civ3 art layout.

This is also the first time the Lua interpreter is embedded into the project! I went with Lua for the config file, instead of JSON, since Lua, as a proper programming language, allows declaring variables and concatenating strings, which is quite handy when working with repeating file paths.

This PR introduces a new C# class, TextureLoader, for reading the texture config and accessing the assets. I want the new class to completely replace Util.LoadTextureFromPCX and Util.LoadPCX, but some uses of these methods are not replaced yet. I want to finish the replacement before merging this, so I'm opening this PR as draft to get the feedback on the general direction.


TomWerner:
> Also, it helps to lessen the project's coupling with Civ3: with the new system custom art packs could come in any form, not necessarily mimicking the Civ3 art layout.

Cool! Do you imagine this making it easier to use https://files.freeciv.org/contrib/tilesets/freeland without lots of graphical copy-pasting?

Freeciv uses per-art file `.spec` files to detail some of the crop regions, do you envision something similar here?

esbudylin:
> Do you imagine this making it easier to use https://files.freeciv.org/contrib/tilesets/freeland without lots of graphical copy-pasting?

Yes, that's exactly what I'm going for here! However, there are some difficulties with the terrain tilesets in particular: in the MapView class, the game calculates the right tile based on the terrain data, so I can't just put static values into the config. I'm thinking of making the logic for slicing some of the texture atlases configurable via Lua functions, but it's just an idea at this point. Let me know what you think!

> Freeciv uses per-art file .spec files to detail some of the crop regions, do you envision something similar here?

I don't know much about Freeciv's format, but I think creating a config per file can lead to too many small files. For now, I'm sticking everything into one Lua file, but in the future it can be easily decomposed with the use of Lua's module system.

esbudylin:
I think this PR is ready for review and merging.

There were many file changes, but most were minor and related to replacing old Util methods with new TextureLoader ones. The most significant changes are localized in the TextureLoader class and Lua configs.

As a proof of concept, I implemented the selection of base tile sprites as a Lua function and made appropriate changes to the TerrainLayer class. The system I introduced allows Lua functions to accept any C# object and return texture metadata (the same metadata that can be statically recorded in the config). This works well, though I'm concerned about performance: there's some overhead when calling Lua code from C#. Usually it's negligible, but for map layers that redraw every frame, we may need additional caching.

There are still a few cases where the game loads PCX files directly instead of going through the config. I think that most of them can be replaced after merging by using dynamic Lua logic.

C7/MapView.cs
```diff
@@ -91,18 +69,34 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
 	}
 
 	public override void onEndDraw(LooseView looseView, GameData gameData) {
+		Dictionary<(int fileID, int imageID), ImageTexture> textureCache = [];
```
TomWerner: Is there a reason to recreate this cache on each frame?

C7/MapView.cs
```diff
@@ -91,18 +69,34 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
...
+					texture = TextureLoader.Load("terrain.base", tTD.tile);
+					textureCache[textureKey] = texture;
+				}
+
 				// Multiply size by 100.1% so avoid "seams" in the map.  See issue #106.
 				// Jim's option of a whole-map texture is less hacky, but this is quicker and seems to be working well.
-				Rect2 screenRect = new Rect2(tTD.tileCenter - (float)0.5 * terrainSpriteSize + terrainOffset, terrainSpriteSize * 1.001f);
-				looseView.DrawTextureRectRegion(tripleSheets[tTD.tile.ExtraInfo.BaseTerrainFileID], screenRect, texRect);
+				Vector2 size = texture.GetSize() * 1.001f;
+				texture.SetSizeOverride((Vector2I)size);
```
TomWerner: There still seem to be some seams in the map: 
image

C7/TextureLoader.cs
```diff
@@ -0,0 +1,270 @@
...
+		if (entry == null)
+			throw new Exception($"Texture config not found for key: {configKey}");
+
+		return LoadFromLuaObject(entry);
+	}
+
+	/// Returns a texture based on the config key and a C# object, using a custom mapping function.
+	/// This overload uses the "map_object_to_sprite" function in the config entry to dynamically
+	/// determine which texture to load based on the provided object's properties.
+	public static ImageTexture Load(string configKey, object obj) {
```
TomWerner, line 91: Would it make sense to put a cache here to avoid any lua interactions on the hot path? Similarly with the 1 arg version of `Load` above?

TomWerner (changes requested):
Would it make sense to kick off a background thread to load all the textures in the config when we start a game, just to pre-cache them?

Overall I think this looks pretty good. The map seams are an issue, and I suspect the cache I suggested might help with some of the performance blips.

C7/MapView.cs
esbudylin: Thank you for the review! No, this is my oversight, the cache can be created as a class field.

C7/MapView.cs
esbudylin: I can't replicate this problem, but judging by the code I can see two issues with it 
1) Casting to Vector2I which holds integer values negates the multiplication 
2) This code is mutating the original texture by resizing it, so simply increasing the coefficient wouldn't help (the texture will become bigger on each frame)

I'll see what I can do.

C7/TextureLoader.cs
esbudylin: I haven't done it initially. My concern was that it would create a cache entry for every tile and lead to excessive memory usage. Looking at it again, I think it might be an oversight. Since this cache can re-use textures loaded in the textureCache, the memory usage wouldn't be that bad.  I'm going to try it.

esbudylin:
@TomWerner I've addressed your comments related to the cache and map seams in the latest commit. Since I can't reproduce the issue with seams, could you check if these changes solved the problem on your end? Also, the idea with background caching seems reasonable to me, I think we should implement it in the future.

C7/MapView.cs
TomWerner: Hmm, I still see the seams when I open the "tutorial" to get the static save. 
image


Curiously this goes away if I zoom in or out using the scroll wheel, but not if I zoom in/out via the "z" shortcut.

esbudylin: Hm, it's a bit tricky for me to debug, since I don't know how to reproduce it. I don't see these seams even if I completely remove this resizing logic. Could you increase the resizing coefficient (1.001) and see if it affects anything?

TomWerner:
> @TomWerner I've addressed your comments related to the cache and map seams in the latest commit. Since I can't reproduce the issue with seams, could you check if these changes solved the problem on your end? Also, the idea with background caching seems reasonable to me, I think we should implement it in the future.

I poked around a bit and found that it is the `tile` argument to the `looseView.DrawTextureRect` call. If the boolean is false everything looks correct.

Possibly relevant facts:
 - My Godot version: `4.2.1.stable.mono.official.b09f793f5`
 - I'm using the compatibility driver: `godot  --rendering-driver opengl3 `

I don't think we need tile=true here so if everything looks ok with it false on your end can we use that?

TomWerner (approved):

esbudylin:
@TomWerner thanks! Everything works well on my end with tiling turned off as well.

TomWerner:
It seems like the custom popheads no longer show up properly after this PR? Any idea why?

esbudylin:
@TomWerner my bad! The popheads are one of the few cases where the pcx is still loaded directly instead of going through the lua config (in PopHeads class). It also affected resource icons on the city screen. I'll send a patch for this soon.

PR #718, opened 2025-05-16
Add custom art for the world setup page

TomWerner:
The landmass shapes are from our actual map generator, but with a gold leaf texture for the land and the turquoise color for the water. I used the same parchment texture for the background, but with a sandstone one for the darker regions.

image


esbudylin:
Were the climate/temperature/age images AI-generated? If so, I would suggest not to merge them. AI-generated assets are often associated with low-quality products, and I'm afraid that adding them to the new game screen (basically the first thing players will see after the main menu) would harm the project's reputation. Also, these images don't really need to be custom. I think there are many satellite photos or landscape photos/art under free licenses that could be used there instead.

TomWerner:
> Were the climate/temperature/age images AI-generated? If so, I would suggest not to merge them. AI-generated assets are often associated with low-quality products, and I'm afraid that adding them to the new game screen (basically the first thing players will see after the main menu) would harm the project's reputation. Also, these images don't really need to be custom. I think there are many satellite photos or landscape photos/art under free licenses that could be used there instead.

I think finding usable triptychs that convey what these selections are supposed to convey is harder than you think, especially restricting things to public domain. Some satellite imagery is CC BY-SA 3.0 which isn't necessarily compatible with the MIT license. Some, like NASA's landsat are public domain as long as attribution is given, but even then it's hard to find 3 images.

For example, here are two useful ones that could be used for climate, but we'd be missing the third:

image

image
 
Again, its not like we're drowning in time or artists right now, and I think it's a reasonable goal to try and get the game to a point where we don't need civ3 textures. If you have some specific satellite imagery locations/sources that you'd like me to look at I can, but I'm not going to devote a huge amount of time to it.

esbudylin:
Since the custom art is opt-in now, I no longer have a problem with it. Let's merge it!

TomWerner:
> Since the custom art is opt-in now, I no longer have a problem with it. Let's merge it!

Great, thanks! And to reiterate a point from the discord, if at any point someone has original non-AI art for these UI components, we should certainly prefer that. This is just a stopgap.

PR #719, opened 2025-05-16
Upload the full size images for various bits of our custom art

TomWerner:
These probably shouldn't live on just my computer, in case we need to do future rescaling. 

The parchment and sandstone textures in particular will come in handy for future UIs. Some of this art had additional filters added during scaling, but is roughly what I was working with for the past custom art. I didn't stick these in the `C7/Art` directory because these shouldn't be copied out on releases.


esbudylin:
I don't think the main repository is an ideal place for storing large image files. My concern is they will inflate the size of the repository and make it harder to clone. I think it would be better to separate these files into a separate repository.

TomWerner:
Sure, that makes sense. I can close this PR.

PR #720, opened 2025-05-17
Allow unhappy cities to go into civil disorder

TomWerner:
This requires a number of things:
 - The AI needs to know how to use the luxury slider (it seems like the AI doesn't usually do this in civ3, but a crude algorithm for this is pretty easy)
 - Entertainers need to be functional
 - We need to be able to reassign citizens so that moods are dealt with

I also significantly simplified how we dealt with some types of animated sprites - we can just call `Play()` instead of doing the frame math ourselves.

Size 3 warning: 
image

After ignoring it: 
image

Fixing via entertainer: 
image
image

Fixing via luxury slider:
image

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -452,15 +452,8 @@ public partial class Game : Node2D {
 			foreach (City city in controller.cities) {
 				if (!controller.isHuman) { continue; }
 
-				city.RecalculateCitizenMoods(gDA.gameData);
-				int happy = 0;
-				int unhappy = 0;
-				foreach (CityResident cr in city.residents) {
-					if (cr.mood == CityResident.Mood.Happy) { ++happy; }
-					if (cr.mood == CityResident.Mood.Unhappy) { ++unhappy; }
-				}
-
-				if (unhappy > happy) {
+				City.Mood cityMood = city.RecalculateCitizenMoods(gDA.gameData);
+				if (cityMood == City.Mood.Unhappy) {
 					popupOverlay.ShowPopup(
 						new ConfirmationPopup(
 							$"{city.name} will riot! Are you sure?",
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -442,13 +442,47 @@ public partial class Game : Node2D {
 	}
 
 	private void OnPlayerEndTurn() {
-		if (CurrentState == GameState.PlayerTurn) {
-			log.Information("Ending player turn");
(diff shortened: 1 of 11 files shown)


C7Engine/C7GameData/City.cs
```diff
@@ -703,6 +721,24 @@ public void RecalculateCitizenMoods(GameData gameData) {
 					ApplyMoodChange(sadPoints, content, unhappy);
 				}
 			}
+
+			if (goIntoDisorderIfUnhappy) {
```
esbudylin: I think the code would be simpler if the logic for going into the civil disorder would be separated into a new method, which would be called separately. As far as I understand, there is only one case when city can go into civil disorder (at the end of turn), so having a boolean argument just for this case may  be excessive.

C7Engine/C7GameData/City.cs
```diff
@@ -703,6 +721,24 @@ public void RecalculateCitizenMoods(GameData gameData) {
...
+				isInCivilDisorder = isCityUnhappy();
+				if (isInCivilDisorder) {
+					log.Information($"City {this} for player {owner} is in civil disorder.");
+				}
+			}
+		}
+
+		// Required: RecalculateCitizenMoods must be called first.
+		public bool isCityUnhappy() {
```
esbudylin: It seems that this method is easy to misuse, since it implicitly requires calling another method beforehand. Can we just recalculate the mood as part of this method?

Also, minor nit: since this is a method of City class, I think it can be named just isUnhappy.

C7/Game.cs
```diff
@@ -442,13 +442,47 @@ private void OnPlayerStartTurn() {
...
 		}
+
+		// Prompt the user if they would have a city riot when the turn ended.
+		{
+			using UIGameDataAccess gDA = new();
+			foreach (City city in controller.cities) {
+				if (!controller.isHuman) { continue; }
+
+				city.RecalculateCitizenMoods(gDA.gameData);
+				int happy = 0;
```
esbudylin: The logic for calculating the mood seems to duplicate IsCityUnhappy method.

TomWerner: Done

C7Engine/C7GameData/City.cs

C7Engine/C7GameData/City.cs

esbudylin (approved):

TomWerner:
Thanks!

PR #721, opened 2025-05-17
Handle more food details - city size caps and adjusted growth rates 

TomWerner:
I put the food needed to grow values in the rules as the c# defaults because they don't seem to be present in the editor.

### DIFF
--- a/C7Engine/C7GameData/City.cs
+++ b/C7Engine/C7GameData/City.cs
@@ -50,7 +50,6 @@ namespace C7GameData {
 		public int shieldsStored = 0;
 
 		public int foodStored = 0;
-		public int foodNeededToGrow = 20;
 
 		public bool capital = false;
 		public Player owner { get; set; }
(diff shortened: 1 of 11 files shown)


C7Engine/C7GameData/City.cs
```diff
@@ -222,8 +237,27 @@ public void HurryProduction() {
 		}
 
 		public void HandleCityGrowth(GameData gameData) {
+			// Ensure borders expand before we assign the new citizen, so that
+			// the new citizen can go on one of our new tiles.
+			if (UpdateCultureAndCheckForExpansion()) {
```
esbudylin: The idea of checking for a cultural expansion inside this method is a bit unintuitive to me. I think cultural expansion should be handled separately from city's population growth. Otherwise, LGTM.

esbudylin (approved):

C7Engine/C7GameData/City.cs
TomWerner: Good call, moved it to `Player::HandleCityUpdates`

TomWerner:
Thanks for the review!

PR #723, opened 2025-05-20
Implement granaries

TomWerner:
Their effect now works properly.

### DIFF
--- a/C7Engine/C7GameData/City.cs
+++ b/C7Engine/C7GameData/City.cs
@@ -241,13 +241,7 @@ namespace C7GameData {
 			int foodGrowth = FoodGrowthPerTurn();
 			bool hasGranary = HasGranary();
 
-			// Granaries cause cities to lose twice as much food when the city
-			// is starving.
-			if (foodGrowth < 0 && hasGranary) {
-				foodStored += foodGrowth * 2;
-			} else {
-				foodStored += foodGrowth;
-			}
+			foodStored += foodGrowth;
 			foodStored = Math.Clamp(foodStored, 0, foodNeededToGrow);
 
 			// Handle the city starving.
--- a/C7Engine/C7GameData/City.cs
+++ b/C7Engine/C7GameData/City.cs
@@ -237,7 +237,18 @@ namespace C7GameData {
 		}
 
 		public void HandleCityGrowth(GameData gameData) {
-			foodStored += FoodGrowthPerTurn();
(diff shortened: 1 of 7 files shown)


C7Engine/C7GameData/City.cs
```diff
@@ -237,7 +237,18 @@ public void HurryProduction() {
...
 
 		public void HandleCityGrowth(GameData gameData) {
-			foodStored += FoodGrowthPerTurn();
+			int foodNeededToGrow = FoodNeededToGrow();
+			int foodGrowth = FoodGrowthPerTurn();
+			bool hasGranary = HasGranary();
+
+			// Granaries cause cities to lose twice as much food when the city
+			// is starving.
+			if (foodGrowth < 0 && hasGranary) {
```
esbudylin: Is this really how the original game works? I've tested it, and it seems that when a granary city is starving, it loses a normal amount of food, despite the game visually crossing the food both in the regular storage and in the granary.

TomWerner: Good catch. I fixed this and the graphics bug that the original game had.

C7Engine/C7GameData/City.cs
```diff
@@ -237,7 +237,12 @@ public void HurryProduction() {
...
-			foodStored += FoodGrowthPerTurn();
+			int foodNeededToGrow = FoodNeededToGrow();
+			int foodGrowth = FoodGrowthPerTurn();
+			bool hasGranary = HasGranary();
+
+			foodStored += foodGrowth;
+			foodStored = Math.Clamp(foodStored, 0, foodNeededToGrow);
 
 			// Handle the city starving.
 			if (foodStored < 0) {
```
esbudylin, line 248: Math.Clamp ensures that foodStored is greater than or equal to zero, so this condition will never be true.

TomWerner: Yikes, that one deserves a test. Added a test and fixed.

esbudylin:
I've noticed a graphical bug: when a granary city is starving the total number of the food "boxes" is going down. 

E.g. there are 10 food boxes in the first screenshot, but only 9 in the second and 8 in the third.

image
image
image

esbudylin:
Whoops, I forgot to pull new changes before writing my last comment. Everything seems to work fine now.

esbudylin (approved):

TomWerner:
Thanks for the review!

PR #724, opened 2025-05-21
Give a free tech to the first civ that researches philosophy

TomWerner:


TomWerner (commented):
Self reviewed after a week

PR #725, opened 2025-05-22
Support building great wonders 

TomWerner:
None of these have any effect yet, but we enforce their exclusivity and can build them.


TomWerner (commented):
Self reviewed after a week

PR #727, opened 2025-05-23
Display luxury icons in the city screen

esbudylin:
This PR adds logic for loading small luxury icons and displaying them in the city screen.

image


TomWerner (approved):
Nice! I can try to fix the food/shield alignment in the city screen with the civ3 graphics - I didn't notice that before.

PR #730, opened 2025-05-26
Support great wonders that add buildings to other cities

TomWerner:
I spent a while considering how to implement this, since it seems like there are two ways:

1. Add the wonder-provided cities to each city, which is what this PR does.

2. Any time we need to look at the buildings in a city, check all of the other cities for great wonders and see if they would add a building to our city.

Option 2 seemed worse to me, and would have to be done carefully to avoid some annoying recursion. Option 1 also makes it easy to track where a building came from, which would allow us to easily implement the "you can sell a granary if you built it before you got the pyramids" logic of the original game.

The downside to option 1 is some duplication of logic between the game runtime and the import logic, but it doesn't seem excessive. Let me know if you disagree.

### DIFF
--- a/C7/Text/c7-static-map-save.json
+++ b/C7/Text/c7-static-map-save.json
@@ -67748,7 +67748,8 @@
       "isSmallWonder": false,
       "culturePerTurn": 4,
       "contentFacesInCity": 0,
-      "iconRowIndex": 30
+      "iconRowIndex": 30,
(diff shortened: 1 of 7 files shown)


TomWerner:
I no longer think this is the right way to go - it gets overly complex with wonders that become obsolete. I'll make a new PR.

PR #731, opened 2025-05-26
Display the correct city graphic based on the city size

TomWerner:
I also updated the custom C7 city art graphic to be a png file like all the other modern graphics and slightly adjusted the positioning so it matched the civ3 artwork. (I didn't change the actual city graphics, just changed the format and made the background transparent)

image

image

Also as a note to myself because I might not get back to this for a couple of days, [here](https://forums.civfanatics.com/threads/art-community-art-for-carthage-milestone.676015/post-16246776) is the civfanatics post with additional city art work for C7, including some with walls.


C7/Map/CityScene.cs
```diff
@@ -49,11 +45,35 @@ public override void _Draw() {
...
 			if (city.isInCivilDisorder) {
 				disorderSprite.Show();
 			} else {
 				disorderSprite.Hide();
 			}
 		}
+
+		//TODO: Support multiple city flavors and walls.
+		private void ConfigureCityGraphics(int citySize, int era) {
+			int size = 0;
```
esbudylin: Could you rename this variable to sizeRank or something like that? Right now it is a bit confusing that it doesn't refer to the city population count directly.

C7/Map/CityScene.cs
```diff
@@ -8,29 +8,25 @@ namespace C7.Map {
...
 
-		private readonly Vector2 citySpriteSize;
-
 		private ImageTexture cityTexture;
 		private TextureRect cityGraphics = new TextureRect();
 		private CityLabelScene cityLabelScene;
 		private AnimatedSprite2D disorderSprite;
 		private City city;
+		private Rules rules;
+		private int cachedCitySize = -1;
```
esbudylin: Storing each caching parameter as a separate field seems excessive to me. Can you use a struct of parameters for that?

C7/Map/CityScene.cs
TomWerner: Dont

C7/Map/CityScene.cs
TomWerner: Good call, done. That also made the lua code simpler.

I haven't had a chance to update the child PR yet, but i'll get to that later today.

C7/Map/CityScene.cs
```diff
@@ -5,6 +5,22 @@
 using System;
 
 namespace C7.Map {
+	struct CityGraphicsDetails {
+		// A size rank of 0 is a town, 1 is a city, etc.
+		public int sizeRank;
+		public int eraIndex;
+
+		public bool Equals(CityGraphicsDetails c) {
```
esbudylin: IIRC, you can declare CityGraphicsDetails as a record struct and get built-in overloads for comparison operators. Otherwise, LGTM.

TomWerner: Oh nifty, done.

TomWerner:
Thanks for the review!

PR #733, opened 2025-05-27
Add city defense bonuses, including from walls 

TomWerner:
This is largely based on [this post](https://apolyton.net/forum/miscellaneous/archives/civ3-strategy-archive/100310-how-do-walls-really-work), but the with the difference that the wall bonus does go away for level 2 and 3 cities (which have their own defensive bonuses).

### DIFF
--- a/C7Engine/C7GameData/Building.cs
+++ b/C7Engine/C7GameData/Building.cs
@@ -9,7 +9,7 @@ namespace C7GameData {
 		public static Dictionary<SaveBuilding.Flag, Func<City, bool>> productionRules = new()
 		{
 			{ SaveBuilding.Flag.MustBeCoastal, MustBeCoastal },
-			{ SaveBuilding.Flag.MustBeNearRiver, MustBeNearRiver }
+			{ SaveBuilding.Flag.MustBeNearRiver, MustBeNearRiver },
 		};
 
 		public static Dictionary<SaveBuilding.Flag, Action<MapUnit>> unitProductionEffects = new()
@@ -87,6 +87,8 @@ namespace C7GameData {
 		public bool allowsCitySize2;
 		public bool allowsCitySize3;
 		public bool doublesCityGrowthRate;
+		public int bombardDefenseBonus;
+		public int combatDefenseBonus;
 
 		public int culturePerTurn = 0;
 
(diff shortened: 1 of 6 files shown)


C7Engine/C7GameData/Building.cs
```diff
@@ -193,6 +197,12 @@ public bool CanProduce(City city, HashSet<Resource> accessibleResources) {
 				return false;
 			}
 
+			// Walls (buildings with non-zero bombard defense) can only be built
+			// in towns, not cities or larger.
+			if (dataSource.bombardDefenseBonus > 0 && city.residents.Count > city.owner.rules.MaximumLevel1CitySize) {
```
esbudylin: This piece of logic seems very ruleset-specific. I can imagine modders would want to create buildings that grant bombard defense in cities and metropolises. I think we should introduce a specific flag for the walls (see my next comment), and then use the existing system of production rules in BuildingRules to properly implement it.

C7Engine/C7GameData/City.cs
```diff
@@ -381,6 +381,40 @@ public bool HasWalls() {
...
+			// Buildings, such as walls, can also give bonuses.
+			foreach (CityBuilding cb in buildings) {
+				if (cb.building.combatDefenseBonus == 0) {
+					continue;
+				}
+
+				// Combat defense bonus with no bombard defense bonus means that
+				// this isn't walls, so we can give the bonus regardless of the
+				// size of the city.
+				if (cb.building.bombardDefenseBonus == 0) {
```
esbudylin: Is it possible to make this part of code more general? The idea that defense bonus is dependent on bombard defense seems hard to understand and not friendly to modders. As an alternative, we can, for example, introduce a special flag for that (OnlyGrantsDefenseBonusInTowns or something like that), and then check for this condition when loading Civ3 data.

C7Engine/C7GameData/Building.cs

C7Engine/C7GameData/City.cs
```diff
@@ -381,6 +381,38 @@ public bool HasWalls() {
...
+
+				// If this building doesn't have the "only useful in towns" flag
+				// we can always provide the bonus regardless of the city size.
+				if (!cb.building.onlyUsefulInTowns) {
+					yield return new StrengthBonus(cb.building.name, cb.building.combatDefenseBonus);
+				}
+
+				// But if the building is only useful in towns we need to check
+				// the city size before providing the bonus.
+				if (residents.Count <= gD.rules.MaximumLevel1CitySize) {
```
esbudylin: I think there is a logical bug here: this method will grant the strength bonus to the buildings that are not restricted to towns twice, since "yield return" doesn't stop the method execution. To fix this the second conditional should be declared as "else if"

esbudylin:
Shouldn't this PR also update the method HasWalls of the City class?

TomWerner:
> Shouldn't this PR also update the method HasWalls of the City class?

Yes! I thought I had done that, I may have lost it in a merge at some point. Either way, good catch.

C7Engine/C7GameData/City.cs

esbudylin (approved):

TomWerner:
Thanks!

PR #734, opened 2025-05-30
Redo the main menu

esbudylin:
This commit improves the main menu display.

- The logic for displaying main menu buttons was rewritten using Godot mechanisms. Now buttons are displayed using a UI container, and an anchor is used to position the button container inside the scene.

- Title screen art preserves its ratio when resizing the screen.

- The [Tool] annotation usage was fixed. Previously, the MainMenu class was marked with the tool annotation. This led to issues when editing a scene, since MainMenu requires access to the Godot singleton which is inaccessible within the editor. It also led to the problem where some of the MainMenu export properties were reset on opening a scene: this was caused by the fact that non-tool classes can't be set as export properties of a tool. Now only MenuButtonContainer is marked as [Tool], so the buttons are correctly displayed in the scene editor.

- New title screen art has been added! To prepare for the upcoming Dutch preview, I decided to choose new art for the menu. It is a painting called "The Great Market in Haarlem" by Gerrit Berckheyde (1696). It doesn't have too many details and conveys some of the themes of the upcoming release (population, buildings, culture...), so I think it works pretty well.
image
image


TomWerner (approved):
Very nice! One question about the `.import` files, which I realize I added to the `.gitignore` a while back. If those are included in the project does it avoid the need to open the project in the Godot editor to load the image?

Should we start including all those `.import` files?

esbudylin:
Thank you!

> If those are included in the project does it avoid the need to open the project in the Godot editor to load the image?

Yes, this should do it! The Godot best practice is to commit all import files. However, the situation with Civ3 replacement art is a bit complicated: since these files are loaded directly from disk and not using the Godot resource system, import files do not affect the way they are loaded. I've also just noticed that we are distributing the replacement art twice in our autobuilds: first, they're packed inside the pck file; second, they are copied directly as files (these copies are actually being used by the game). The situation with the menu art is the opposite: since it's loaded as a resource, Godot actually uses the import file for it and reads this image from pck.

For now, I've just added the import file only for the title screen art. I think in the future we'll need to come up with a better system for handling these two types of assets.

PR #735, opened 2025-05-30
Fix which tech makes a wonder obsolete

TomWerner:
I accidentally marked the required tech as the tech making it obsolete, making these wonders impossible to build.


TomWerner:
Merging since this is a pretty trivial bug fix.

PR #736, opened 2025-05-30
Support great wonders that add buildings to other cities

TomWerner:
I originally implemented this via https://github.com/C7-Game/Prototype/pull/730, which added the buildings directly to the city's list of buildings when the wonder was created or the building was created, and removed them if the wonder was destroyed. However, this gets much more complicated once wonders can go obsolete, and especially if there was a scenario where two wonders provided the same buildings. As a result I reworked the PR to instead have us get any relevant wonder-supplied buildings when we iterate over the list of buildings, which works much nicer.

I also enforced that these great wonder properties can only be active for great wonders in the type system, by putting the properties in a struct and using that to identify great wonders. As a bonus, this makes the JSON much shorter.


C7Engine/C7GameData/City.cs
```diff
@@ -54,7 +54,10 @@ public class City {
 		public bool capital = false;
 		public Player owner { get; set; }
 		public List<CityResident> residents = new List<CityResident>();
-		public List<CityBuilding> buildings = [];
+
+		// The list of buildings built in this city. You probably want to use
+		// GetBuildings, which can also include buildings granted by wonders.
+		public List<CityBuilding> buildings_built_in_this_city { private get; set; }
```
esbudylin: I don't quite understand the intent behind the combination of private getter and public setter here. This seems unusual. Also, nit: I think this field could use a more concise name (constructed_buildings or something).

C7Engine/C7GameData/City.cs
```diff
@@ -95,6 +101,54 @@ public bool IsCapital() {
 			return capital;
 		}
 
+		public List<CityBuilding> GetBuildings(bool includeWonderProvidedBuildings) {
```
esbudylin: Is it necessary for this method to have a flag argument? It makes calls to this method somewhat verbose. I think the vast majority of the method calls would set it to true anyway. In the rare cases where it is crucial to receive only constructed buildings, "buildings_built_in_this_city" could be used instead

C7Engine/C7GameData/City.cs
```diff
@@ -95,6 +101,54 @@ public bool IsCapital() {
...
+				buildingsSeen.Add(cb.building);
+			}
+
+			if (!includeWonderProvidedBuildings) {
+				return result;
+			}
+
+			// Loop through all the wonders we control, and if they aren't
+			// obsolete and provide a building, add it to the result.
+			foreach (City c in owner.cities) {
```
esbudylin: Have you considered separating the code for finding the player's wonders into a new method? I think there are a few more places in the game where it would be useful.

C7Engine/C7GameData/City.cs
TomWerner: I was hoping it would be a way to force the use of the `GetBuildings` accessor while still allowing the `SaveBuilding` code to set it. But now that the flag is gone this is no longer relevant so it's just an normal public member.

C7Engine/C7GameData/City.cs
TomWerner: Removed

C7Engine/C7GameData/City.cs

esbudylin (approved):

TomWerner:
Thanks!

PR #737, opened 2025-05-30
Import civ and building traits and apply the production bonus

TomWerner:
Now harbors are cheaper for seafaring civs, for example.

### DIFF
--- a/C7Engine/AI/CityProductionAI.cs
+++ b/C7Engine/AI/CityProductionAI.cs
@@ -94,7 +94,7 @@ namespace C7Engine {
 					//N.B. Eventually this should be influenced by the military strategy
 					baseScore = baseScore + baseScore / 2 * (unit.movement - 1);
 				}
-				baseScore = baseScore - unit.shieldCost;
+				baseScore = baseScore - unit.ShieldCost(null);
 				baseScore = baseScore - 10 * unit.populationCost;
 				baseScore = baseScore + 5 * unit.bombard;
 				return baseScore;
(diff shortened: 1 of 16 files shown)


C7Engine/C7GameData/Building.cs
```diff
@@ -206,6 +206,19 @@ public bool CanProduce(City city, HashSet<Resource> accessibleResources) {
...
 		}
 
+		public int ShieldCost(Civilization? civ) {
+			if (civ == null) {
+				return shieldCost;
+			}
+
+			foreach (Civilization.Trait trait in dataSource.traits) {
+				if (civ.traits.Contains(trait)) {
+					return shieldCost / 2;
```
esbudylin: Could you make this number a constant or a field in the Rules class?

C7Engine/AI/CityProductionAI.cs
```diff
@@ -94,7 +94,7 @@ public static float GetItemScore(IProducible item) {
 					//N.B. Eventually this should be influenced by the military strategy
 					baseScore = baseScore + baseScore / 2 * (unit.movement - 1);
 				}
-				baseScore = baseScore - unit.shieldCost;
+				baseScore = baseScore - unit.ShieldCost(null);
```
esbudylin: Should we allow the ShieldCost method to be called with a null civilization parameter? I think the code would become easier to understand if this method would always require a non-null civilization.

C7Engine/C7GameData/Building.cs

C7Engine/AI/CityProductionAI.cs
TomWerner: Done

esbudylin:
Sorry for hesitating with this PR, but I just don't like how verbose ShieldCost calls are. E.g.

```
itemBeingProduced.ShieldCost(owner.civilization.traits)
```

I think it just makes this method hard to use.

Do you think there is a way to factor this better? For example, we could make shield cost calculations a responsibility of the Player class, so this method would only require an IProducible as an argument. This would transform the call above into something like this

```
owner.ShieldCost(itemBeingProduced)
```

What do you think?

TomWerner:
Seems reasonable to me, the calls are certainly much more concise now. Thanks for the review!

esbudylin (approved):

PR #739, opened 2025-05-31
Add some barbarians to generated maps!

TomWerner:
Right now the barbarians spawn with 3 health and we don't have the player vs barb bonus, so they're a little strong. But they can attack our cities and we can get bonuses for clearing their camps.

image
image

While we do have custom art for barbarian camps, we aren't using the new style of texture loading for them yet, so the civ3 graphics are currently always used.


pcen (approved):

TomWerner:
Thanks!

PR #741, opened 2025-06-02
Support military police and fix getting out of civil disorder

TomWerner:
Previously once in civil disorder we'd be stuck. And once in civil disorder the luxury slider wouldn't seem to work because the commerce was corrupted.

### DIFF
--- a/C7Engine/C7GameData/City.cs
+++ b/C7Engine/C7GameData/City.cs
@@ -509,7 +509,7 @@ namespace C7GameData {
 			return result;
 		}
 
-		public CommerceBreakdown CurrentCommerceYield() {
+		public CommerceBreakdown CurrentCommerceYield(bool respectCivilDisorder = true) {
 			int uncorruptedCommerce = location.commerceYield(this).yield;
 			foreach (CityResident r in residents) {
 				uncorruptedCommerce += r.tileWorked.commerceYield(this).yield;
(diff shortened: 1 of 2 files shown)


C7Engine/C7GameData/City.cs
```diff
@@ -853,6 +853,12 @@ public enum Mood {
 		//  - https://codehappy.net/apolyton/threads/83368-1.htm
 		//
 		public Mood RecalculateCitizenMoods(GameData gameData) {
+			// Always act like we're not in disorder during these calculations
+			// so that the luxury slider has an effect - otherwise the commerce
+			// is always marked as corrupt.
+			bool cachedIsInDisorder = isInCivilDisorder;
+			isInCivilDisorder = false;
```
esbudylin: Disabling the civil disorder state during calculations and then restoring it afterward feels like a workaround. Are there any alternatives to this solution?

TomWerner: The main alternative I can think of would be to add a parameter to `CurrentCommerceYield` like `ignoreCivilDisorder` - would that be preferable?

esbudylin: Yes, I think it would be better, since it'll make the logic more explicit.

esbudylin (approved):

TomWerner:
Thanks!

PR #742, opened 2025-06-03
Stop using `UIGameDataAccess`, use a method accessor instead

TomWerner:
This will allow us to stop using a `Mutex` and instead just use `lock`, as we don't require inter process coordination. This also gives us a natural scoping mechanism - a couple of times I've accidentally hit a deadlock from forgetting the `using` statement on the `UIGameDataAccess` object.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -153,49 +153,55 @@ public partial class Game : Node2D {
 	}
 
 	private void InitializeMapView() {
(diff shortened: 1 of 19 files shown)


TomWerner:
I highly suggest turning off whitespace diffs while reviewing this, there were a lot of indentation changes but there should be no functional changes.

pcen (approved):

PR #743, opened 2025-06-03
Replace `GetHumanPlayers()[0]` with `GetFirstHumanPlayer()` 

TomWerner:
This removes an apparently high source of memory usage (according to Jetbrains Rider) since this is called for every tile. It also means that several UI elements work better in observer mode.


pcen (approved):
very nice!

TomWerner:
Thanks!

PR #745, opened 2025-06-04
Optimize the calculation of trade networks 

TomWerner:
Rather than recomputing the pathing on the fly every time we need a city's luxury count, we can compute the whole network and cache it, and only invalidate it when roads are built (and eventually pillaged) and when cities are built and destroyed.

This cuts the runtime of the load conquest scenarios test from ~53 seconds to ~16 seconds on my (admittedly slow) laptop.

### DIFF
--- a/C7Engine/AI/Pathing/TradeNetwork.cs
+++ b/C7Engine/AI/Pathing/TradeNetwork.cs
@@ -1,0 +1,80 @@
+using System.Collections.Generic;
+using System.Linq;
+using System;
+using C7GameData;
+
+namespace C7Engine.Pathing {
+	public class TradeNetworkSegment {
+		public Dictionary<Resource, int> resourceCounts = new();
+		public HashSet<Tile> tiles = new();
+
+		public void AddTile(Tile t, Player p) {
+			tiles.Add(t);
+			if (t.Resource != Resource.NONE && t.OwningPlayer() == p && p.KnowsAboutResource(t.Resource)) {
+				if (resourceCounts.TryGetValue(t.Resource, out int currentCount)) {
+					resourceCounts[t.Resource] = currentCount + 1;
+				} else {
+					resourceCounts[t.Resource] = 1;
+				}
+			}
+		}
+	}
+
+	// A class for calculating the trade networks of a particular player.
+	//
+	// The main idea is that calculating the entire empire's trade network once
+	// is faster than doing it repeatedly for each city. By doing it once we can
+	// do a simple flood fill of the road network, rather than needing to do
+	// actual pathfinding between different tiles.
+	//
+	// TODO: Handle harbors and airports
+	// TODO: Account for passing through the borders of civs we're at war with
+	public class TradeNetwork {
+		public Dictionary<City, TradeNetworkSegment> segments = new();
+
+		public TradeNetwork(Player player) {
+			HashSet<Tile> seen = new();
+
+			foreach (City c in player.cities) {
+				if (segments.ContainsKey(c)) {
+					continue;
+				}
+
+				// If we don't know about this city yet, start a new network
+				// segment and do a flood fill for all roads coming from the
+				// city.
+				TradeNetworkSegment segment = new();
+				segments[c] = segment;
+
+				Queue<Tile> toCheck = new();
+				toCheck.Enqueue(c.location);
+				seen.Add(c.location);
+
+				while (toCheck.Count > 0) {
+					Tile x = toCheck.Dequeue();
+					segment.AddTile(x, player);
+
+					if (x.cityAtTile != null) {
+						segments[x.cityAtTile] = segment;
+					}
+
+					foreach (Tile n in x.neighbors.Values) {
+						if (n.overlays.HasRoad() && !seen.Contains(n)) {
+							seen.Add(n);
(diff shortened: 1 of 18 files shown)


C7Engine/AI/Pathing/TradeNetwork.cs
```diff
@@ -0,0 +1,80 @@
...
+using C7GameData;
+
+namespace C7Engine.Pathing {
+	public class TradeNetworkSegment {
+		public Dictionary<Resource, int> resourceCounts = new();
+		public HashSet<Tile> tiles = new();
+
+		public void AddTile(Tile t, Player p) {
+			tiles.Add(t);
+			if (t.Resource != Resource.NONE && t.OwningPlayer() == p && p.KnowsAboutResource(t.Resource)) {
```
esbudylin: When calculating resource counts, you check if a player knows about a resource. Doesn't this mean this data should be invalidated when discovering a new tech? Personally, I think it would be better to keep this check out of TradeNetwork.

C7Engine/C7GameData/Terraform.cs
```diff
@@ -6,14 +6,20 @@
...
-		{UnitAction.BuildRailroad, tile => tile.overlays.Add(TerrainImprovement.railroad)},
-		{UnitAction.ClearWetlands, tile => tile.ClearTerrainOverlay()},
+	public static readonly Dictionary<UnitAction, Action<Player, Tile>> ActionEffects = new() {
+		{UnitAction.BuildMine, (_, tile) => tile.overlays.Add(TerrainImprovement.mine)},
+		{UnitAction.Irrigate, (_, tile) => tile.overlays.Add(TerrainImprovement.irrigation)},
+		{UnitAction.BuildRoad, (player, tile) => {
+				tile.overlays.Add(TerrainImprovement.road);
+
+				// The trade network needs to be recomputed whenever roads are added.
+				player.InvalidateCachedTradeNetwork();
```
esbudylin: I think a better place for invalidating the trade network cache when building a road might be the Add method of the TileOverlays class. This class will likely also handle cache invalidation for road pillaging when we implement it.

esbudylin (commented):
Nice work! As a TODO, I guess we'll also need to invalidate cache on war declarations and creation/destruction of airports and harbors?

C7Engine/AI/Pathing/TradeNetwork.cs
TomWerner: Good call, moved the check out.

C7Engine/C7GameData/Terraform.cs
TomWerner: Good call - this also made me realize we need to invalidate all the players trade networks, due to potential transitive effects.

esbudylin:
Can you explain how your code handles border changes related to cultural expansion? For example, if a city's borders expand to include a roaded tile with a resource.

TomWerner:
> Can you explain how your code handles border changes related to cultural expansion? For example, if a city's borders expand to include a roaded tile with a resource.

Oh good catch, that was a bug. Fixed.

esbudylin:
I think there is still an unhandled case here. How does the game behave if civilization A culture-flips a resource tile from civilization B? Will civilization B's network cache be invalidated? I think some of the cache invalidation complexity can be reduced by storing all resources accessible via the trade network in the cache, and checking for tile ownership only when accessing this data. My guess that it shouldn't be too bad performance-wise.

TomWerner:
> I think there is still an unhandled case here. How does the game behave if civilization A culture-flips a resource tile from civilization B? Will civilization B's network cache be invalidated? I think some of the cache invalidation complexity can be reduced by storing all resources accessible via the trade network in the cache, and checking for tile ownership only when accessing this data. My guess that it shouldn't be too bad performance-wise.

Yes, that's a good catch. I realized that we always need to recalculate the trade network of all the players after any network is invalidated, so I switched to storing it all in one class. This ensures there's no cases of mismatches, and with a few nicer methods I was able to pull the logic for resource knowledge into the class, but at lookup time, not network building time. So it still fixes the issue of automatically updating things once a tech is learned.

Sorry for the delay on this, I was most of the way through the player setup logic and wanted to finish that up before this.

C7Engine/C7GameData/City.cs
```diff
@@ -137,21 +137,17 @@ public List<CityBuilding> GetBuildings() {
 			return result;
 		}
 
-		public IEnumerable<IProducible> ListProductionOptions() {
-			HashSet<Resource> accessibleResources = GetAccessibleResources();
+		public IEnumerable<IProducible> ListProductionOptions(GameData gameData) {
```
esbudylin, line 140: Non-blocking: Why do methods of the City class accept gameData as an argument? Can't they just access it directly via EngineStorage?

esbudylin (approved):

C7Engine/C7GameData/City.cs
TomWerner: Some of these `City` methods are invoked on the path of loading a scenario, so the singleton isn't necessarily the relevant `GameData` object. Sorry, I noted that in the new commit description but neglected to mention that in the GitHub comment.

TomWerner:
Thanks for the careful reviews!

PR #747, opened 2025-06-05
Rewrite the logic for choosing production to include buildings 

TomWerner:
This also better centralizes the logic so it's easier to do things like negate the population penalty of a settler in the early expansion phase without having to hop between multiple files.

It definitely isn't perfect - sometimes it ends up with fairly nonsensical cities building wonders, but generally it does reasonable things. The AI now does a decent job of growing their cities past size 6, expanding borders, improving tiles, building the trade network, etc.

I also fixed a couple of odd behaviors I noticed in observer mode:
 - workers mining but not roading a resource, due to the despotism penalty
 - workers not hooking up resources if they weren't worked by a citizen
 - a civ's second city being excessively far from the first city
 - the order in which we checked being over unit support was wrong, leading to units sometimes disappearing due to being disbanded after budget shortfalls


C7Engine/AI/ChooseProducible.cs
```diff
@@ -0,0 +1,369 @@
...
+
+			// Ditto with hospitals.
+			if (building.allowsCitySize3 &&
+				city.residents.Count == player.rules.MaximumLevel2CitySize && city.FoodGrowthPerTurn() > 0) {
+				score += 70;
+			}
+
+			// A town that is at war might consider building walls, otherwise it
+			// probably shouldn't.
+			if (building.providesWalls && building.onlyUsefulInTowns && city.residents.Count < player.rules.MaximumLevel1CitySize) {
```
esbudylin: Aren't these checks (```building.onlyUsefulInTowns && city.residents.Count < player.rules.MaximumLevel1CitySize```) excessive? If they are not true, Walls shouldn't  show up on the list of producible options in the first place.

esbudylin (commented):
I think at some point we'll need to separate the AI-scoring associated with building flags from the ChooseProducible class and come up with a more modular system for defining it. It would be a must-have if we want modders to define buildings with custom effects. However, I think this PR approach works great as a first iteration.

esbudylin (approved):

C7Engine/AI/ChooseProducible.cs
TomWerner: Good point, done.

TomWerner:
> I think at some point we'll need to separate the AI-scoring associated with building flags from the ChooseProducible class and come up with a more modular system for defining it. It would be a must-have if we want modders to define buildings with custom effects. However, I think this PR approach works great as a first iteration.

Agreed, if we want buildings that are more than the fixed set of things that civ3 buildings can do we'll have to figure out some mechanism here. I could imagine mod-specific logic that accepted some stats object (at war, in expansion, best unit, etc) and returned a score.

PR #750, opened 2025-06-06
Improve Godot import files handling

esbudylin:
This commit addresses the issues described in comments to #734

- In accordance with Godot docs, import files are excluded from gitignore.

- To prevent the engine from generating import files for Civ3 replacement art, a ".gdignore" file has been added to "Art" subdirectory.

- Since title screen art should be handled separately from the Civ3 replacement art (this file is loaded as a Godot resource, so the engine shouldn't ignore it), it was moved from the "Art" subdirectory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # Godot-specific ignores
 .import/
-*.import
 export.cfg
 logs/
 .godot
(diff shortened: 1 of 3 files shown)


TomWerner (approved):
Nice, thanks!

PR #752, opened 2025-06-08
Allow selecting the civ to play as and the opponents!

TomWerner:
This continues to hackily use the static save as a base, which will eventually need to be replaced by a `conquests.bic` type file that we can use to pull the rules from.

This PR is fairly large so I didn't want to do more than the minimum, but here are some things that I'd like to get to:
 - Actually supporting different difficulties
 - Handling the case of two civs having the same color
 - Adding a "random" button for the player's civ (this requires importing another pcx file and reworking some of the button handling code so I put it off)
 - Making this work for scenarios (which should be doable with some tweaks)

A couple of notes on this PR:
 - When I previously imported the leader heads a couple of civ's were missed because the pediaicons lookup shouldn't be case sensitive, but some civ's have fields that aren't all in caps.
 - I had to fix an alignment issue with the civ3 menu button when the button is on the right
 - I've continued the easier approach of using the pcx leader head rather than loading the full screen thing, like I did in the diplomacy screens. We can always improve this later.

image

### DIFF
--- a/C7/UIElements/NewGame/PlayerSetup.cs
+++ b/C7/UIElements/NewGame/PlayerSetup.cs
@@ -1,0 +1,322 @@
+using Godot;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using C7GameData;
+using C7Engine;
+using ConvertCiv3Media;
+using C7GameData.Save;
+using Serilog;
+
+[Tool]
+public partial class PlayerSetup : Control {
+	private static ILogger log = LogManager.ForContext<PlayerSetup>();
+
+	[Export] TextureRect background;
+
+	[Export] GridContainer playerListContainer;
+	List<Civilization> civilizations = new();
+	Civilization civilization = null;
+	ButtonGroup playerListButtonGroup = new();
+	TextureRect leaderHead = new();
+	[Export] Label civLabel;
+
+	[Export] GridContainer opponentListContainer;
+	List<OptionButton> opponentSelectors = new();
+
+	[Export] GridContainer difficultyContainer;
+
+	Difficulty difficulty = null;
+	ButtonGroup difficultyButtonGroup = new();
+
+	[Export] TextureButton confirm;
+	[Export] TextureButton cancel;
+
+	[Export] Label loadingLabel;
+
+	ID.Factory ids;
+
+	// TODO: read this from the rules based on the world size
+	const int NUM_OPPONENTS = 7;
+
+	// Called when the node enters the scene tree for the first time.
+	public override void _Ready() {
+		background.Texture = TextureLoader.Load("player_setup.background");
+
+		SaveGame save = GetSave();
+
+		// Set up buttons for the civs the player can play as.
+		//
+		// TODO: Bump this to Dutch once we bump the release.
+		civilizations = save.Civilizations;
+		playerListContainer.Columns = (int)Math.Ceiling(save.Civilizations.Count / 12.0);
+		string initiallySelectedCiv = save.Civilizations.Any(x => x.name == "Carthage") ? "Carthage" : save.Civilizations[1].name;
+		foreach (Civilization civ in save.Civilizations) {
+			if (civ.name == "A Barbarian Chiefdom") {
+				continue;
+			}
+
+			Civ3MenuButton button = new() {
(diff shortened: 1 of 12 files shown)


esbudylin:
I think something may be broken with this PR. On my end the game crashes when generating the world with the following stack trace: 
```
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Generic.Dictionary`2.FindValue(TKey key)
   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
   at System.Collections.Generic.CollectionExtensions.GetValueOrDefault[TKey,TValue](IReadOnlyDictionary`2 dictionary, TKey key, TValue defaultValue)
   at C7GameData.ID.Factory.CreateID(String key) in /home/egor/Documents/Prototype/C7Engine/C7GameData/ID.cs:line 99
   at PlayerSetup.AddUnit(SaveGame save, SavePlayer player, String unitType, TileLocation location) in /home/egor/Documents/Prototype/C7/UIElements/NewGame/PlayerSetup.cs:line 302
   at PlayerSetup.AddPlayer(SaveGame save, Civilization civ, ID defaultGovernment, Boolean isHuman) in /home/egor/Documents/Prototype/C7/UIElements/NewGame/PlayerSetup.cs:line 294
   at PlayerSetup.DoWorldGenerationAndstartGame(GlobalSingleton Global) in /home/egor/Documents/Prototype/C7/UIElements/NewGame/PlayerSetup.cs:line 240
   at PlayerSetup.<>c__DisplayClass24_0.<CreateGame>b__0() in /home/egor/Documents/Prototype/C7/UIElements/NewGame/PlayerSetup.cs:line 215
Aborted (core dumped)
```

TomWerner:
> I think something may be broken with this PR. On my end the game crashes when generating the world with the following stack trace:

The new commit should fix it. The call to `GetSave` wasn't on the UI thread, so we couldn't get the `Global` object, and then the fallback loaded the autosave file, which isn't guaranteed to have the updated rules that specify the starting units.

My autosave file happened to have the starting units, which is why I didn't see it. I reset the autosave file, reproed the bug, and confirmed this fixed it.

C7/UIElements/NewGame/PlayerSetup.cs
```diff
@@ -0,0 +1,322 @@
...
+					new(false, [255]));
+		leaderHead.Scale = new Vector2(1.7f, 1.7f);
+		leaderHead.SetPosition(new Vector2(414, 46));
+
+		string traits = string.Join(", ", civilization.traits);
+		civLabel.Text = $"{civilization.leader} of the {civilization.noun}\n({traits})";
+	}
+
+	private SaveGame GetSave() {
+		try {
```
esbudylin: Godot has a special method for checking if code is running inside the editor, so exception handling isn't needed here: https://docs.godotengine.org/en/stable/classes/class_engine.html#class-engine-method-is-editor-hint

C7/UIElements/NewGame/PlayerSetup.cs
```diff
@@ -0,0 +1,322 @@
...
+
+		UpdateOpponentSelectors();
+	}
+
+	private void UpdateOpponentSelectors() {
+		foreach (OptionButton ob in opponentSelectors) {
+			int index = civilizations.FindIndex(x => x == this.civilization);
+
+			for (int i = 0; i < ob.GetPopup().ItemCount; ++i) {
+				ob.SetItemDisabled(i, this.civilization.name == ob.GetItemText(i));
```
esbudylin: I think some logic for disabling opponent selectors is missing. I can select the same civ as an opponent multiple times. Also, I can first select my opponent as, for example, Greece, then switch to Greece as my civilization, and my opponent selection won't be reset.

C7/UIElements/NewGame/PlayerSetup.cs
TomWerner: Oh nice, thanks!

C7/UIElements/NewGame/PlayerSetup.cs
TomWerner: Done

esbudylin (approved):

TomWerner:
Thanks!

PR #753, opened 2025-06-08
Fix import of building defense bonuses

esbudylin:
This commit fixes a bug related to the import of building defense bonuses from Civ3 files. Previously, the defense bonuses were imported as integers representing the percentage of the bonus (e.g., 50 for Walls). However, the C7 Engine expected these values to be in decimal format (e.g., 0.5 for a 50% bonus)

This commit also refactors the Building class to use the StrengthBonus struct to store the defense bonus. This change makes it more consistent with the codebase.


TomWerner (approved):
Nice catch! The warrior beating a tank meme is great, but probably shouldn't be the expected result if a city has walls...

PR #759, opened 2025-06-15
Award shields after chopping a forest

TomWerner:
While here I also fixed the cost of an improvement so it is more expensive on terrain with higher movement costs (otherwise forest chops took 2 turns) and added handling for industrious civs.

image

### DIFF
--- a/C7Engine/C7GameData/Tile.cs
+++ b/C7Engine/C7GameData/Tile.cs
@@ -492,13 +492,8 @@ namespace C7GameData {
 				return;
 			}
 
-			for (int i = 1; i < 25; ++i) {
-				// We don't award shields beyond rank 2.
-				Tile other = GetTileAtNeighborIndex(i);
-				if (other.rankDistanceTo(this) > 2) {
-					continue;
-				}
-
+			// Check all the tiles within rank 2 of the forest.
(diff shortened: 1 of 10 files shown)


C7Engine/C7GameData/Tile.cs
```diff
@@ -476,6 +480,44 @@ public string YieldString(Player player) {
...
+
+			// Shields can only be awarded if the forest is within some city's
+			// borders.
+			if (OwningPlayer() == null) {
+				return;
+			}
+
+			for (int i = 1; i < 25; ++i) {
+				// We don't award shields beyond rank 2.
+				Tile other = GetTileAtNeighborIndex(i);
```
esbudylin: If I understand correctly, you iterate over the first 25 neighbor tiles while checking if a neighbor is above rank 2 to break out of the loop. This seems like a workaround to me. Would it be possible to write a function that returns all neighbors below rank n instead?

TomWerner: Good call. I forgot that `GetTilesWithinRankDistance` returned a list, not a hash set.

Thanks for all the reviews!

esbudylin (approved):

PR #760, opened 2025-06-15
Award the bonus shield to an industrial civ metropolis

TomWerner:
I also replaced some hardcoded constants with the relevant rules.

### DIFF
--- a/C7Engine/C7GameData/Tile.cs
+++ b/C7Engine/C7GameData/Tile.cs
@@ -328,12 +328,16 @@ namespace C7GameData {
 				yield = 1;
 
 				// There is a size bonus for larger cities.
-				if (cityAtTile.residents.Count >= 7 && cityAtTile.residents.Count < 13) {
+				if (cityAtTile.residents.Count > EngineStorage.gameData.rules.MaximumLevel1CitySize
+					&& cityAtTile.residents.Count <= EngineStorage.gameData.rules.MaximumLevel2CitySize) {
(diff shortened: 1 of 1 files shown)


C7Engine/C7GameData/Tile.cs
```diff
@@ -328,12 +328,16 @@ private int baseProductionYield(Player player) {
...
-				if (cityAtTile.residents.Count >= 7 && cityAtTile.residents.Count < 13) {
+				if (cityAtTile.residents.Count > EngineStorage.gameData.rules.MaximumLevel1CitySize
+					&& cityAtTile.residents.Count <= EngineStorage.gameData.rules.MaximumLevel2CitySize) {
 					yield += 1;
-				} else if (cityAtTile.residents.Count >= 13) {
+				} else if (cityAtTile.residents.Count > EngineStorage.gameData.rules.MaximumLevel2CitySize) {
 					yield += 2;
 
-					// TODO: +1 more for industrial civs.
+					// Industrious civs get +1 production in metropolises
```
esbudylin, line 337: I don't think this is the right way to implement tile yield bonuses for civ traits. I think that for future moddability, we should implement it similarly to the tileModifier fields in the Government and Building classes. This approach will make it easier to separate this logic in user scripts later. However, it will likely require refactoring civ traits from an enum to a class.

pcen (approved):
lgtm - we can revamp tile bonuses for civ traits in a future change

Issue #683, opened 2025-04-19
Not working on mac


zmarlon:
What I have done 

1) I downloaded C7
2) xattr -cr C7.app
3) Downloaded Civ 3 from Steam
4) Set CIV3_HOME variable to root directory of Civ 3
5) Started C7

What happens:
C7 opens, it says: Unable to lead default graphics

Even if I manually select the root folder again, the game does not work.

When I output the Env variable in the terminal, I see that it works
If I open $CIV3_HOME, I end up in the Civ 3 folder

TomWerner:
What do you get if you run `$ echo $CIV3_HOME`? Also if you find the `log.txt` file in the C7 directory, what does it say?

PR #763, opened 2025-06-18
Removing more direct path accesses

TomWerner:
This is a test to see if it gets things to load on mac

### DIFF
--- a/C7/Credits.cs
+++ b/C7/Credits.cs
@@ -10,7 +10,7 @@ public partial class Credits : Node2D {
 	public override void _Ready() {
 		log.Information("Now rolling the credits!");
 		try {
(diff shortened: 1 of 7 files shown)


WildWeazel:
@TomWerner was this successful? Do we have anyone actively testing Mac?

esbudylin:
The problem with using ProjectSettings.GlobalizePath with "res://" paths is that it [does not work](https://docs.godotengine.org/en/stable/classes/class_projectsettings.html#class-projectsettings-method-globalize-path ) in exported projects. 

The ```OS.GetExecutablePath().GetBaseDir()``` could be an alternative (```OS``` is a Godot class here), but I have no idea if this would actually fix loading issues in exported mac projects. Since no one on the dev team owns a mac and can work on this further, I'm going to close this and #755 

Relevant issue: #683

PR #764, opened 2025-07-01
Lua rules engine

esbudylin:
This PR implements a prototype of the Lua rules engine. The new mechanism allows to define game behaviors as Lua functions and use them from the C# code. As a proof of concept, it implements the existing building rules (production prerequisites, tile yield modifiers, unit promotion effects) as Lua code.

The new mechanism implements the following modding interface:

Now, when initializing the game, C7 executes a Lua ruleset script. This script returns a nested table of an arbitrary structure with functions as its leaves. The JSON data for buildings now include fields that allow to specify the paths to Lua functions. This way, specific buildings can be assigned behaviors described by Lua functions.

For example, if the rules script returns a table,

```lua
{
   buildings = {
      production_rules = {
         must_be_near_river = function(city)
            return city.location:BordersRiver()
         end,
      }
   }
}
```

in the JSON file, this production requirement can be applied as follows

```js
{
  "name": "Hydro Plant",
  // other building fields
  // ...
  "productionPrerequisites": [
    "buildings.production_rules.must_be_near_river"
  ]
}
```

**Implementation notes**

I first integrated the Lua interpreter into the project in the PR #717 to use Lua scripts for defining texture metadata. At that time, I used NLua library which provides bindings to the mainstream Lua implementation (Lua PUC). When working on that PR I faced an issue with performance. In NLua, due to it being a C-binding library, interaction between C# and Lua layers bears a performance cost that becomes noticeable when calling Lua frequently (for example, each frame).

That time the problem was solved with some ad-hoc caching. However, when I started working on this PR and initially implemented the rules engine with NLua, I faced the same issue with tile yield modifiers calculations, since they happen each frame. After that, I tried to rewrite the engine with Moonsharp: an alternative implementation of Lua written in C#. This solved the noticeable frame rate drop.

I think we should go with Moonsharp for future use. Lua scripts will likely have a lot of small functions that are called frequently and interact heavily with C# code. So, even if we somehow solve the issue with per-frame updates, the performance cost would still be noticeable when calculating moves of AI civs. However, there are also some downsides to using Moonsharp. The implementation targets Lua 5.2 (current version is 5.4), and hasn't been updated in years, so we would miss out on some handy new Lua features. Also, the documentation of Moonsharp is kind of sparse.

Interestingly, while writing this, I found a [C# Lua implementation](https://github.com/nuskey8/Lua-CSharp) that is actively supported and claims even better C#-Lua interop performance than Moonsharp. The project seems quite early in its development, but maybe one day we'll be able to use it instead.

**TODOs**

After this PR, I'd be interested in porting the TextureLoader from NLua to Moonsharp. Also, we can separate existing government tile yield modifiers (commerce bonus and despotism penalty) into Lua scripts. As a more distant goal, I want to take a stab at implementing a system for custom terrain improvements and terraform actions. I think the combination of custom texture loaders and custom scripted rules would be enough to do something interesting.


TomWerner (approved):
This is very cool!! A couple of things:

What's the experience like editing the lua ruleset - how quickly can you determine if there's an error? Is running the unit tests sufficient? Whenever there's non-compiled code I always wonder when an error would pop up. For example, I played through  the first 60 turns or so, to check barracks/walls/aqueducts/etc. I didn't try to test hydro plants. Should I have?

> implementing a system for custom terrain improvements and terraform actions

I think that'd be interesting as well. I could imagine some modern-era worker move, like "flatten hill", or "build canal" that would modify the landscape. (Hill flattening would likely require picking a neighbor to adopt the overlay terrain of due to the texture rules, and build canal would require recalculating continents to avoid path issues, but both seem feasible).

I more broadly think this lua scripting might give the project a nice example of things that would be impossible with civ3 that are possible with c7. And that's really cool!

TomWerner:
Oh, and maybe let WildWeazel review too? It looks like he added himself.

esbudylin:
Thanks for the review!

When creating a new game, an exception will be thrown only on syntax errors in Lua code or when trying to access a nonexistent Lua function. Type errors in Lua code (accessing nonexistent fields and methods of a C# object, passing an object of an incorrect type to a C# method, or returning a value of the wrong type) can only be caught at runtime.

I think the best way to test the Lua code would be to test the general behavior of the game. Errors in ancient building scripting can already be caught by the headless mode test. Maybe we can expand this test to start in different eras, so it wouldn't only affect ancient buildings.

For now, I've also tested the building scripts manually. That also includes the Hydro Plant: during the testing, I removed its required building and required tech to test it in the Ancient Era.

esbudylin:
@WildWeazel Hi! Are you still interested in reviewing this?

WildWeazel:
I am and thought I might get to it this weekend, but if it's holding up anything don't wait for me. This is one of the areas I have a lot of ideas about, and I think deserves some group discussion before we get too tied up in how it's implemented.

TomWerner:
When you sync could you be sure to move https://github.com/C7-Game/Prototype/blob/Development/C7/Text/TextureConfigs/civ3/unit_control.lua too?

esbudylin:
Since this was open for quite a while and I want to continue working on some Lua-related features, I'll go ahead and merge this. I'm ready to have a discussion about the implementation after the merge if needed.

PR #769, opened 2025-07-13
Upgrade to godot 4.4 to fix CI builds

TomWerner:
The docker image we were using for Godot 4.2 no longer seems to work, as it is using a version of Debian past its long term support window. The 4.4 container seems to use a more up to date version of Ubuntu.


esbudylin (approved):
Huge thanks for fixing that!

PR #772, opened 2025-07-17
Port TextureLoader to Moonsharp

esbudylin:
Since the rules engine was implemented using Moonsharp (see reasons for this choice in #764), TextureLoader is ported to Moonsharp from NLua to avoid having two Lua implementations in the project.

The port itself was pretty straightforward, the only significant difference is that now types of objects passed to TextureLoader should be explicitly registered with Moonsharp.

This PR also makes a few minor code reorganizations:

- Logic for toggling between civ3 and modern graphics was moved from TextureLoader to GlobalSingleton. Now TextureLoader can load configs from arbitrary paths, which would be useful for modding.

- Constant file paths was moved from GlobalSingleton to a new static class GamePaths. In my view, GlobalSingleton wasn't really a good fit for storing constant paths, since its responsibility is to manage mutable state shared between the scenes, not provide access to constants.

### DIFF
--- a/C7/C7.csproj
+++ b/C7/C7.csproj
@@ -14,7 +14,6 @@
     <ProjectReference Include="..\C7Engine\C7Engine.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="NLua" Version="1.7.5" />
     <PackageReference Include="Serilog" Version="2.12.0" />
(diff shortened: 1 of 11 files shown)


TomWerner (approved):
Nice!

TomWerner:
(I'm going to merge this so I can rebase my texture change on top of it and register the c# type I need)

PR #779, opened 2025-07-22
Lua-ify civ colors and add our own c7 colors

TomWerner:


esbudylin (approved):
I think in the future we should come up with a better way for handling colors. It should allow us to have an arbitrary number of civilization colors and provide more flexibility in associating them with civilizations. Ideally, TextureLoader shouldn't be a part of the color handling logic, except for providing access to the PCX file when loading a Civ3 save. But for now, I think the approach of this PR is fine.

TomWerner:
Agreed! I just didn't want to inline the civ3 hex codes in our JSON rules file, since we aren't distributing other civ3 pcx files, and a 1x1 px file is pretty similar to the hex code. I wasn't sure where the line was.

PR #782, opened 2025-07-26
Lua-ify the unit and building icons

TomWerner:
This is pretty crude "art" for now, but it at least makes it easy to distinguish between the two.

image


C7/TextureLoader.cs
```diff
@@ -76,6 +77,8 @@ public ConfigEntry() {
 	static TextureLoader() {
 		UserData.RegisterType<CityGraphicsDetails>();
 		UserData.RegisterType<PopHead.TextureKey>();
+		UserData.RegisterType<UnitPrototype>();
```
esbudylin: There is no need to explicitly register classes of C7GameData namespace as Moonsharp userdata here, since they are being registered in the LuaRulesEngine static constructor. This coupling is kind of awkward, but that's how Moonsharp works: it registers a class as userdata for all lua instances in the project.

esbudylin (approved):

C7/TextureLoader.cs
TomWerner: Done

PR #786, opened 2025-07-30
Moddable terrain improvements

esbudylin:
This PR continues the work on Terrain Improvements from #710. Previously, all terrain improvements were hard-coded in the game code. This PR moves terrain improvement definitions into the JSON file, which makes them moddable. 

This PR also moves the tile yield bonuses  into the TerrainImprovement class: previously, they were part of the TerrainType class which allowed defining bonuses only for mine, irrigation, and road improvements. With the new system, a bonus yield of any type can be defined for any terrain improvement. 

To fully match the Civ3 bonuses, a special tileModifier field was introduced to the TerrainImprovement class. Similar to the building class (#697), it holds a delegate loaded from a Lua function. Its goal is to allow bonus yields depending on tile context: for the Civ3 ruleset, this is useful for implementing the railroad bonus (only grant additional yield if a mine or irrigation is present on a tile).

Finally (the size of this PR got out of hand!), I made a few changes to the Terraform class. Now terraforms (worker's jobs) can be associated directly with a terrain improvement. Also, the existing logic for terraform effects and prerequisites was moved to Lua.

Right now we have almost all the elements for moddable terrain improvements: it is possible to define new terrain improvements, worker jobs for them, and associate them with custom graphics. The only missing system is the UI for worker jobs: right now the game displays buttons only for a hard-coded list of worker actions. Making it moddable would be a logical step forward.


C7Engine/C7GameData/TerrainImprovement.cs
```diff
@@ -8,20 +11,14 @@ public enum Layer {
...
-		public static readonly TerrainImprovement fortress = new(nameof(fortress), Layer.Holdings, defenseBonus: new(nameof(fortress), 0.5));
-		public static readonly TerrainImprovement barricade = new(nameof(barricade), Layer.Holdings, fortress, defenseBonus: new(nameof(barricade), 1));
-
 		public readonly string key;
 
 		public readonly Layer layer;
 		public readonly StrengthBonus? defenseBonus;
 
+		// On the game map, overlapping terrain improvement with a higher zIndex cover those with a smaller one
+		public readonly int zIndex = 0;
```
TomWerner, line 22: How does this differ from the layer?

C7Engine/C7GameData/TerrainImprovement.cs
```diff
@@ -32,24 +29,58 @@ public enum Layer {
 		// 3) result of pillaging (after pillaging, an upgraded improvement will downgrade)
 		public readonly TerrainImprovement upgradesFrom;
 
+		public readonly Action<Tile.Yield> tileModifier;
+
+		private Dictionary<(TerrainType, Tile.YieldType), int> bonusYields = [];
+		private readonly SaveTerrainImprovement dataSource;
+
 		public TerrainImprovement(
```
TomWerner, line 39: Do we need both constructors?

C7/Lua/rules/civ3/terraforms.lua
```diff
@@ -0,0 +1,39 @@
+local UnitAction = ENUMS.UnitAction
+local YieldType = ENUMS.Tile_YieldType
+
+local terraforms = {
+  build_mine = {
+    validator = function(context)
```
TomWerner: I found it a little confusing at first that we have some tables here with a validator, and some with an effect, but none with both.

I think that's just a naming artifact, right? Would naming these like irrigate_validator/mine_validator/etc work, or would that break things?

If it would break things, it might be worth the duplication to duplicate the clear_foliage validator so the wetlands/forests have a validator and an effect, and then have a comment in mine/irrigate that the default effect is to add the improvement.

C7/Lua/rules/civ3/terrain_improvements.lua
```diff
@@ -0,0 +1,26 @@
+local YieldType = ENUMS.Tile_YieldType
+local Layer = ENUMS.TerrainImprovement_Layer
+
+local terrain_improvements = {
+  railroad = {
+    tile_modifier = function(yield)
+      local production_improvement = yield.tile.overlays:ImprovementAtLayer(Layer.ResourceDevelopment)
```
TomWerner: maybe "yield_improvement"? I don't normally think of irrigation as a production improvement

C7/Lua/rules/civ3.lua
```diff
@@ -1,67 +1,5 @@
...
-    local has_fresh_water = city.location:NeighborsFreshWater() or city.location:BordersRiver()
-    return is_town and not has_fresh_water
-  end,
-
-  allows_city_size_3 = function(city)
-    return #city.residents <= rules().MaximumLevel2CitySize
-  end,
+return {
+  buildings = require "civ3.buildings",
+  terraforms = require "civ3.terraforms",
```
TomWerner, line 8: It might be worth a comment in the Lua file explaining what the difference between a terraform and a terrain improvement is.

C7Engine/C7GameData/TerrainImprovement.cs
TomWerner: Edit: is one for the hardcoded civ3 rule list?

C7Engine/C7GameData/Save/SaveTerrainImprovement.cs
```diff
@@ -0,0 +1,72 @@
+using System.Collections.Generic;
+using Layer = C7GameData.TerrainImprovement.Layer;
+
+namespace C7GameData.Save;
+
+public class SaveTerrainImprovement {
+	public static IEnumerable<SaveTerrainImprovement> Civ3Improvements() {
```
TomWerner, line 9: Could you leave a comment explaining why we need this list of improvements?

TomWerner (approved):
This is a pretty big PR so I probably missed something, but this is super cool!

I stuck this in the discord, but this allows for crazy worker moves like "clear hill"

image

image

image

C7Engine/C7GameData/TerrainImprovement.cs
esbudylin: zIndex only affects the drawing order on the game map. The layer affects the gameplay (each tile can have only one type of improvement per layer). Two improvements can belong to the same layer, but have different zIndexes (e.g. irrigation is drawn before roads, and mine is drawn after roads). I added a comment with an explanation to the layer field.

C7Engine/C7GameData/TerrainImprovement.cs
esbudylin: The first constructor is used in EdgeWalkerTest to create a test road instance

C7/Lua/rules/civ3/terrain_improvements.lua
esbudylin: Good catch, done.

C7/Lua/rules/civ3.lua
esbudylin: I added a comment for now, but I think terraform is just a bad name for this type of entity. We should name it something clearer like "worker job".

C7Engine/C7GameData/Save/SaveTerrainImprovement.cs
esbudylin: Done

C7/Lua/rules/civ3/terraforms.lua
esbudylin: I agree, this table should be simpler. There's no need for each function to have a sub-table. Fixed it.

esbudylin:
@TomWerner Thanks for the review! I think I've addressed all your comments. I'll leave this PR open for a few days, in case anyone else wants to contribute a review.

PR #788, opened 2025-08-03
Allow launching in "standalone" mode, with no civ3 assets

TomWerner:
Now if you launch the game and the civ3 install path isn't auto detected, there is an option to select the install path, but also to play in standalone mode. The standalone mode button warns that AI was used for some placeholder art - we want to be up front about that.

image

In this mode we can then start a game, after picking the world settings, civ, opponents, and difficulty level:

image

At that point things partially break, because we don't have any units included in our c7 art directory besides the settler. We can still build buildings and such, but the log is full of complaints about `Art/Units/Worker/Worker.INI`.

### DIFF
--- a/C7/UIElements/MainMenu/MainMenu.cs
+++ b/C7/UIElements/MainMenu/MainMenu.cs
@@ -18,12 +18,14 @@ public partial class MainMenu : Node {
 	MenuButtonContainer ButtonContainer;
 	[Export]
 	AudioStreamPlayer player;
+	[Export] Button UseStandaloneMode;
 
 	GlobalSingleton Global;
 
 	public override void _Ready() {
 		log = LogManager.ForContext<MainMenu>();
 		log.Debug("enter MainMenu._Ready");
+		UseStandaloneMode.Pressed += UseStandaloneModePressed;
 
 		DisplayServer.WindowSetTitle("C7 - Godot 4");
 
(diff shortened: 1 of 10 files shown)


C7/UIElements/MainMenu/MainMenu.cs
```diff
@@ -42,6 +44,9 @@ private void DisplayTitleScreen() {
 		LoadDialog.SetDirectoryForLoading(@"Conquests/Saves");
 		LoadScenarioDialog.SetDirectoryForLoading(@"Conquests/Scenarios");
 
+		if (ButtonContainer.NewGame == null) {
```
esbudylin, line 52: Why do we need to call CreateButtons here? Shouldn't the _Ready method of ButtonContainer take care of that?

C7/UIElements/MainMenu/MainMenu.cs
```diff
@@ -59,8 +64,14 @@ private void DisplayTitleScreen() {
 		};
 		SetToggleGraphicsText();
 
+		// We can't toggle to using civ3 graphics without a root.
+		if (Util.GetCiv3Path == null) {
```
esbudylin: Wouldn't the UseStandaloneMode method work better as a condition? I think it shouldn't be possible to toggle graphics in standalone mode, even if the Civ3 path is set.

C7/UIElements/MainMenu/MainMenu.cs
TomWerner: Added a comment

C7/UIElements/MainMenu/MainMenu.cs
TomWerner: Good call, switched.

esbudylin:
I think there is a bug here: when I start a game or open a save in the Civ3 (non-standalone) mode, the game can't display any units.

TomWerner:
> I think there is a bug here: when I start a game or open a save in the Civ3 (non-standalone) mode, the game can't display any units.

Whoops, yes. I used `i` instead of `index`, fixed.

esbudylin (approved):

TomWerner:
I fixed one more small bug, to allow using c7 paths when modern graphics are enabled, but only if the path doesn't exist in civ3.

PR #789, opened 2025-08-04
Support picking which civ to play as in a scenario

TomWerner:
The scenario picker is 90% duplicated from the player setup logic, but is different enough (no map generation, no starting unit logic, etc) that I think it deserves its own class/scene.

Amusingly I think this code could also be used to allow loading a save game and playing it as a different civilization from whatever turn the game was saved as.

image

Because this now loads/saves a scenario I had to fix a parsing bug that "Man-O-War-1" turned up as well.

### DIFF
--- a/C7/UIElements/NewGame/ScenarioSetup.cs
+++ b/C7/UIElements/NewGame/ScenarioSetup.cs
@@ -32,11 +32,6 @@ public partial class ScenarioSetup : Control {
 
 	[Export] Label loadingLabel;
 
-	ID.Factory ids;
-
-	// TODO: read this from the rules based on the world size
-	const int NUM_OPPONENTS = 7;
-
 	// Called when the node enters the scene tree for the first time.
 	public override void _Ready() {
 		background.Texture = TextureLoader.Load("player_setup.background");
@@ -44,11 +39,9 @@ public partial class ScenarioSetup : Control {
 		SaveGame save = GetSave();
 
 		// Set up buttons for the civs the player can play as.
-		//
-		// TODO: Bump this to Dutch once we bump the release.
 		civilizations = save.Civilizations;
 		playerListContainer.Columns = (int)Math.Ceiling(save.Civilizations.Count / 12.0);
-		string initiallySelectedCiv = save.Civilizations.Any(x => x.name == "Carthage") ? "Carthage" : save.Civilizations[1].name;
+		string initiallySelectedCiv = save.Civilizations[1].name;
 		foreach (Civilization civ in save.Civilizations) {
 			if (civ.name == "A Barbarian Chiefdom") {
 				continue;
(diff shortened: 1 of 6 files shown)


C7/UIElements/NewGame/ScenarioSetup.cs
```diff
@@ -0,0 +1,159 @@
...
+
+	Difficulty difficulty = null;
+	ButtonGroup difficultyButtonGroup = new();
+
+	[Export] TextureButton confirm;
+	[Export] TextureButton cancel;
+
+	[Export] Label loadingLabel;
+
+	ID.Factory ids;
```
esbudylin: I think this field is left unused.

C7/UIElements/NewGame/ScenarioSetup.cs
```diff
@@ -0,0 +1,159 @@
...
+
+	// Called when the node enters the scene tree for the first time.
+	public override void _Ready() {
+		background.Texture = TextureLoader.Load("player_setup.background");
+
+		SaveGame save = GetSave();
+
+		// Set up buttons for the civs the player can play as.
+		//
+		// TODO: Bump this to Dutch once we bump the release.
```
esbudylin: Is this TODO relevant?

C7/UIElements/NewGame/ScenarioSetup.cs
```diff
@@ -0,0 +1,159 @@
...
+
+	[Export] TextureButton confirm;
+	[Export] TextureButton cancel;
+
+	[Export] Label loadingLabel;
+
+	ID.Factory ids;
+
+	// TODO: read this from the rules based on the world size
+	const int NUM_OPPONENTS = 7;
```
esbudylin: This field seems unused too. I guess this is a left-over from the PlayerSetup class?

C7/UIElements/NewGame/ScenarioSetup.cs
```diff
@@ -0,0 +1,159 @@
...
+		leaderHead.Texture = TextureLoader.Load("leader_heads", civilization);
+		leaderHead.Scale = new Vector2(1.7f, 1.7f);
+		leaderHead.SetPosition(new Vector2(414, 46));
+
+		string traits = string.Join(", ", civilization.traits);
+		civLabel.Text = $"{civilization.leader} of the {civilization.noun}\n({traits})";
+	}
+
+	private SaveGame GetSave() {
+		GlobalSingleton global = GetNode<GlobalSingleton>("/root/GlobalSingleton");
```
esbudylin: Right now this method causes Godot to generate a bunch of errors when opening a project, since ScenarioSetup is a tool script and the Godot singletons are not accessible when running tool scripts in the editor. You should consider adding a fallback for this case.

C7/UIElements/NewGame/ScenarioSetup.cs
TomWerner: Done

C7/UIElements/NewGame/ScenarioSetup.cs
TomWerner: Done

C7/UIElements/NewGame/ScenarioSetup.cs
TomWerner: Yeah, just forgot to remove it.

C7/UIElements/NewGame/ScenarioSetup.cs
TomWerner: Added a fallback

TomWerner:
Thanks for the review!

esbudylin (approved):

Issue #264, opened 2022-05-22
Architectural refactoring


WildWeazel:
This is a big one. Following #261 refactor the code to follow the proposed layered architecture: cleanly separate Godot scripts, game logic, data types, C7 & Civ3 file IO, and anything else so that each layer depends only on the one(s) below it. See also #132

QuintillusCFC:
One of the things that has occasionally come up, and is a bit awkward currently, is the AI, and AI data.  Right now, the AI behavior is in C7Engine, but the data classes that stored AI data (which can thus be persisted across a save game, important if we want the AI to have long-term planning) live in C7GameData.  This is consistent with wanting C7GameData to contain what can be written out, but means the AI is somewhat split.

There may be a better way to do this, especially one that considers the potential for pluggable AI components, including those that may include data written as part of the save file.  That's a longer-term thing I want to consider (but before that there are things like teaching the AI how to arbitrate among potential strategies).  Still, it's something that there's been some pushback on (mainly suggesting putting the data in with the C7Engine).  I haven't seen pushback with what I'd consider a better alternative being put forward at the same time.  But FYI, while I support the general separation of concerns, I see why the AI split in particular has garnered a few PR questions.

rchoffardet:
Just stopping by to write here some curiosities I found while scanning the repository:

I'm not sure the messaging pattern is appropriate with C7 (as a Civ3 remake) because Civ3 is very sequential (make an action, watch it being rendered, then make another action, watch it being rendered, ...) while messaging is async by nature and thus requires the workaround with the mutex.

In a similar idea, C7Engine hosts the responsibility of "waiting" for an animation to finish or not. It shouldn't have to. In Civ3, units in the fog of war aren't animated (obviously), but units outside are, by default. IMHO, the fog of war handling is a rendering responsibility and should be handled by the rendering part. Determining to animate and wait for the end of the animation, should be done in the rendering part.

C7Engine / C7GameData separation seems a bit artificial to me. It's like separating behavior and data, which is odd in the OOP paradigm. Even if you want to keep the separation, I don't think having two separate projects is necessary (my 2cts).
However, having no separation between live data (units, cities, ...) and constant data (unit prototypes, techs, strength bonuses, ...) is odd to me. 

I'm willing to contribute if you're interested by my remarks :)

PR #791, opened 2025-08-13
Refactor: improve game creation

esbudylin:
This commit improves mechanism for game creation.

Previously, some logic for initializing the game state was run as part of the save loading class, before the EngineStorage was initialized. As a result, some methods had accepted GameData object as an argument, instead of reading it from the EngineStorage, as is commonly done in the codebase.

In my view, this was problematic, since it introduced a split in codebase and required adjusting the code based on whether it would be run during game creation. This commit addresses this issue by introducing an action field "onGameCreation" to GameData. Now, the SaveGame class only binds the initializing logic to this action, which is executed in the CreateGame class after the EngineStorage is properly initialized.

Methods that accept GameData as argument still exist in the project. Some of them relate to the game creation mechanism, and some of them were written this way because previously it wasn't possible to access EngineStorage from the C7GameData namespace. I think we should gradually refactor them out in favor of accessing GameData through the EngineStorage.

### DIFF
--- a/C7Engine/C7GameData/GameData.cs
+++ b/C7Engine/C7GameData/GameData.cs
@@ -64,6 +64,9 @@ namespace C7GameData {
 		// a road is built or a city is created/destroyed.
 		private TradeNetwork tradeNetwork;
 
+		// An action called after initialization of EngineStorage
+		internal Action onGameCreation;
+
 		public GameData() {
 			map = new GameMap();
 			if (seed == -1) {
--- a/C7Engine/C7GameData/Save/SaveGame.cs
+++ b/C7Engine/C7GameData/Save/SaveGame.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using System.Text.Json.Serialization.Metadata;
+using C7Engine;
 
 namespace C7GameData.Save {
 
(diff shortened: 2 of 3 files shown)


TomWerner (approved):
I think this is reasonable, but I'd like to note that just about every time I've updated some method to accept `GameData` as a parameter instead of accessing the singleton, I've found a case where we were incorrectly accessing the singleton from the UI thread without the mutex.

This will probably be problematic in the future if we have multiplayer support or something.

esbudylin:
Thanks for the review!

It's hard for me to reason about the multiplayer, since we don't have any prototype yet. But shouldn't it be enough to just lock the GameData with EngineStorage.ReadGameData, without passing the GameData down to the method call chain? Could you give an example so I can understand the problem better?

TomWerner:
> Could you give an example so I can understand the problem better?

https://github.com/C7-Game/Prototype/pull/787/files#diff-fe38a4b9d59185ba89749970648f0f9dfb044b91687fa729892d30384b10ecabR766-R768 is the latest example: prior to that PR we were calling `city.owner.GetKnownSpecialists()` on the UI thread, which internally accessed `EngineStorage.gameData`. After that PR, when we more explicitly pass the gamedata reference around, we had to switch to accessing it via `ReadGameData`.

Is there a C# equivalent to mutex annotations like [ABSL_GUARDED_BY](https://chromium.googlesource.com/external/github.com/abseil/abseil-cpp/+/HEAD/absl/base/thread_annotations.h#40) in c++? That would probably help fix this type of thing.

esbudylin:
Wouldn't the same issue occur for any method call on an object that is part of GameData, even if it isn't explicitly accessing the singleton? If an object is coming from GameData, then any method call to it is a shared state access, and should be protected by the lock.

Personally, I think the multithreaded access to shared state just overcomplicates our lives without a good reason. We probably should move away from running the Engine in a separate thread and instead integrate in into the Godot game loop, processing the engine actions inside a Node's _Process method. That should eliminate a need for locking the GameData.

Since all of this is out of scope for this PR, let's move this discussion to #264.

TomWerner:
> Personally, I think the multithreaded access to shared state just overcomplicates our lives without a good reason.

Agreed.

PR #795, opened 2025-08-28
Refactor Engine<->UI to async/await from multithreading

esbudylin:
This PR ports the Engine<->UI relationship from multithreaded to asynchronous execution.

Currently, Engine and UI code execute on separate threads. This creates a risk of data races, and requires locks for accessing GameData from the UI thread. The issue with locks is that they require additional discipline to use and make it harder to reference GameData objects from the UI, see https://github.com/C7-Game/Prototype/pull/791#issuecomment-3194755685.

This PR integrates the processing of engine messages from the dedicated thread into the Godot game loop. For Engine<->UI coordination, this PR introduces an async model: where previously the Engine thread was locked while waiting for UI events, the Engine now uses async/await to pause execution until a UI message arrives. This allows us to maintain the same execution flow while staying in a single-threaded environment, which eliminates the risk of data races.

**Additional notes**

On the UI side, this PR introduces a new class, AnimationController. It takes responsibility for handling animation messages from the Engine. Previously, this was a function of the Game class, but I found it better for readability to give this responsibility to a new class.

This PR also required changes to the AI code. Some AI methods currently mutate the game state by moving units, and since unit moves can trigger battle animations that need to be awaited by the Engine, these methods now need to be async. I think, in the future, it would be better to move away from mutating the game state directly in AI code and instead move the actual execution of AI plans to a higher level.

**Bugfixes**

In addition to the refactoring, this PR fixes two bugs:
- A bug that allowed players to contact a civilization despite that civilization refusing contact: the game displayed both the refusal popup and the talk screen simultaneously.
- A bug which caused trade deals offered by an AI to the player to be executed twice: first, the deal was executed from the talk screen; then, the deal was executed in the AttemptTrading method on the Engine side.


WildWeazel:
Sounds great! I still don't know Godot well enough to comment on the implementation but if @pcen is happy then I'm happy. 

Since the engine is pausing to wait for a response from the UI, what happens with asynchronous (in a gameplay sense) actions originating from the UI, like the player selecting a unit or opening diplomacy? Are those kinds of things only possible when the engine is already paused? (Now that I think of it, that should probably be the case)

>  I think, in the future, it would be better to move away from mutating the game state directly in AI code and instead move the actual execution of AI plans to a higher level.

Agreed

esbudylin:
@WildWeazel, thanks for the comment!

Right now the engine only uses async/await for completing certain animations (movement animations triggered by the player, and combat animations) and completing trade when the AI makes an offer to the player during the AI turn. To be clear, the awaiting doesn't pause the whole engine, like it was done previously with the engine thread - only the code that's waiting for a UI event is paused.

Can you elaborate a bit more on the part of your question related to pausing the engine for handling UI actions? I don't see why these actions should happen only when the engine is paused.

C7/Animations/AnimationController.cs
```diff
@@ -0,0 +1,66 @@
+using System;
+using C7Engine;
+using C7GameData;
+using Godot;
+
+[GlobalClass]
+public partial class AnimationController : Node {
```
TomWerner, line 11: Could you add a brief summary of what this class is for?

C7/Animations/AnimationController.cs
```diff
@@ -0,0 +1,66 @@
...
+	Game game;
+
+	public override void _Ready() {
+		var animSoundPlayer = new AudioStreamPlayer();
+		AddChild(animSoundPlayer);
+		civ3AnimData = new AnimationManager(animSoundPlayer);
+		animTracker = new AnimationTracker(civ3AnimData);
+	}
+
+	public void HandleEngineMessage(AnimationMessage msg) {
```
TomWerner, line 25: This might be worth mentioning in the PR description - something in this PR fixed a bug where automated units had their animations play in sync. Now they go one at a time, which obvious takes a bit longer, but I think is the correct behavior in the base game if "show automated moves" is checked.

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -154,19 +155,23 @@ private static int GetWorkerJobCost(Tile tile, Terraform workerJob) {
...
+			if (!EngineStorage.animationsEnabled) return;
+
+			var msg = new MsgStartUnitAnimation(this, action, ending);
+			msg.send();
+
+			await EngineStorage.WaitForAnimationFinished(msg.animationId);
+		}
+
+		public void animate(MapUnit.AnimatedAction action, AnimationEnding ending = AnimationEnding.Stop) {
+			_ = animateAsync(action, ending);
```
TomWerner, line 168: Question from a c# noob, why does assigning the result here matter?

TomWerner (approved):
Very cool!

C7/Animations/AnimationController.cs
esbudylin: Thanks for the review! Added a comment.

C7/Animations/AnimationController.cs
esbudylin: Interestingly, it wasn't an intended change, I hadn't noticed it before your comment :) It's related to how the engine schedules the automated moves: it does this by sending MsgPerformUnitAction from the UnitInteractions class, and since the engine doesn't process new messages while there are animations pending, these messages end up being processed one-by-one.

Looking closely at this system, it seems a bit odd that this is done as part of the unit selection logic. Probably, it would be more logical to do automatic moves at the beginning of the player's turn? Maybe we should revisit this later.

C7Engine/C7GameData/MapUnit.cs
esbudylin: This is done to suppress the compiler warning about a task not being awaited. It's OK to use it here, since the point of this method is to allow fire-and-forget animation execution.

esbudylin:
If anyone else wants to review this, let me know. Otherwise, I'm going to merge this on Monday.

Issue #347, opened 2022-09-19
Tall terrain graphics don't display greatly


QuintillusCFC:
One of the more challenging terrain packs to draw is the Manhattan terrain, because of the skyscrapers.  C7 is affected by this as well.  Consider this part of the map, by the Williamsburg Bridge:

![image](https://user-images.githubusercontent.com/5287977/190975812-e4218eaa-7534-474b-8560-f8bc5c9a6d2b.png)

It's obviously incorrect that Delancey Street is drawn on top of the skyscrapers to its southeast.  But also note the skyscraper closest to the bridge - FDR drive is drawn over part of it, but not all of it.  This is because tiles that are partially-hidden by fog currently only have the base terrain rather than the overlays drawn.  Since the tarmac of FDR drive is an overlay, the upper part of the skyscraper is not drawn over, though part of the upper-midsection.

Fundamentally, the problem is that we're kind of cheating on isometrics.  Here's out map drawing code at a high level:

```
			foreach (LooseLayer layer in layers.FindAll(L => L.visible)) {
				layer.onBeginDraw(this, gD);
				foreach (VisibleTile vT in visibleTiles)
					layer.drawObject(this, gD, vT.tile, vT.tileCenter);
				layer.onEndDraw(this, gD);
			}
```

We draw each layer in turn, finishing that layer before going on to the next one.  Other than a tweak to exclude the FogOfWar layer (which is drawn after everything else), this is unchanged since 2/16/2022.

It works well most of the time, and we order our layers so that e.g. forests draw before mountains, so tall mountains aren't covered by forests from another tile at the top.  There's also an optimization around drawing the same thing multiple times (see https://github.com/C7-Game/Prototype/pull/81), which adds a significant performance boost.  We may lose that by switching to drawing everything on a tile at once before moving to the next tile, rather than everything in a layer and then the next layer.

Manhattan is unique because the base terrain regularly exceeds the tile bounds, but is intended to take full advantage of the isometric view, and also because some of the overlays, especially roads/railroads (highways) take up the whole tile.  It's kind of a worst-case test scenario for this.  But there should probably be at least an option to have slower performance but full graphical fidelity in these cases.  We might be able to detect what is best based on graphics dimensions.

Probably not a near-term priority but I wanted to get it written up since it's easy to forget otherwise.

QuintillusCFC:
One other Manhattan map challenge, in the logs I got gobs of these messages:

```
[...]

PR #807, opened 2025-09-14
Display textures for borders over hilly terrain

esbudylin:
This PR adds logic for displaying textures for borders that go through mountains, hills and volcanoes. It also includes a minor cleanup in lua texture scripts. The change doesn't affect the standalone mode, since we don't yet have custom graphics for this type of border.

image

### DIFF
--- a/C7/Lua/texture_configs/c7.lua
+++ b/C7/Lua/texture_configs/c7.lua
@@ -1,9 +1,10 @@
 --[[
-	This configuration file holds a table with texture definitions for "modern" graphics.
+	This configuration file returns a table with texture definitions for "modern" graphics.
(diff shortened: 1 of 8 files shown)


TomWerner (approved):
Nice! Do we have some sort of subtle layering bug though? Some of the borders in your screenshot seem to be on top of mountains when I think they should be below them.

esbudylin:
Yeah, you're right, I just noticed it. I think we already have a ticket for this. https://github.com/C7-Game/Prototype/issues/347

It doesn't seem like there is an easy way to fix this with the current layering system. But if we ever port the map rendering logic to the Godot Tilemap, it would be trivial to fix this through y-sorting. It's a pity that the previous port to the Tilemap was never merged...

PR #809, opened 2025-09-28
Refactor: Decouple Terraform from UnitAction

esbudylin:
This PR continues the work on making worker jobs more modular (#786). 

Previously, there were two entities that described worker jobs: the UnitAction enum, which included worker jobs among other actions, and the Terraform class, which described data on worker jobs loaded from a JSON save. This PR makes Terraform the only entity responsible for worker jobs, and removes worker jobs from UnitAction.

The old code relied on hard-coded UnitAction values to retrieve unit animations and button textures for worker jobs, and to associate them with UI actions. Now, all these parameters are specified as part of the Terraform class, and so can be defined directly in a JSON save. 

As a side effect, this PR allows building fortresses and barricades. It's not really a complete feature yet: for now the defense bonuses of terrain improvements don't count when calculating the defense score, and these improvements are allowed to be built anywhere, since data on which terrain types they can be built is not yet imported from CIV3 saves.

### DIFF
--- a/C7/Actions.cs.uid
+++ b/C7/Actions.cs.uid
@@ -1,1 +1,0 @@
-uid://cs3anpqv1ke8b
--- a/C7/C7Game.tscn
+++ b/C7/C7Game.tscn
@@ -236,6 +236,7 @@ control = NodePath("../Control")
 unique_name_in_owner = true
 visible = false
 
+[connection signal="GameInitialized" from="." to="CanvasLayer/Control/UnitButtons" method="SetUpControlButtons"]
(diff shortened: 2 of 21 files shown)


TomWerner (approved):
Nice! Something seems wrong with the turns it takes to build roads, but it seems to be an issue going back a ways. Not sure why

esbudylin:
@TomWerner Thanks! I think this issue can be caused by an incorrect cast of worker progress from float to integer in the PerformBusyAction method of the MapUnit class. Also, I've noticed that this is specific to industrious civs due to their worker speed bonus.

Issue #798, opened 2025-09-08
Graphics FR: render the effects of a specialist in the city screen ui


TomWerner:
Relevant TODO: https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7/UIElements/CityScreen/CityScreen.cs#L759

Original game example, the entertainers have smileys: 

<img width="817" height="613" alt="Image" src="https://github.com/user-attachments/assets/2051fdbe-255c-4f71-bae4-6f996b943bbb" />

We already have the art assets configured, just not associated with the citizen types. The citizen types should have all the details configured already:
 - https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7Engine/C7GameData/CitizenType.cs#L26-L43
 - https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7/UIElements/CityScreen/CityScreen.cs#L130-L136
 - (the domestic advisor should have any not in the city UI): https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7/UIElements/Advisors/DomesticAdvisor.cs#L130-L139

PR #810, opened 2025-09-28
#798 Render specialist' effects

stavrosfa:
The original game as far as I understand, uses different icons for the beaker/gold/shield effect icons than the icons that are used in the commerce, shield production, etc.. 

That is because the effect icons are smaller than the other ones, so my guess is they didn't want to scale the icons, as in 30px max size the icons would not render as nicely. 

That is why I updated the civ3.lua file, to get the proper icons for the job. I also rearranged the other CITY_SCREEN_ICONS to be in order, I didn't remove anything. I am aware I added some duplicate icons, but I was hopping we could discuss this and find a unified way to get these icons from the texture files (original or otherwise) since pretty much all of them have different crops and sizes. 

Regarding the implementation of the effects rendering, I made it so that it obviously renders what you would expect from the  original game.

c7-rendering-specialists'-effects

But I had another thought.

Maybe in the future someone modding it, would want for example to have an entertainer's effects during a golden age to be 1 smiley face and 1 tax and 1 science effect. The code accomodates for that, while being "backwards" compatible. 

Screenshot_12

I chose the most straight forward way to implement this with the dictionaries and such, because I didn't want to do any refactoring regarding the data structures already provided. But again, I would love a discussion on how this could be done otherwise.

Lastly, I made a change regarding the height of the popheads (and thus the effect icons), so now it's all flush with the original and has more breathing room on the bottom.

### DIFF
--- a/C7/Lua/texture_configs/civ3.lua
+++ b/C7/Lua/texture_configs/civ3.lua
@@ -187,21 +187,93 @@ textures.icons = {
     path = CITY_SCREEN_ICONS,
     crop_region = { 34, 2, 30, 30 },
   },
-  luxury = {
+  beaker = {
     path = CITY_SCREEN_ICONS,
-    crop_region = { 376, 2, 30, 30 },
+    crop_region = { 34, 2, 29, 29 },
   },
-  food = {
+  good_gold = {
+	  path = CITY_SCREEN_ICONS,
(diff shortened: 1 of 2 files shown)


TomWerner (approved):
Very nice, thank you!

PR #812, opened 2025-10-02
Add basic tooltips to the unit buttons 

TomWerner:
I'd like to get the turn count for worker moves next.

image

### DIFF
--- a/C7/UIElements/UnitButtons/UnitButtons.cs
+++ b/C7/UIElements/UnitButtons/UnitButtons.cs
@@ -85,22 +85,6 @@ public partial class UnitButtons : VBoxContainer {
 		}
 	}
 
-
-	public static string ActionToTooltipText(string input) {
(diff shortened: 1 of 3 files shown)


C7/UIElements/UnitButtons/UnitButtons.cs
```diff
@@ -84,11 +85,36 @@ private void SetUpTerraformButtons() {
 		}
 	}
 
+
+	public static string ActionToTooltipText(string input) {
```
esbudylin: I think we shouldn't parse tooltip texts from UIAction keys: it makes it impossible to localize them to other languages, and creates an implicit assumption about the format of UIAction keys.  I think a better solution could be making a dictionary that would explicitly associate the UIAction keys with tooltip texts.

TomWerner: Good call, done.

esbudylin (approved):

PR #817, opened 2025-10-06
Issue #799 : Make movement aware of tile owners

### LINKED ISSUE #799: Make movement aware of tile owners
Right now movement cost calculations [here](https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7Engine/C7GameData/AIData/TilePath.cs#L92) do not take into account whether or not the tile is within a civ's borders. That can be found via [this variable](https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7Engine/C7GameData/Tile.cs#L80) in the tile class. Civ A's units shouldn't be able to use Civ B's roads without a right of passage.

Possible bonus work, implement right of passage agreements (though this is a big enough change that it probably warrants a separate issue, as we don't yet have multi-turn trade deals implemented)

stavrosfa:
Besides the tile ownership, I implemented the Engineering effect of bridging rivers with roads. It seemed logical to include it in there since it affects tile movement cost.

### DIFF
--- a/C7Engine/C7GameData/AIData/TilePath.cs
+++ b/C7Engine/C7GameData/AIData/TilePath.cs
@@ -33,7 +33,7 @@ namespace C7GameData {
 			return path != null ? path.Count : -1;
 		}
 
-		public int PathCost(MapUnit unit, Tile from, float perTurnMovePoints, float remainingMovementPoints) {
+		public int PathCost(Player player, Tile from, float perTurnMovePoints, float remainingMovementPoints) {
 			// If we have no path (such as if we are a land unit trying to move to the water)
 			// return -1 so we don't display a goto cursor.
 			if (path == null || path.Count == 0) { return -1; }
@@ -46,7 +46,7 @@ namespace C7GameData {
 
 			foreach (Tile tile in path) {
 				// Subtract the cost of the next move.
-				float cost = GetMovementCost(unit, from, from.directionTo(tile), tile);
+				float cost = GetMovementCost(player, from, from.directionTo(tile), tile);
 				movementPoints.onUnitMove(cost);
 
 				// If we can't do any more moves, bump up the turn cost and reset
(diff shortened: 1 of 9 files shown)


C7Engine/C7GameData/AIData/TilePath.cs
```diff
@@ -89,9 +88,21 @@ public static TilePath EmptyPath(Tile destination) {
 			return new TilePath(destination, new Queue<Tile>());
 		}
 
-		public static float getMovementCost(Tile from, TileDirection dir, Tile newLocation) {
+		public static float GetMovementCost(MapUnit unit, Tile from, TileDirection dir, Tile newLocation) {
```
TomWerner: It looks like the only thing needed from the mapunit here is the owner - could we follow [the law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) and just pass the player in?

C7Engine/C7GameData/Player.cs
```diff
@@ -249,6 +249,31 @@ public bool SitsOutFirstTurn() {
...
+		// since multiturn deals are not yet implemented
+		public bool HasRightOfPassageAgreementWith(Player other) {
+			return false;
+		}
+
+		public static bool CanMoveFreely(Player player, Tile sourceTile, Tile targetTile) {
+			Player targetTileOwner = targetTile.OwningPlayer();
+			Player sourceTileOwner = sourceTile.OwningPlayer();
+
+			// any to any && any to own && own to any && own to own
```
TomWerner: I'd personally find a comment like this a bit easier to understand:

// We are free to move if:
//  - the tile we are on is unowned, or we own the tile
//  - and the tile we are moving to is unowned, or we own the tile

TomWerner (approved):
Very nice, thank you!

C7Engine/C7GameData/AIData/TilePath.cs
stavrosfa: I get your point, but I will have to disagree on that one. How would we implement something like the Privateer as a land unit (like a spy or something)? Or even the Privateer if we decide to make enemy ocean tiles cost more (and this wouldn't apply to the Privateer)?
And as I am writing this, I am thinking of the Marine unit, maybe it has some implications there as well (or similar unit in the future)?

C7Engine/C7GameData/Player.cs
stavrosfa: Yeah, that's better, I just wrote it like this cause it made sense to me in the heat of the moment, I will change it

C7Engine/C7GameData/AIData/TilePath.cs
TomWerner: We would pass some bit of information about the unit if/when we implement something like marines. 

I don't feel super strongly about this, just pointing out that it tends to result in less coupled code.

TomWerner (approved):

Issue #797, opened 2025-09-08
Improve the science advisor UI


TomWerner:
Right now there are two visual-related TODOs:
 - Avoid having the tech name trail off the edge: https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7/UIElements/Advisors/TechBox.cs#L54
 - Figure out how to pick which size tech box to use: https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7/UIElements/Advisors/TechBox.cs#L28

Right now we always pick the small size. A bonus for the multi-size tech box would be to also update our custom art to handle the larger sizes, since right now we only use the small size: https://github.com/C7-Game/Prototype/blob/Development/C7/Art/Advisors/techboxes.png

I'm pretty sure the original game uses things like number of buildings the tech unlocks, worker actions, etc, to pick the size. I don't think it's in the rules, though a modder might be able to tell you more.

PR #819, opened 2025-10-15
Improved Science advisor UI and functionality - Issue #797

stavrosfa:
1. Added size and era appropriate textures for tech boxes.
2. Added tech multiselection/queue functionality
3. Era helper class, should come handy in the future
4. White pop up tooltip for techs
5. Advisor name (probably temp way)

I had gone way further than that, pulling unit and building images with era and culture variations, extended the biq and pediaicons classes, but since there is an ongoing discussion as to how we will structure the data, I removed all that, leaving what I think will stay unaffected.

### DIFF
--- a/C7/Lua/texture_configs/civ3.lua
+++ b/C7/Lua/texture_configs/civ3.lua
@@ -78,75 +78,6 @@ textures.advisors = {
   },
 }
 
-function add_tables(t1, t2)
-	result = {}
-	for i = 1, math.min(#t1, #t2) do
-		result[i] = t1[i] + t2[i]
-	end
-	return result
-end
-
-tech_boxes_texture = ADVISORS .. "techboxes.pcx"
-
--- different era boxes have slightly different sizes in pcx files
--- this is the max size they could be
--- seeing where the "Not required" sign renders in the original as well
--- this is probably how they did it originally
-tbl_small = { 1, 1, 106, 82 }
-tbl_medium = { 1, 84, 163, 82 }
-tbl_large = { 1, 167, 163, 106 }
-tbl_long = { 1, 274, 188, 82 }
-
-local function create_entry(offset)
-	local entry = {
-		small = {
-			path = tech_boxes_texture,
-			crop_region = add_tables(tbl_small, offset),
-		},
-		medium = {
-			path = tech_boxes_texture,
-			crop_region = add_tables(tbl_medium, offset),
-		},
-		large = {
-			path = tech_boxes_texture,
-			crop_region = add_tables(tbl_large, offset),
-		},
-		long = {
-			path = tech_boxes_texture,
-			crop_region = add_tables(tbl_long, offset),
-		},
-	}
-	return entry
-end
-
-textures.tech_box = {
-	known = {
-		ancient = create_entry({0, 0, 0, 0}),
-		middle = create_entry({0, 356, 0, 0}),
-		industrial = create_entry({0, 712, 0, 0}),
-		modern = create_entry({0, 1068, 0, 0}),
-	},
-	in_progress = {
-		ancient = create_entry({189, 0, 0, 0}),
-		middle = create_entry({189, 356, 0, 0}),
-		industrial = create_entry({189, 712, 0, 0}),
-		modern = create_entry({189, 1068, 0, 0}),
-	},
-	possible = {
-		ancient = create_entry({378, 0, 0, 0}),
-		middle = create_entry({378, 356, 0, 0}),
-		industrial = create_entry({378, 712, 0, 0}),
-		modern = create_entry({378, 1068, 0, 0}),
-	},
-	blocked = {
-		ancient = create_entry({567, 0, 0, 0}),
-		middle = create_entry({567, 356, 0, 0}),
-		industrial = create_entry({567, 712, 0, 0}),
-		modern = create_entry({567, 1068, 0, 0}),
-	},
-	non_required = ADVISORS .. "non_required.pcx",
-}
-
 textures.ui = {
   button = {
     inactive = {
@@ -526,6 +457,7 @@ textures.civ_colors = require "civ3.civ_colors"
 textures.unit_icons = require "civ3.unit_icons"
 textures.building_icons = require "civ3.building_icons"
 textures.tech_icons = require "civ3.tech_icons"
+textures.tech_boxes = require "civ3.tech_boxes"
 textures.leader_heads = require "civ3.leader_heads"
 textures.borders = require "civ3.borders"
 
--- a/C7/Lua/texture_configs/civ3.lua
+++ b/C7/Lua/texture_configs/civ3.lua
@@ -78,24 +78,73 @@ textures.advisors = {
   },
 }
 
+function add_tables(t1, t2)
+	result = {}
+	for i = 1, math.min(#t1, #t2) do
+		result[i] = t1[i] + t2[i]
+	end
+	return result
+end
+
+tech_boxes_texture = ADVISORS .. "techboxes.pcx"
+
+-- different era boxes have slightly different sizes in pcx files
+-- this is the max size they could be
+-- seeing where the "Not required" sign renders in the original as well
+-- this is probably how they did it originally
+tbl_small = { 1, 1, 106, 82 }
+tbl_medium = { 1, 84, 163, 82 }
+tbl_large = { 1, 167, 163, 106 }
+tbl_long = { 1, 274, 188, 82 }
+
+local function create_entry(offset)
+	local entry = {
+		small = {
+			path = tech_boxes_texture,
+			crop_region = add_tables(tbl_small, offset),
+		},
+		medium = {
+			path = tech_boxes_texture,
+			crop_region = add_tables(tbl_medium, offset),
+		},
+		large = {
+			path = tech_boxes_texture,
+			crop_region = add_tables(tbl_large, offset),
+		},
+		long = {
+			path = tech_boxes_texture,
+			crop_region = add_tables(tbl_long, offset),
+		},
+	}
+	return entry
+end
+
 textures.tech_box = {
-  known = {
-    path = ADVISORS .. "techboxes.pcx",
-    crop_region = { 1, 1, 180, 80 },
-  },
-  in_progress = {
-    path = ADVISORS .. "techboxes.pcx",
-    crop_region = { 192, 1, 180, 80 },
-  },
-  possible = {
-    path = ADVISORS .. "techboxes.pcx",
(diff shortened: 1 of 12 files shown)


C7/UIElements/Advisors/ScienceAdvisor.cs
```diff
@@ -17,19 +19,41 @@ public partial class ScienceAdvisor : TextureRect {
...
 	private string eraName;
 
+	// store the last opened era window so next time we open the advisor, it opens at the same era window
+	private static string lastOpenedEra = string.Empty;
+
+	private string advisorTitleString;
+
+	private FontFile regularFont = new();
+	private Theme regularBigFontTheme = new();
+	// private Theme regularMediumFontTheme = new();
```
TomWerner: Could you remove the commented out code, here and on line 32?

TomWerner: And 52, 55.

C7/UIElements/Advisors/ScienceAdvisor.cs
```diff
@@ -150,12 +213,35 @@ private void ChangeEraAndDrawTree(int delta) {
...
+				: EraIndexToEra(GetEraIndex(lastOpenedEra) + delta);
 			DrawTechTree(eraName, player, techs, player.GetAvailableTechsToResearch(gameData));
 		});
 	}
 
 	private void ReturnToMenu() {
 		GetParent<Advisors>().Hide();
 	}
+
+	private string GetAdvisorName() {
```
TomWerner: Is this logic turning the class name of "ScienceAdvisor" into the string "Science Advisor"? I think using a string would be better, since that would be more amenable to having a translated title in the future if we wanted.

C7/UIElements/Advisors/TechBox.cs
```diff
@@ -17,54 +40,306 @@ public enum TechState {
...
-		// we should use for a given tech.
-		//
-		// NOTE: this pcx has 16 rows, 4 per era, with different sizes.
-		//
-		// NOTE: the X coordinates of each column were found via guess+check.
-		ImageTexture knownTechBox = TextureLoader.Load("tech_box.known");
-		ImageTexture inProgressTechBox = TextureLoader.Load("tech_box.in_progress");
-		ImageTexture possibleTechBox = TextureLoader.Load("tech_box.possible");
-		ImageTexture blockedTechBox = TextureLoader.Load("tech_box.blocked");
+
```
TomWerner: Could you remove the blank line? It's a little odd at the top of a method

C7/UIElements/Advisors/TechBox.cs
```diff
@@ -17,54 +40,306 @@ public enum TechState {
...
-		};
+		ImageTexture techIconTexture = TextureLoader.Load("tech_icons.small", tech, useCache: true);
+		TextureRect icon = new() { Texture = techIconTexture };
 		icon.SetPosition(new Vector2(12, 32));
 		AddChild(icon);
 
-		// TODO: Have the name trail off if it is too large for the box.
-		Label techName = new() {
-			Text = tech.Name,
+		// we could be calculating this every time based on the font size,
```
TomWerner: Thanks for the comment here! Could you add something like this first?

// Figure out how many characters can fit in the tech box's width.
// We use this for truncation when creating the label below.
//
// We could be calculating .... <rest of comment here>

C7/UIElements/Advisors/TechBox.cs
```diff
@@ -17,54 +40,306 @@ public enum TechState {
...
-		AddChild(techName);
+
+		this.MouseEntered += ShowTooltip;
+		this.MouseExited += UpdateLabelTheme;
+
+		int offsetX = techIconTexture.GetWidth() + 7;
+		int offsetY = 0;
+		int counter = 0;
+
+		// TODO : would be nice to figure out a way to cache these textures.
```
TomWerner: I think as long as we cache the texture loading itself (like we do for most textures already) it should be sufficient. We clear that cache of textures between games already so I don't think the scenario thing is a problem.

C7/UIElements/Advisors/TechBox.cs
```diff
@@ -17,54 +40,306 @@ public enum TechState {
...
+		customTheme.SetStylebox("panel", "TooltipPanel", TemporaryPopup.PopupTechStyleBox());
+		customTheme.SetColor("font_color", "TooltipLabel", Colors.Black);
+		customTheme.SetFontSize("font_size", "TooltipLabel", 12);
+		this.Theme = customTheme;
+		if (TruncatedToFullTechTextMap.TryGetValue(tech.Name, out string fullText))
+			this.TooltipText = $"{fullText}";
+	}
+
+	private void UpdateLabelTheme() {
+
```
TomWerner: same blank line thing here

TomWerner: (and in a couple other methods in this class)

C7Engine/C7GameData/EraUtils.cs
```diff
@@ -0,0 +1,77 @@
+namespace C7GameData;
+
+public static class EraUtils {
+
+	public const string ANCIENT_TIMES_CVLPD = "ERAS_Ancient_Times";
```
TomWerner, line 6: Could you add a comment here, something like "// The civilopedia (CVLPD) version of the era name"? I was wondering what CVLPD meant at first

C7Engine/C7GameData/Player.cs
```diff
@@ -45,6 +45,9 @@ public class Player {
 		// The tech the player is currently researching.
 		public ID currentlyResearchedTech { get; private set; }
 
+		// A queue of technologies the player has specified they want to research
+		public Queue<Tech> ResearchQueue { get; private set; } = new();
```
TomWerner, line 49: In a followup PR it might be cool to:
 - save the research queue (see `SavePlayer.cs`)
 - be able to import the research queue in `ImportCiv3.cs`

Neither is super critical, but it might be worth a TODO here to mention that so we don't forget.

C7/Lua/texture_configs/civ3.lua
```diff
@@ -78,24 +78,73 @@ textures.advisors = {
   },
 }
 
+function add_tables(t1, t2)
+	result = {}
+	for i = 1, math.min(#t1, #t2) do
```
TomWerner: What does `math.min(#t1, #t2)` do here since these are tables?

Ah, after looking more closely I think I see what this is doing.

Would the following work for this function?

```
-- Add the x, y, width, height variables of two different rectangles,
-- to allow us to avoid repeating the sizes of each tech box.
--
-- `offset` is assumed to be a crop_region with a width and height
--   of 0.
--
-- Returns a crop region table.
function combine_offset_with_crop_region(offset, crop_region)
```

TomWerner (approved):
Absolutely fantastic work, thank you! This is really cool, and the tech queue is a great quality of life improvement.

My comments are mostly nits or suggestions on comments - feel free to merge after addressing them.

C7/UIElements/Advisors/ScienceAdvisor.cs
stavrosfa: This is the font I had used for the Current_Era label on the bottom, like<- Previous {Current_Era} Next -> 
I removed that label because after a while I realized that these values come from the .biq file and need to be parsed properly and not harcoded. I left it in there to be ready for when this is done. I can either add a TODO, or remove them for now, I don't really mind

C7/UIElements/Advisors/ScienceAdvisor.cs
stavrosfa: Oh, I didn't even think about translations. I knew it was wrong for other reasons, I even have this in my commit as "5. Advisor name (probably temp way)". Will do

C7/UIElements/Advisors/TechBox.cs
stavrosfa: Fair, I didn't think about that part. I just tested it with Carthage first, and I correctly had the Numidian mercenary as an icon, but then after retiring I switched to another civ, and the mercenary remained instead of having a Spearman. I ll just remove the comment then

C7/UIElements/Advisors/TechBox.cs
stavrosfa: It's the java in me, it doesn't let go!

C7Engine/C7GameData/Player.cs
stavrosfa: I had half a mind to look into that after I was done, but then other things got in the way, and I just forgot it

C7/Lua/texture_configs/civ3.lua
```diff
@@ -78,24 +78,73 @@ textures.advisors = {
   },
 }
 
+function add_tables(t1, t2)
```
esbudylin: Note that you are declaring the function `add_tables` and the following variables as globals: in Lua, the variable is considered global, unless the `local` statement was used for its declaration. It would be cleaner to use locals instead.

Also, since tech box configuration now uses specific utility functions and variables, I think it would make sense to separate this piece of configuration into a new file (similarly to how it is done with popheads, for example).

esbudylin (commented):
Very impressive work!

> I had gone way further than that, pulling unit and building images with era and culture variations, extended the biq and pediaicons classes, but since there is an ongoing discussion as to how we will structure the data, I removed all that, leaving what I think will stay unaffected.

I think you should go for it and submit these changes for review too :) As I see it, the recent discussion was mostly hypothetical, I don't think most of its points will materialize any time soon.

C7/Lua/texture_configs/civ3.lua
stavrosfa: I refactored it based on your suggestions. Please do have a look, I am extremely new to lua, i am still stumbling my way through it, so I will probably make a lot of weird things there.
Also, is there a way  to reuse ADVISORS, I don't like the fact that I am duplicating code there, but I don't know how else to do it without making it global. Perhaps if we moved these "constant" variables to their own file and import them accordingly?

C7Engine/C7GameData/Player.cs
stavrosfa: I was looking into it, and it didn't take much time (or code for that matter) to complete.
Since this MR is not closed yet, I figured I would commit it as part of this one.

C7/Lua/texture_configs/civ3.lua
esbudylin: The new Lua code looks good to me! You could move the string constants into a new file, but I don't find it necessary, since the code duplication is minimal.

esbudylin (approved):

PR #820, opened 2025-10-22
Intergrate unit assets

esbudylin:
This PR
- Removes the C7 assets from the repository and adds the [assets repository](https://github.com/C7-Game/Assets) as a submodule.
- Introduces a new json save for the standalone mode. It excludes the units we don't have graphics for and updates art names for units that got an art replacement.
- Changes the BarbarianInfo class: previously, when loading data on barbarian units from json, this class relied on a unit's position in the list of all units. This PR makes it rely on a unit name instead, which makes it easier to make changes to the unit list.

There are a few issues with the standalone mode:
- Some unit sprites are displayed with wrong offsets. I've noticed this issue with warrior, horseman, and curragh, but I haven't tested all the units in game.  
- There are no tech box textures for different sizes and eras.

A screenshot of the game running standalone:
image_2025-10-22_20-50-18


TomWerner (approved):
Nice, just a couple questions:
 - Do we need to update the dev instructions? I had to run `git submodule init` and `git submodule update` to get the assets loaded
 - Should we update `SaveTest` to run with the old and new save files?

WildWeazel:
I was wondering whether we should bother with the submodule, and leaning towards not:

- As long as Assets contains only media files, it allows us to ignore it in most CI jobs
- It saves a few tedious steps when pushing updates, which artists contributing to Assets would need additional permissions to perform
- Since I don't see any reason to bother with branches in Assets anyway, there's no need to track specific commits. Just checkout/pull the latest Assets if you need them.

What do you guys think?

esbudylin:
@WildWeazel I don't think the choice between the submodule and the monorepo is a big deal, but I'm leaning towards the submodule option. I think keeping the assets in the separate repo will make development easier for devs not interested in running the standalone version of the game. For example, I usually run the game with Civ3 assets, so when pulling new changes from the origin, I wouldn't want the changes to the code to be mixed with assets I won't even see in-game. Also, separating assets in a submodule will save time for initial checkout of the repo: it will allow a developer to clone the full history of the main repo, without cloning all the changes to asset files. 

I've done some additional work on this PR (added the standalone to SaveTest and fixed loading of new unit assets in a debug build). I'll add submodule instructions to the wiki as soon as this PR is merged. In case there are no objections, I'm going to merge this in a day or two.

WildWeazel:
Oh I didn't mean to go back to a monorepo, just that we wouldn't need to track the submodule from here. If you want the assets, just clone them manually. I see now that there's a simple CI setting to toggle submodules off and we don't have any others, so my first point is probably moot. I agree it's not a big deal either way.

esbudylin:
I've created a wiki page documenting the difference between standard and standalone modes and how to set up a submodule: https://github.com/C7-Game/Prototype/wiki/Standard-and-standlone-modes I'll go ahead and merge this. I think if the submodule becomes unhandy, we can return to this discussion.

Issue #821, opened 2025-10-23
Visual improvements for Fog of war and various terrain pieces


stavrosfa:
There are very clearly some nasty visual artefacts, like dots from the pcx file of FogOfWar, and some of the background (the green part of the pcx ) of things like forests, and pretty much anything that uses both green and magenta as alphas and shadows are not turned off (they are on by default).
Here is an image of what it currently looks 

Fog of war

<img width="623" height="462" alt="Image" src="https://github.com/user-attachments/assets/755c2468-9b7e-4d35-8a16-687b283094ed" />

Exagerated (only so compression doesn't mess it up) terrain where the outline and background are clearly bleeding

<img width="1147" height="768" alt="Image" src="https://github.com/user-attachments/assets/fdc81168-1147-4b1d-822e-9d4476646e6a" />

PR #822, opened 2025-10-23
Visual improvements for Fog of war and various terrain pieces | Issue #821

stavrosfa:
This PR tackles mostly visual bugs that, except for the actual texture file for C7, come from the code itself.

I have left a pretty lengthy comment describing the change for the pure_alpha issue, so I won't reiterate that here.

Regarding the shadows part, although we take out 255, 254 (usually magenta and green), the shadows are on, and it seems to be using the green part as the shadows part, resulting in the texture background bleeding through. 
At least, that's what has worked for me, removing the shadows from these textures. Let me know if you already know of a similar case for something I haven't included.

I also created issue #821  so that it will be more easily searchable in the future as an issue if anything ever comes up


TomWerner (approved):
Thanks! You might want to check with @esbudylin to make sure the fog of war png change makes it into the new assets repo too.

stavrosfa:
I thought to mention it, but then I saw that there are only units included. Let me know if I can do something more.

PR #823, opened 2025-10-26
Go-to line, AI pathfinding modifications

stavrosfa:
1. Refactored the go-to layer, to show path lines between tiles.
2. The pathing algorith has been modified, now AI can see the whole map (as before) but the human player can't.
3. Account for narrow land strips in map for water units (see before below)

c7_sea_travel

and afer

image

5. Grid layer is now computed by itself, before it was rendering only known tiles, resulting in missing lines etc. The tiles for the grid are already being computed for the other layers, so it's not like we add a huge overhead or something.

I wanted to include a "fixed" cursor animation, but I can't get it to render properly (civil disorder suffers too), and I 've tried for like hours and hours.
I don't know if it's the shadows, or the tint, or the overall way it's being extracted from the flc.
Someone smarter than me can understand it and fix it.

### DIFF
--- a/C7/Map/GotoLayer.cs
+++ b/C7/Map/GotoLayer.cs
@@ -1,8 +1,6 @@
 using System;
 using System.Collections.Generic;
 using C7GameData;
-using C7Engine;
-using ConvertCiv3Media;
 using Godot;
 
 // The layer responsible for drawing the cursor when the player is selecting a
@@ -16,7 +14,7 @@ public partial class GotoLayer : LooseLayer {
 		//
 		// We use FixedSize so Godot can calculate the width of the text for centering.
 		gotoLabelFont = ResourceLoader.Load<FontFile>("res://Fonts/NotoSans-Regular.ttf", null, ResourceLoader.CacheMode.Ignore);
-		gotoLabelFont.FixedSize = 25;
+		gotoLabelFont.FixedSize = 20;
 
 		whiteFontTheme.DefaultFont = gotoLabelFont;
 		whiteFontTheme.SetColor("font_color", "Label", Colors.White);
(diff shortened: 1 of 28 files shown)


C7Engine/AI/Pathing/AStarAlgorithm.cs
```diff
@@ -76,6 +81,10 @@ public override TilePath PathFrom(Tile start, Tile destination) {
 				foreach (Edge<Tile> neighborEdge in edgeWalker.getEdges(current)) {
 					Tile neighbor = neighborEdge.current;
 
+					// TODO: the possibility here is to create a building/improvement/tech effect like "canal"
+					// (much like what the Panama canal is irl) that allows that kind of movement
+					if (unit.IsWaterUnit() && IsLandStrip(current, neighbor)) continue;
```
TomWerner: Does this break canal cities?

TomWerner: Edit, I checked basic canals and they work, but would they work in the case from the land strip function?

C7Engine/AI/Pathing/EdgeWalker.cs
```diff
@@ -14,23 +14,24 @@ public UnitWalker(MapUnit unit) {
...
 				Tile neighbor = pair.Value;
 				bool neighborHasCityWithSameOwner = neighbor.cityAtTile != null && neighbor.cityAtTile.owner == unit.owner;
 
+				bool isPassable = false;
 
-				// Land units can only go on to land.
-				if (landUnit && !neighbor.IsLand()) {
-					continue;
+				if (unit.owner.isHuman && !unit.owner.HasExploredTile(neighbor)) {
+					isPassable = true;
```
TomWerner, line 26: This bit is very cool and faithful to the original game - could you call it out in the PR description?

C7/Map/GotoLayer.cs
```diff
@@ -73,15 +75,49 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
...
+				// The visible region is only what you see at the screen plus the tiny buffer.
+				// These are the only tiles the player has seen and calculated the centers of, so far.
+				// If we try to draw lines between tiles that are outside this visible region, we will
+				// get an error because we don't know yet what these centers are. We either have to move the camera
+				// and calculate them, or precompute a huge buffer of tiles which is not practical at all.
+				if (looseView.tileCenters.TryGetValue(currentTile, out Vector2 currentTileCenter)
+					&& looseView.tileCenters.TryGetValue(nextTile, out Vector2 nextTileCenter)) {
+					staticCursorRect?.Hide();
+					looseView.DrawLine(currentTileCenter, nextTileCenter, Colors.Red, width: lineWidth);
+					// drawGotoCursor(looseView, nextTileCenter, gotoInfo.moveCost, gotoInfo.attackingMove);
```
TomWerner: Remove?

TomWerner (approved):
Thanks for the PR!

I did a couple of play tests and read most of the code, but I'll confess I didn't fully digest all of the map view/gotolayer changes.

On that note, it can be nice sometimes to have more focused PRs, for ease of reviewing/culprit finding. I think the pathfinding changes could probably have been their own PR (not a big deal at all - I'd rather have this great PR as one unit than no PR at all, but just some general feedback).

I found that it's significantly easier to do this with https://github.com/jj-vcs/jj, since `jj split` lets you very easily break up one commit into two stacked ones.

C7Engine/AI/Pathing/AStarAlgorithm.cs
stavrosfa: If by canal cities you mean a city a water unit can go through, no it doesn't.

A simple one

image

And another, where the boat would never be able to get to because of the new restrictions.

image


The LandStrip() method is only for sea-to-sea movement anyway

C7Engine/AI/Pathing/EdgeWalker.cs
stavrosfa: Will do, thanks!

Actually, doesn't point no. 2 cover it?

stavrosfa:
Sorry, I know it's kind of a messy PR, and I had water unit sea/ocean restrictions, wonders and techs that allow it, and other small things here or there, but I decided to keep it out of this one.
I will try to have more focused PRs in the future, thanks for the feedback!

Also, I think I just found a bug, so let's hold this off for now, I will try to even write some more unit tests.

TomWerner (approved):

Issue #826, opened 2025-10-31
Fix unit placement on tiles and Animation Speed


stavrosfa:
Right now the unit are positioned based on an hardcoded approximation.

To correctly place the units, we need to read some data from the .flic file of each animation that provide the correct X and Y offsets.

The .flic files also contain information on how fast the playback of each animation is, so we need to read that data as well. The current approach is also based on an hardcoded approximation.

PR #827, opened 2025-10-31
Fix unit position on tile and animation speed - Issue #826

stavrosfa:
1. Read offsets and animation info from .flic file
2. Calculate units' positions taking into account these offsets
3. Apply animation speed based on the .flic file

Issue #826

### DIFF
--- a/C7/Animations/AnimationManager.cs
+++ b/C7/Animations/AnimationManager.cs
@@ -232,7 +232,14 @@ public partial class C7Animation {
 
 	public double getDuration() {
 		Util.FlicSheet flicSheet = getFlicSheet();
-		double frameCount = flicSheet.indices.GetWidth() / flicSheet.spriteWidth;
-		return frameCount / 20.0; // Civ 3 anims often run at 20 FPS   TODO: Do they all? How could we tell? Is it exactly 20 FPS?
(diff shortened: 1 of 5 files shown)


WildWeazel (approved):
For some reason things are all borked when I try to build this PR with the Assets, but the CI build looks good and it seems well written so I'm approving on the assumption that it will all work out once we merge it together :)

stavrosfa:
I tried this on Yegor's branch with the new assets and it seemed to work as expected.
But someone else apart from me should verify this as well before merging, although it should work fine with or without the new assets.

esbudylin:
I can confirm that the offset fix works well with the new assets, but there is something off with the speed of some animations. Settler's animation for founding a city and worker's animations for mining and irrigating are now played at much slower speed than in the original game.

stavrosfa:
If you are referring to the civ3 assets, I had the same feeling, that now they run much slower than they should have. So I placed a worker mining & roading right next to an actual Civ3 Conquests worker doing the same, and they moved at the same speed exactly.
Now if you are referring to the c7 assets, I am not so sure..
Also, could you share your worker's ini file?

esbudylin:
Yes, I was talking about the civ3 sprites. I've re-checked the original game, and yes, I was remembering this wrong, these animations are that slow in the original game too. Well, then let's go ahead and merge this!

PR #828, opened 2025-11-02
#780 Update credits and project metadata for Dutch

### LINKED ISSUE #780: Update credits for the next release
Just making sure we don't forget this. No need to merge until just before preparing a release. 

As we vastly increase the custom assets I imagine we'll want to find a more concise format for those credits. I'm thinking we just have a section listing artist names similar to developers, and leave the individual accreditation for the source files (see #563 for my thoughts on that)

WildWeazel:
- Updated credits with new contributors, additional thanks, and simple list of all artists whose work is used (individual art credits will be part of Assets)
- Condensed the project summary in README to mostly match the updated homepage
- Renamed all non-source references to C7, including the executables


WildWeazel:
I see the renamed export variable is causing problems where it does not match the name of the directory the project is in. We could rename just this directory to make it more clear that it's the main entry point (I don't want to bother with mass replacing "C7" in source), but is there any reason not to just have a hard coded `cd` here? I don't see why the directory we're building should need to be a variable.

WildWeazel:
Since there were some other instances I went with a half measure and introduced a new variable PROJECT_DIR to distinguish between where the source is and the name of the executable.

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 C7-Game
+Copyright (c) 2021-2025 The OpenCiv3 Project
```
esbudylin: Wouldn't this change violate our own license? The license requires to include the copyright notice "in all copies or substantial portions of the Software", so I guess we can't just remove the old notice without all the contributors agreeing. I think a safer way would be to add a new notice below the old one.

WildWeazel: Hm, you mean in the sense that we'd no longer be carrying the copyright notice that the code was originally published under? I was going to discuss renaming the GitHub organization after this. The entity "C7-Game" only "exists" as the name of that account, which doesn't even include all contributors. Maybe it should have two lines, 2021-2025 C7-Game and 2025- OpenCiv3. I'll have to do some more digging. Surely this is a known scenario.

WildWeazel:
This one went a bit out of scope, but I think I got everything :D

C7/Text/credits.txt
```diff
@@ -1,33 +1,56 @@
 [color=black][center]
-[b]C7 0.2 "Carthage"[/b]
-[url]c7-game.github.io[/url]
+[b]OpenCiv3 0.3 "Dutch" Preview 1[/b]
+© The OpenCiv3 Project 2021-2025
```
esbudylin: There is a mismatch between the copyright notice here, which credits the OpenCiv3 Project, and the notice in the License file that credits OpenCiv3 contributors. I guess this one just wasn't updated? Everything else LGTM.

esbudylin (approved):

C7/Text/credits.txt
WildWeazel: Good catch!

PR #830, opened 2025-11-04
1. Update units' HP bar

stavrosfa:
Supports 2 camera zoom options.
Supports large HP levels (such as a 24HP army and beyond). Smaller hp fraction lines are drawn the more max hp the unit has. Now we can only see our units' movement status.
Also, no lines under enemy units are drawn to indicate how many there are on a tile.

A few more minor 1-liner type of things

Cursor Animation now plays below the unit (no side effects were noticed)
Pass current player controlling the UI and not the first human when getting visible units to draw

### DIFF
--- a/C7/Map/UnitLayer.cs
+++ b/C7/Map/UnitLayer.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using C7GameData;
 using Godot;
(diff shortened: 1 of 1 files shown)


esbudylin (approved):

TomWerner:
> Also, no lines under enemy units are drawn to indicate how many there are on a tile.

In the original game you can see how many enemy units are on a tile if:
 - the tile is visible
 - the units aren't on a city

PR #831, opened 2025-11-05
Terraform AI scoring

### LINKED ISSUE #814: Worker Automation Bug
Worker on 1st turn when automated walks one tile NW of the capital to the forest and immedietly we are awarded 10 shields for the forest clearing. 

The forest texture also remains as is. 

The worker on the next turn starts building a road (just pointing out that the worker wasn't going to clear the forest in the first place). 

We also get a warrior unit because of the shields.

The seed I reproduced it with may times in a row:  **1197716345**

To reproduce, enter the seed, build the capital, exit the city screen, press A for the selected worker and that's it.

Image

Image

Image

esbudylin:
This commit updates the AI for choosing the best worker action (terraform) for a tile.

Previously, the AI scored terraforms by making a copy of a tile, applying a terraform to it and comparing yields of the original tile and its copy. This system had problems. If the terraform had an effect on the world outside of a tile (for example, forest chopping granting shields to the nearest city), these changes were mistakenly applied during the AI score calculations. There was also no way to account for such effects in AI calculations.

This commit fixes that by extending the Terraform class with a new Lua "hook" for AI score calculations. This is a function bound to a terraform that returns its AI points. This way, the effect of the terraform is run only on its completion. This also allows implementing custom AI for new terraforms without modifying the game core.

closes #814

### DIFF
--- a/C7/Lua/rules/civ3/terraforms.lua
+++ b/C7/Lua/rules/civ3/terraforms.lua
@@ -1,40 +1,102 @@
-local UnitAction = ENUMS.UnitAction
 local YieldType = ENUMS.Tile_YieldType
+local ResourceCategory = ENUMS.ResourceCategory
 local Civ3FoliageAction = ENUMS.TerrainType_Civ3FoliageAction
 
-local terraforms = {
-  mine = {
-    validator = function(context)
-      local mine = context.terraform.Improvement
-      return mine:GetYieldBonus(context.tile.overlayTerrainType, YieldType.Production) > 0
-    end,
-  },
+local terraforms = {}
 
-  irrigate = {
-    validator = function(context)
-      local irrigation = context.terraform.Improvement
-      return context.tile:CanBeIrrigated(irrigation, context.player)
-    end,
-  },
-
-  clear_wetlands = {
-    validator = function(context)
-      return context.tile.overlayTerrainType.allowedFoliageAction == Civ3FoliageAction.ClearWetlands
-    end,
-    effect = function(context)
-      context.tile:ClearTerrainOverlay()
-    end,
-  },
-
-  clear_forest = {
-    validator = function(context)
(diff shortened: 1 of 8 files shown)


C7/Lua/rules/civ3/terraforms.lua
```diff
@@ -1,40 +1,102 @@
...
+  return clear_wetlands_points
+end
+
+function terraforms.ai_score.clear_forest(context)
+  if context.tile.hasHadForestCleared then
+    return 0
+  end
+
+  -- TODO: Take the city's current production into account.
+  -- There is no sense in chopping the forest, if the chopping bonus would be lost to overflow
```
TomWerner, line 64: similarly if there's no city within range (though the AI worker logic might avoid that already)

C7/Text/c7-static-map-save-standalone.json
```diff
@@ -74761,6 +74765,7 @@
         "Coal"
       ],
       "improvement": "railroad",
+      "aiScore": "terraforms.ai_score.default",
```
TomWerner: Maybe leave a TODO that this should be implemented? Railroads boost yield and have high strategic value

TomWerner (approved):
Very nice! It's cool to see things like the AI logic starting to be scriptable.

C7/Lua/rules/civ3/terraforms.lua
esbudylin: Thanks for the review! I think currently workers don't perform terraforms on neutral/foreign lands, so it shouldn't be an issue.

C7/Text/c7-static-map-save-standalone.json
esbudylin: Good point, I added some rudimentary logic for railroads, so we can improve it later.

PR #833, opened 2025-11-11
Lower right info box improvements

stavrosfa:
1. Added unit thumbnail in the lower right info box
2. Now it also displays the unit's rank, and also if the unit is in a city, the name of the city instead of the underlying terrain info
3. Clicking on it will center on the currently active unit
4. Clicking on it at the end of the turn will end the turn

To be honest this one felt a bit "hacky",  for lack of a better word. If you have any suggestions as to how to improve it, let me know.

### DIFF
--- a/C7/Animations/AnimationManager.cs
+++ b/C7/Animations/AnimationManager.cs
@@ -40,6 +40,13 @@ public partial class AnimationManager {
 		return AnimationKey(BaseAnimationKey(unit, action), direction);
 	}
 
+	public static readonly Dictionary<string, ImageTexture> AnimationThumbnails = new();
+	public static readonly Dictionary<string, ImageTexture> AnimationTintThumbnails = new();
+
+	private const int thumbnailFrame = 0;
+	private const TileDirection thumbnailDirection = TileDirection.SOUTHEAST;
+	private const MapUnit.AnimatedAction thumbnailAction = MapUnit.AnimatedAction.DEFAULT;
+
 	private AudioStreamPlayer audioPlayer;
 
 	public SpriteFrames spriteFrames;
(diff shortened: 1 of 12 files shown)


TomWerner:
Very cool! Sorry for the slow review, missed the email. 

My one comment is that I'd rather not duplicate the "get the first frame of the animation" logic that we already have in the city screen: 
https://github.com/C7-Game/Prototype/blob/370b55db9c9f9789975ffcd13f1934876fcc3230/C7/UIElements/CityScreen/CityScreen.cs#L597-L627

Could you share that code for the lower right hand box?

stavrosfa:
No worries, I though we were in some kind of freeze before the release. 

Regarding what you said, you are absolutely right, but I think instead of using the city screen logic in the lower right hand box,  the city screen method should call the method in the AnimantionManager.
Maybe that's what you meant all along, because calling the cityScreen from the LowerRightInfoBox doesn't make sense to me.

C7/Animations/AnimationManager.cs
```diff
@@ -40,6 +40,13 @@ public static string AnimationKey(UnitPrototype unit, MapUnit.AnimatedAction act
 		return AnimationKey(BaseAnimationKey(unit, action), direction);
 	}
 
+	public static readonly Dictionary<string, ImageTexture> AnimationThumbnails = new();
```
TomWerner, line 43: Can we ensure these caches are cleared when we clear the other caches? Otherwise we might carry things over between scenarios or when switching the textures being used.

TomWerner (approved):

TomWerner:
> Maybe that's what you meant all along, because calling the cityScreen from the LowerRightInfoBox doesn't make sense to me.

Yeah, I should have been clearer sorry. I just wanted to make sure the code was in some common location. I agree it wouldn't make sense for the lower right box to depend on the city screen.

stavrosfa:
@TomWerner Don't hate me, I know I drifted a bit from the described task. 😅
But, I was working to actually update the worker buttons when they move, and after I was done, I figured using what I just did, I could get rid of some hacky stuff I didn't like here, and so I merged them into one instead of opening a new PR with a few lines of code.

TomWerner:
Still looks good, thanks for the improvements!

PR #834, opened 2025-11-12
Apllied shadows to resources' textures

stavrosfa:
Basically what the title says.

An alternative way of doing this that I tested, instead of exposing a public getter for **ModernGraphicsActive** and checking it in the ResourceLayer, would be to add an empty texture so Lua doesn't freak out, and render a cropped region of nothing (since newer graphics can simply contain their shadows in the same image) when the new graphics are enabled, which I didn't like, it's just wasted draw calls.

Maybe a boolean in Lua like "Civ3Exclusive" would be helpful, but then the Texture.Load() would have to return null or something that doesn't break everything and can be checked by the caller merhod in order to skip the draw call.

Let me know what you think.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -60,7 +60,7 @@ public partial class Game : Node {
 	private Vector2 OldPosition;
 
 	Stopwatch loadTimer = new Stopwatch();
-	GlobalSingleton Global;
+	public GlobalSingleton Global { get; private set; }
 
 	[Export]
 	private PopupOverlay popupOverlay;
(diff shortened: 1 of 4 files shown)


WildWeazel:
Hm, interesting point. I don't love the rendering classes knowing which mode we're in. It seems like that should be solved with polymorphism. My gut reaction was to just pass null, but I see the API won't accept that. Maybe we're solving this at the wrong level? Like what if we abstract away the two images and produce one combined argb texture when we read the resource files? That would imply a more contextual wrapper around PcxToGodot, which seems like it could be useful.

Anyway I'm not totally opposed to doing it this way for now, but I'm curious if anyone else has an opinion.

stavrosfa:
This is a very minor PR in terms of my "work" and what it offers as an end result, so I don't mind at all not merging it and in turn come up with a solid solution that will work with everything else.
In fact if we just merge this, I think it's going to create more tech debt that will not be looked at for another year, if it ever will.

PR #837, opened 2025-11-24
Terrain improvements enchancements

stavrosfa:
1. Terrain Improvements of the same type can replace each other, but no downgrades are allowed.
2. Ask for confirmation when we are about to replace an improvement


TomWerner (approved):
Thanks! 

By the way, feel free to merge your own PRs once they've been approved, you should have permission to do that.

PR #838, opened 2025-11-26
new title & icon graphics

WildWeazel:
Made a title banner (approved by consensus on Discord) to replace the old embedded text on the main menu, and a very placeholder icon until we can come up with something iconic. Also trimmed down the menu options to only show those that are hooked up.

image


TomWerner (approved):
Very nice - is there any way to change the color of the "dutch preview" bit to be a less shocking red color though?

WildWeazel:
The color is set in the scene's theme override, I just took the Dutch-est orange I could get. I'm afk now but feel free to change it if you feel strongly.

PR #839, opened 2025-11-30
Border tile assignment improvements

stavrosfa:
1. Bug fix where an ocean tile cannot be assigned if the city is at expansion level greater than 2.
2. Added more conditions to how a tile ownership is determined, and how/if unassigned tiles are correctly assigned to a city.

// Ocean tiles may only hold claims of rank 2.
The current version has a bug, it should check if the tile is more than rank 2, not the city's current expansion level. 
Case study: (Scenario) WWII In the Pacific
City: Manokwari (Above Australia)
Tiles: [37, 71] [38, 72] should belong to the city

You can also see this with Tokyo as well, it skips a row of tiles because its expansion level is more than 2, but the tiles themselves aren't.

I have tested this mostly with the Conquests scenarios that come with the game. There a few minor discrepancies, but I think they are temporary, as you play either game (openCiv3/original), they iron out.

### DIFF
--- a/C7Engine/C7GameData/Tile.cs
+++ b/C7Engine/C7GameData/Tile.cs
@@ -616,6 +616,28 @@ namespace C7GameData {
 			return map.tileAt(XCoordinate + xDelta, YCoordinate + yDelta);
 		}
 
+
+		/// <summary>
+		/// <para>
+		/// Walks clockwise/counter-clockwise the nth ring around
+		/// the specified tile starting on the northmost tile
+		/// and tries to find the first tile that matches our boolean criterion.
+		/// </para>
+		/// <para>
+		/// This differs from <see cref="GetTilesWithinRankDistance"/>,
+		/// because it includes all the tiles regardless of the distance.
+		/// An example would be that GetTilesWithinRankDistance() with a rank of 2
+		/// will not return a NN, SS, WW, or EE tile, whereas this method will.
+		/// </para>
+		/// <para>
+		/// It is mostly used to calculate to whom we should assign tiles
+		/// that are being claimed by more than 1 city or civilization.
+		/// </para>
+		/// </summary>
+		/// <param name="rank"></param>
+		/// <param name="predicate"></param>
+		/// <param name="clockwise"></param>
+		/// <returns></returns>
 		public Tile FindInRing(int rank, Func<Tile, bool> predicate, bool clockwise = true) {
 			Tile startingTile = map.tileAt(this.XCoordinate, this.YCoordinate - (2 * rank));
 
--- a/C7Engine/C7GameData/Tile.cs
+++ b/C7Engine/C7GameData/Tile.cs
@@ -186,9 +186,13 @@ namespace C7GameData {
 		/// This is used by some graphics algorithms.
 		/// </summary>
 		/// <returns></returns>
-		public Tile[] getEdgeNeighbors() {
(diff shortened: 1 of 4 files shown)


C7Engine/C7GameData/Tile.cs
```diff
@@ -612,6 +616,37 @@ public Tile GetTileAtNeighborIndex(int neighborIndex) {
 			return map.tileAt(XCoordinate + xDelta, YCoordinate + yDelta);
 		}
 
+		public Tile FindInRing(int rank, Func<Tile, bool> predicate, bool clockwise = true) {
```
TomWerner, line 641: Do we need this in addition to GetTilesWithinRankDistance, which returns things in spiral order?

TomWerner (approved):

C7Engine/C7GameData/Tile.cs
stavrosfa: GetTilesWithinRankDistance returns tiles in a big fat cross style. I tried to use it, but for borders, although it's fine for the first rank, the 2nd rank doesn't include a NN, WW, EE or SS tile. This method goes around the "actual" ring around a tile. I called it ring to distinguish it from rank, which behaves differently.
I hope that makes sense.

TomWerner: Makes sense, thanks! Could you put that in a comment on the method to explain how it's different and why we need it?

stavrosfa:
Played about 2 games of 200 rounds each, skipping animations etc, only building Temples, Libraries and whatever would give me culture points, did not come across any after the latest fix.
Tom, let me know when you have a look again, I'd like your OK again before I merge it.

C7Engine/C7GameData/Tile.cs
```diff
@@ -612,6 +616,69 @@ public Tile GetTileAtNeighborIndex(int neighborIndex) {
...
+			// Going NW(counter-clockwise) or NE(clockwise)
+			for (int _ = 1; _ < (2 * rank); _++) {
+				if (clockwise) { dx++; dy--; } else { dx--; dy--; }
+				if (!IsInValidCoordinates(dx, dy, out Tile currentTile) || !predicate(currentTile)) continue;
+				return currentTile;
+			}
+			return null;
+		}
+
+		private bool IsInValidCoordinates(int dx, int dy, out Tile currentTile) {
```
TomWerner: Hmm, apparently adding a comment on a commit doesn't show up in the review (the GitHub code review tool suuuuuucks)

---

A couple of questions from my comment on the commit:

 - Why "dx" and "dy" if these are absolute positions, not deltas?
 - Is the method name "is in valid coordinates" or "is invalid coordinates"? Would "GetTileIfValid" work, and have the return type be Tile??
 - How is this function different from map.tileAt which I believe does the tile wrapping for us

stavrosfa: Sorry I missed the part where map.tileAt was wrapping, to be honest I just checked that it returns a Tile.None and not a null.
Thanks for the feedback!

TomWerner (approved):

stavrosfa:
I found that for the tiles that are opposite to each other and need to be filled in, Law II (Ocean tiles may only hold claims of rank 2.) was not being applied.
Because a picture will explain it better than I could :

OeanTiles

Top row is the original game, bottom left what we had before and on the right what we have after this fix.
The tile with the X should not be part of our city, because its rank is larger than 2.

I refactored the method that was calculating because there was too much duplication, really without any good reason.

I had to also refactor the ResolveTileOwnershipConflict method, because we somehow have to assign null to a tile that doesn't belong to a city and the bool/out City setup made more sense to me.

TomWerner (approved):

PR #840, opened 2025-12-02
Fixed bugs in the lower right info box behaviour

stavrosfa:
1. Fixed bugs in the lower right info box behaviour (initial PR https://github.com/C7-Game/Prototype/pull/833) where labels are overlapping and units' info doesn't update as it should in certain scenarios.
2. Units are waken up when selected.
3. When ending the turn the currently selected unit gets reset.

All changes are pretty much concerned with visuals, of how and when the box gets updated, except of course 2 and 3, but they are pretty much 1 line each.

Number 2 and 3 were added so that the box knows if a unit is awake or nothing is selected, instead of adding some hacky behaviour to the box.

A lot of code has moved around, but it's mostly to accomodate the order of operations rather than changing the behaviour.

The only thing I am sceptical about is the waking of a newly selected unit, although I didn't see anything weird.


TomWerner (approved):
I didn't extensively playtest this, but I clicked around on some units and hit "W" to cycle and things seemed reasonable.

stavrosfa:
What we had before was fine for simple actions. It was failing when multiple units were in play, worker automations were going on, and partly because of the asynchronicity of the units' movement. That's why I missed it in the first place.

PR #842, opened 2025-12-08
Hotkey and UI fixes

stavrosfa:
1. Unit animations can be toggled with Caps Lock
2. The map can't be dragged when the mouse is over UI elements.
3. Unit animations can be temporarily be switched on and off by holding down Shift
4. Refactored actions that are mapped to G (single and combo)
5. Fix disband unit throwing an exception when no unit is selected

For caps lock sadly there is no API in Godot (yet, there is a PR I think that fixes that but hasn't been merged) to check if caps lock is on or off. So for now, simple toggling will do.

For holding down shift to disable animations, I had to create a method like ProcessAction, that handles the release of the keys so that we know to stop disabling the animations.

I couldn't find a way to better handle the G presses, as it was triggering both go-to and toggle grid if CTRL+G (or CTRL+SHIFT+ALT+G) was pressed. And after a point it would stop triggering the grid action.
So I removed the mapping that was done inside Godot, and handle the seperate cases manually.

### DIFF
--- a/C7/Animations/AnimationController.cs
+++ b/C7/Animations/AnimationController.cs
@@ -49,9 +49,6 @@ public partial class AnimationController : Node {
 					mSEA.markCompleted();
 				}
 				break;
-			case MsgStartStopAllAnimations mSSAA:
-				animTracker.endAllImmediately = !mSSAA.animationsEnabled;
-				break;
 		}
 	}
 
@@ -67,9 +64,11 @@ public partial class AnimationController : Node {
 
 	public void ToggleAnimationsEnabled() {
 		new MsgToggleAnimationsEnabled().send();
+		animTracker.endAllImmediately = !animTracker.endAllImmediately;
 	}
 
 	public void SetAnimationsEnabled(bool enabled) {
 		new MsgSetAnimationsEnabled(enabled).send();
+		animTracker.endAllImmediately = !enabled;
 	}
 }
--- a/C7/Animations/AnimationController.cs
+++ b/C7/Animations/AnimationController.cs
@@ -1,4 +1,3 @@
-using System;
 using C7Engine;
 using C7GameData;
 using Godot;
@@ -50,6 +49,9 @@ public partial class AnimationController : Node {
 					mSEA.markCompleted();
 				}
 				break;
+			case MsgStartStopAllAnimations mSSAA:
+				animTracker.endAllImmediately = !mSSAA.animationsEnabled;
+				break;
 		}
 	}
 
@@ -63,8 +65,11 @@ public partial class AnimationController : Node {
 		animTracker.update();
 	}
 
+	public void ToggleAnimationsEnabled() {
+		new MsgToggleAnimationsEnabled().send();
+	}
+
 	public void SetAnimationsEnabled(bool enabled) {
 		new MsgSetAnimationsEnabled(enabled).send();
-		animTracker.endAllImmediately = !enabled;
 	}
 }
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -445,7 +445,9 @@ public partial class Game : Node {
 
 	private void HandleLeftMouseButton(InputEventMouseButton eventMouseButton) {
 		GetViewport().SetInputAsHandled();
-		if (eventMouseButton.IsPressed()) {
+		Control uiHover = GetViewport().GuiGetHoveredControl();
+		// Can't drag the map when the mouse is over a ui element
+		if (eventMouseButton.IsPressed() && uiHover is not TextureButton) {
 			OldPosition = eventMouseButton.Position;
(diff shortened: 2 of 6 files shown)


TomWerner (changes requested):
I wasn't able to get the shift/caps lock thing working for skipping animations, but I was testing on a Chromebook without a real caps lock key. So that would explain the caps lock thing.

How should the holding shift thing work? If I start a unit moving and then press shift, should it disable animations?

stavrosfa:
I can see that you have requested a change, but I don't think I can see what the change is?! Either that, ot it means something in github I am not familiar with, I 've been using gitlab for work and some other personal stuff, so not really sure what I should do here..

Regarding the caps lock not triggering for you, I will have a look tomorrow, see if there is anything on "virtual" keys, although not sure how I would go about testing it.

Now, about the functionality, caps lock is meant to be a one-off, fire and forget toggle, that will enable/disable all animations.
Shift, is meant to be hold down while playing.

About whether the animations should stop immediately or not, I am not sure. There was existing code for this, I just made sure I call it myself when toggling the animations with caps lock.

But what you said got me thinking. In the current version, if you press shift once, you can't go back. But the first time, if you do that while a unit is moving (which I had never done before) it indeed stops the animation immediately, while this verion doesn't.

I enabled and disabled parts of this commit's code, and I got it to work as intended, for about 5 rounds, then my units completely dissapear.

Not sure what's going on, I 'll have a more in depth look tomorrow as I said, if you have any more insights, let me know.

TomWerner:
Ah, I figured it out. Holding shift prior to making a move via the arrow keys, or via pressing the go-to button does indeed temporarily disable animations. But you can't hold shift and use the "G" key to start the go-to process, nor can you hit shift once the G-based go-to process starts. And once the move starts it's too late.


> I enabled and disabled parts of this commit's code, and I got it to work as intended, for about 5 rounds, then my units completely dissapear.

If you didn't build a city I bet you ran out of money and the unit was disbanded.

This commit looks good now that I understand what was going on.

TomWerner (approved):

stavrosfa:
> If you didn't build a city I bet you ran out of money and the unit was disbanded.

Now I feel dumb as a rock! Thanks for telling me, I wouldn't have found it :D

TomWerner:
>> If you didn't build a city I bet you ran out of money and the unit was disbanded.

> Now I feel dumb as a rock! Thanks for telling me, I wouldn't have found it :D

Nah, I think there might be a toast text or something, but the UI for it is not great right now. I think I added that when I got general commerce stuff working and never went back to make it better.

stavrosfa:
I didn't find anything on the chromebook-caps lock issue you encountered. The obvious solution would be to hard code ALT+Search to trigger that, but then, it's hardcoded and only for very specific hardware, and I don't think anyone would like that.

Another way would be to somehow pseudo-type a letter like a, see if it's a or A, and have that as a way to check if caps lock is "on" or "off", but then again, what happens to non-english keyboards, etc..
So, at least for now, I think it's ok to leave it like that, and maybe wait for that Godot PR to go through.

I am not sure why, but the message to UI, although it's getting trigerred, it doesn't have any effect, especially with the Caps-Lock functionality.

I have removed it and in its place I am toggling the **endAllImmediately** field manually in every call, that seem to work better for both.

Now, it should work like expected. 
You press once the caps lock(in-game), you can toggle whether the animations are enabled-disabled.
If you hold down Shift, it should skip animations, while it's down, enable them on release.
Both Shift and CapsLock should stop animations even if a unit is **on** the move. You can't check the other way around because the animations never start.

Let me know if I missed anything, but that's pretty much all I wanted to add with this PR.

_-- Not exactly PR related --_

After fixing that, I can see that, when you have a worker mining for example, and you press caps-lock, when the worker is done, the game continues to play the mining animation. 
That is because of the way we handle animations, and how they stop. I have experimented a bit with it, but I didn't want to commit the changes here, since it's going out of scope for this one.

But in essence, we need to allow all animations to go through, and if we want them disabled, firstly, we don't await their completion, and secondly, we set the end time in the AnimationTracker to be the start time, so they take 0 ms to complete, and the end type to Stop, no matter what it was, because otherwise the next animation would start, play a frame, and freeze there. So we want a Stop ending so they can return in the default position.

This will enable other things, like, playing the animation sounds even if they are disabled.

If that sounds remotely ok, let me know, I can open a new PR and discuss there.

TomWerner (approved):

TomWerner:
> If that sounds remotely ok, let me know, I can open a new PR and discuss there.

Definitely open to that. How does the original game do it? I thought that worker animations always played even if movement animations didn't?

Feel free to merge this (and feel free in general if I've approved a change and then there was a fairly trivial change afterwards)

PR #843, opened 2025-12-10
Laying the foundation for multi-turn deals

stavrosfa:
As the title says.

I didn't implement anything concrete, as in right-of-passage, gold-per-turn, etc, we can do that later, one by one.

I have also tested saving and loading that new info in our own format, they seem to work fine.

Also, I didn't include any information on the binary format, as I was afraid it would be a very long comment, let me know if you would like one, either in the code or here.

Regardnig unit tests, it may seem that I have just included a few too many that check for the same thing over and over again, but they do have differences in their binaries and I wanted to make sure they worked as well.

### DIFF
--- a/C7Engine/C7GameData/ImportCiv3.cs
+++ b/C7Engine/C7GameData/ImportCiv3.cs
@@ -596,7 +596,7 @@ namespace C7GameData {
 			// ---------------------- Second Deal -------------------------------
 			// [2] =
 			//   Data1 = 1       <- "Gives" ROP to Player2
-			//   Data2 = 40       <- This deal will end at turn 40 (it's being active for 5 turns)
+			//   Data2 = 40       <- This deal will end at turn 40 (it's been active for 5 turns)
 			//   EntryType = -1
 			// [3] =
 			//   Data1 = 2       <- Right-of-passage (DealSubtype)
@@ -612,7 +612,7 @@ namespace C7GameData {
 			// ---------------------- Second Deal -------------------------------
 			// [2] =
 			//   Data1 = 1       <- "Gives" Spices to Player2
-			//   Data2 = 40       <- This deal will end at turn 40 (it's being active for 5 turns)
+			//   Data2 = 40       <- This deal will end at turn 40 (it's been active for 5 turns)
 			//   EntryType = -1
 			// [3] =
 			//   Data1 = 12      <- Luxury per turn (Spices)
@@ -627,7 +627,7 @@ namespace C7GameData {
 			// ---------------------- Second Deal -------------------------------
 			// [2] =
 			//   Data1 = 1       <- "Gives" 3 gpt to Player2
-			//   Data2 = 40       <- This deal will end at turn 40 (it's being active for 5 turns)
+			//   Data2 = 40       <- This deal will end at turn 40 (it's been active for 5 turns)
 			//   EntryType = -1
 			// [3] =
 			//   Data1 = 0
@@ -642,7 +642,7 @@ namespace C7GameData {
 			// ---------------------- Second Deal -------------------------------
 			// [2] =
 			//   Data1 = 0       <- "Receives" 3 gpt from Player1
-			//   Data2 = 40       <- This deal will end at turn 40 (it's being active for 5 turns)
+			//   Data2 = 40       <- This deal will end at turn 40 (it's been active for 5 turns)
 			//   EntryType = -1
 			// ---------------------- End ------------------------------
 			//
--- a/C7Engine/C7GameData/ImportCiv3.cs
+++ b/C7Engine/C7GameData/ImportCiv3.cs
@@ -454,6 +454,7 @@ namespace C7GameData {
 
 		private void ImportSavLeaders() {
 			BiqData theBiq = biq.Eras is null ? defaultBiq : biq;
+			int currentTurn = save.TurnNumber;
 			int i = 0;
 			foreach (QueryCiv3.Sav.LEAD leader in savData.Lead) {
 				if (leader.RaceID == -1) {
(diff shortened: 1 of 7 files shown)


C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -519,6 +520,124 @@ private void ImportSavLeaders() {
...
+				int turnsRemaining = 0;
+				int dealStart = 0;
+				int defaultDealDuration = save.Rules.DefaultDealDuration;
+				List<int> contacts = leader.GetContact();
+				for (int j = 0; j < contacts.Count; ++j) {
+					if (contacts[j] > 0) {
+						// skip barbarians, they can't have any diplomatic relationships, they are at war against everyone, at all times
+						if (i <= 0 || j <= 0) continue;
+
+						// skip indexes of players that are not active
```
TomWerner, line 664: Does "not active" here mean "defeated"?

C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -519,6 +520,124 @@ private void ImportSavLeaders() {
...
+						for (int d = 0; d < diplomacy.Length; ++d) {
+							DealType deadType = DealType.None;
+							DealSubType dealSubType = DealSubType.None;
+							DealDetails dealDetails = DealDetails.None;
+							int goldPerTurn = 0;
+							int resourceIndex = -1;
+							string resourcePerTurn = null;
+							ID againstPlayer = null;
+							if (diplomacy[d].EntryType == -1) {
+								turnsRemaining = diplomacy[d].Data2 - currentTurn > 0 ? diplomacy[d].Data2 - currentTurn : 0;
```
TomWerner, line 680: Could you give `Data2` a better name?

C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -519,6 +520,124 @@ private void ImportSavLeaders() {
...
+								dealStart = diplomacy[d].Data2 > 0 ? currentTurn - (defaultDealDuration - turnsRemaining) : 0;
+								// I am pretty sure that is what these values represent, take it with a grain of salt.
+								// We assign these manually anyway in each deal
+								// but still could come handy at some point to know what they are.
+								if (diplomacy[d].Data1 == 0) {
+									dealDetails = DealDetails.Inbound;
+								} else if (diplomacy[d].Data1 == 1) {
```
TomWerner, line 687: Same with Data1 - feel free to even make it private with an accessor to return an enum if that would be cleaner.

C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -519,6 +520,124 @@ private void ImportSavLeaders() {
...
+							}
+
+							if (d + 1 <= diplomacy.Length - 1) {
+								deadType = (DealType)diplomacy[d + 1].EntryType;
+								if (deadType == DealType.DiplomaticAgreement) {
+									if (diplomacy[d + 1].Data1 == 0) {
+										dealSubType = DealSubType.Peace;
+									} else if (diplomacy[d + 1].Data1 == 1) {
+										dealSubType = DealSubType.MutualProtectionPact;
+									} else if (diplomacy[d + 1].Data1 == 2) {
```
TomWerner, line 699: Same with these fields - feel free to directly make them an enum

EngineTests/GameData/MultiTurnDealsTest.cs
```diff
@@ -0,0 +1,715 @@
...
+		Exception ex = Record.Exception(() => {
+			game = ImportCiv3.ImportSav(saveFile.FullName, biqPath == "default" ? defaultBicPath : biqPath, (relativeModePath) =>
+				{ return pediaPath == "default" ? defaultPediaIconsPath : pediaPath; });
+		});
+		return (game, ex, savePath);
+	}
+
+	[Fact]
+	public async void TestMultiTurnDeal_Save_A() {
+		string is_on_github = System.Environment.GetEnvironmentVariable("CI");
```
TomWerner, line 56: Could you leave a comment explaining why this is done?

TomWerner (changes requested):
> Also, I didn't include any information on the binary format, as I was afraid it would be a very long comment, let me know if you would like one, either in the code or here.

Please do! Putting it in ImportCiv3 makes the most sense I think.

C7Engine/C7GameData/ImportCiv3.cs
stavrosfa: No, simply that if there are 8 players in the save file, don't check for the relationship between players 8 and 9 for example, it's just an empty array.

C7Engine/C7GameData/ImportCiv3.cs
stavrosfa: Answering here for all the comments regarding naming. This comes from the existing LEAD_LEAD_Diplomacy struct, I didn't name them. And at first I wanted to rename them myself, but, this is not as easy as it seems. 
The represent different kind of data depending on their values/what array they are in. So I am guessing whoever named them like this saw it as well, and gave them more of a generic name.
It might make more sense when I add the info for the binary structure.

EngineTests/GameData/MultiTurnDealsTest.cs
stavrosfa: Sure, I didn't do it because it already documented in other unit tests like SaveTests, so I didn't want to be super repetitive.

C7Engine/C7GameData/ImportCiv3.cs
stavrosfa: Added the relevant documentation, it only came out at 125 lines 🥲 
Let me know if you want to suggest something more specific after reading the documentation regarding the naming, I do still think it's fine how it is, given the nature of the data it represents.

I am still unsure about what you meant here :
_feel free to even make it private with an accessor to return an enum if that would be cleaner._

C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -522,6 +522,134 @@ private void ImportSavLeaders() {
...
+			//   Data2 = 0       <- This deal will end at turn 0 (still active, never at war)
+			//   EntryType = -1
+			// [1] =
+			//   Data1 = 0       <- Peace (DealSubtype)
+			//   Data2 = 0
+			//   EntryType = 0   <- Diplomatic Agreement (DealType)
+			// ---------------------- Second Deal -------------------------------
+			// [2] =
+			//   Data1 = 1       <- "Gives" ROP to Player2
+			//   Data2 = 40       <- This deal will end at turn 40 (it's being active for 5 turns)
```
TomWerner: "it's been active"? (same typo in a couple places below)

C7Engine/C7GameData/ImportCiv3.cs
TomWerner: Perfect, thank you!

> feel free to even make it private with an accessor to return an enum if that would be cleaner.

I guess with the fields as overloaded as it it this doesn't make sense. I was thinking if the diplomacy array value is a class, you could have an accessor that directly returned the deal type/deal subtype enums, something like that, instead of having to repeat it here. I don't think that makes as much sense as I originally thought though, so ignore me.

TomWerner (approved):

PR #844, opened 2025-12-11
Refactor: terrain improvement replacement logic

esbudylin:
Previously, the code assumed that every terraform had a terrain improvement associated with it. As a result, it caused NullReferenceException when executing terraforms for chopping forest and clearing wetlands.

This commit fixes the logic for replacement checking and separates it into a helper method in the TileOverlays class.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -809,11 +809,11 @@ public partial class Game : Node {
 			|| terraform == null || !CurrentlySelectedUnit.canPerformTerraformAction(terraform))
 			return;
 
-		TerrainImprovement currentImprovement = CurrentlySelectedUnit.location.overlays.ImprovementAtLayer(terraform);
-		if (currentImprovement != null && terraform.Improvement.upgradesFrom != currentImprovement) {
+		TerrainImprovement replacementTarget = CurrentlySelectedUnit.location.overlays.GetReplacementTarget(terraform);
+		if (replacementTarget != null) {
 			popupOverlay.ShowPopup(
 				new ConfirmationPopup(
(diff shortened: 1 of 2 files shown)


stavrosfa:
I thought I had fixed that with #841. Does it still throw a NPE? Can you give me some steps to reproduce this on the current Development branch? I tried and I couldn't trigger it.

esbudylin:
Yes, my bad, I missed your PR and tested this on an outdated version of Development branch (I tripped over this bug while playing the release build). I tested it again, and it doesn't reproduce now. I still think we should merge this PR, since it gives slight readability improvement, but it's not a big deal.

stavrosfa:
I agree it reads better, I just wanted to make sure there wasn't an issue with the previous PR.

My only comment would be to rename the method from GetReplacementTarget to something like GetTerrainImprovement, and also maybe rename ImprovementAtLayer to the same name. The method I had had the same name, but that maybe is not that accurate and GetReplacementTarget seems to be very specific. It would be easier to find and use the API in the future.

I don't know what you think, but otherwise it looks fine.

esbudylin:
> My only comment would be to rename the method from GetReplacementTarget to something like GetTerrainImprovement, and also maybe rename ImprovementAtLayer to the same name

I don't think GetTerrainImprovement is a better name for this method. The method specifically checks for an improvement that will be replaced by a terraform, not just retrieves an improvement. I also don't think that it should be an overload of ImprovementAtLayer, since these two methods do two essentially different things.

TomWerner (approved):

PR #847, opened 2026-01-05
World Size Options for New Games Added

### LINKED ISSUE #800: Allow creating games with different world sizes

zack-the-coding-actuary:
UI looks a little clunky for the world size buttons + the list of potential civs when you select "Huge", that part will need some extra help.

The linking of loading map size info from original Civ 3 files has NOT been touched. If I could get some pointers on what needs to be done with the QueryCiv3 tools I'd be happy to work on that too.

Happy to refactor the WorldSize object data into a separate file if that would be cleaner than putting it all in WorldSetup.cs, lmk.

### DIFF
--- a/C7/UIElements/NewGame/WorldSetup.cs
+++ b/C7/UIElements/NewGame/WorldSetup.cs
@@ -322,15 +322,6 @@ public partial class WorldSetup : Control {
 		billion4Large.Visible = true;
 		billion4.ButtonPressed = true;
 
-		// TODO: handle different map sizes properly (including loading the
-		// optimal city number, etc)
-		tinySize.Visible = true;
-		smallSize.Visible = true;
-		standardSize.Visible = true;
-		largeSize.Visible = true;
-		hugeSize.Visible = true;
-		randomSize.Visible = true;
-
 		tinySize.Pressed += () => {
 			sizeSelected = 0;
 		};
--- a/C7/UIElements/NewGame/WorldSetup.cs
+++ b/C7/UIElements/NewGame/WorldSetup.cs
@@ -1,6 +1,7 @@
 using Godot;
 using System;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


C7/UIElements/NewGame/WorldSetup.cs
```diff
@@ -269,12 +324,37 @@ public override void _Ready() {
 
 		// TODO: handle different map sizes properly (including loading the
 		// optimal city number, etc)
```
esbudylin: Is this TODO still relevant? Looks like you've taken care of it.

C7/UIElements/NewGame/WorldSetup.cs
```diff
@@ -269,12 +324,37 @@ public override void _Ready() {
...
-		tinySize.Visible = false;
-		smallSize.Visible = false;
-		standardSize.Visible = false;
-		largeSize.Visible = false;
-		hugeSize.Visible = false;
-		randomSize.Visible = false;
+		tinySize.Visible = true;
```
esbudylin: No sense in setting visibility of these elements in the code, since they're already marked as visible in the scene file.

esbudylin (commented):
Nice work!

> the list of potential civs when you select "Huge", that part will need some extra help

You can fix it by adding [ScrollContainer](https://docs.godotengine.org/en/stable/classes/class_scrollcontainer.html) as a parent to OpponentListContainer node in "player_setup" scene.

C7/UIElements/NewGame/WorldSetup.cs
zack-the-coding-actuary: Done!

C7/UIElements/NewGame/WorldSetup.cs
zack-the-coding-actuary: Done!

PR #848, opened 2026-01-05
Modular game mode definitions

esbudylin:
This is my take on the idea of a modular mod system that we discussed a few months ago in [Discord](https://discord.com/channels/909680526498082837/909857733388406844/1428102735625977896). The goal of this PR is to lay the foundation for a OpenCiv3 scenario format, and to allow scenario definitions to be modularized into several files.

## Game generation

This PR introduces a new class, GameModeLoader, that implements the logic for loading the new scenario format. The process for loading a scenario consists of several steps:

1) The game loads the "base" file of a scenario. This can be either a JSON file or a Lua script that returns a table. The format of the data is the same as we use for save files.

2) The game passes the data to a pipeline of Lua functions loaded from separate scripts. Each of the functions accepts the data as a Lua table, modifies and returns it. The resulting table is passed to the next function in the pipeline.

3) The final table is converted to JSON and deserialized as SaveGame.

As before, the final game generation is handled by GameSetup class, which generates map and places players on it if the scenario doesn't specify this data.

## Basic and standalone modes

This PR uses this new system to redefine the basic and the standalone modes. Previously, these two modes were defined in two separate JSON files. The difference in the "standalone" save was that it excluded the units we didn't have graphics for and updated art names for other units.

With the new system, both modes share the same JSON save as their base, while the logic for replacing the units in standalone mode is defined procedurally in Lua! And, besides being a good example of the new system, it also removes the burden of maintaining two almost identical JSON files :)

Right now, there is no way to define new modes without changing the C# source. However, if we settle on the system, we can easily implement some config or even GUI for defining new modes by selecting a base file and its addons.

## Other changes

This PR reorganizes the C7Engine classes related to Lua into a new sub-namespace: C7Engine.Lua.

There is also a new class, JsonConverter, which is used to convert data between Lua tables and JSON objects. This class was written by an LLM. Before going with it, I tried the converter built-in in the Moonsharp (it had bugs that made it unusable), and pure Lua libraries - dkjson and lunajson (they worked too slow under Moonsharp runtime).

There is also a new extension class, ScriptExtensions, for MoonSharp Script class. The goal of new extension methods is to provide logging for errors in Lua code, making debugging more straightforward.


C7/Lua/game_modes/standalone.lua
```diff
@@ -0,0 +1,54 @@
+local unit_replacement_art_map = {
```
TomWerner, line 6: Could you leave a comment explaining the purpose of this, and how the new strings correspond to folders in the art directory?

C7/Lua/game_modes/standalone.lua
```diff
@@ -0,0 +1,54 @@
...
+  ["Galleon"] = "East_Indiaman1",
+  ["Privateer"] = "Pirate Ship",
+  ["Medieval Infantry"] = "Gothic Swordsman",
+  ["Trebuchet"] = "Medieval Trebuchet",
+  ["Crusader"] = "Black Hospitaller Swordsman",
+  ["Ancient Cavalry"] = "Oscan Companion",
+  ["Curragh"] = "MinoanGalley",
+}
+
+return function(civ3_game_mode)
```
TomWerner, line 45: Since this will the example people look at when creating new versions of these Lua files, could you leave some documentation about what this function is/does/etc?

TomWerner (commented):
I skimmed through this pretty quick and it seems reasonable to me.

C7/Lua/game_modes/standalone.lua
esbudylin: Thanks! I've added comments to this script, hope they help.

stavrosfa:
I just had a thought. Let's say we have the standalone version like in your example here. What if I want to add another mod, that reuses the **standalone.lua**, but then has another addon, that does something with the artNames.

I see now that we essentially load the addons in a for loop.
What if my second addon runs first and then the **standalone.lua** runs, which will replace the art names, but my first addon has already ran, but using the wrong names.

Does this sound like a realistic scenario (not the artNames per se, but using the output of one addon in another essentially), do we expect/allow this to happen?

If so, maybe we can put something like a priority property in each script, and sort them by this before we load/execute.
We could add a base priority of 10 let's say, so there is wiggle room to lower/highten the priority of other scripts that you want to run after certain others.

esbudylin:
@stavrosfa I think this idea is out of scope for this PR. Right now there is no system to customize the load order (we just hardcode the load order for standalone mode), so parameterizing load order would be premature.

Overall, I think the viability of this idea depends on how we want the addon system to work. The initial idea by @WildWeazel that inspired me to write this PR was to provide small modules that players and modders could mix and match. So, in this case the loading order would be the responsibility of the player or modder who are composing the addons for their game. But if we decide to go with something more monolithic, baking the load order into the addons would probably make more sense.

I'm going to move forward and merge this PR over the weekend. Please let me know, if there are any objections.

stavrosfa:
Mine was more of a question "do we expect this to happen?" rather than, "we should do something immediatly about it". These days I had some trouble with some wiremocks in my day job, and most of it was due to bad priorities, that's what triggered this question.
I am in favour of merging this, I liked this approach from the start 😄

WildWeazel:
I'll try to look into this tomorrow

esbudylin:
Thanks everyone the for comments, this is now merged! @WildWeazel, if you have any comments, let me know, I'm ready to discuss.

PR #850, opened 2026-01-05
800 - Add world sizes

### LINKED ISSUE #800: Allow creating games with different world sizes

zack-the-coding-actuary:
Sorry, deleted my fork instead of my branch. I'm a noob GitHub user but I'm learning.

UI looks a little clunky for the world size buttons + the list of potential civs when you select "Huge", that part will need some extra help.

The linking of loading map size info from original Civ 3 files has NOT been touched.


stavrosfa:
Regarding UI, I don't know if you saw this comment from Yegor in your previous PR
https://github.com/C7-Game/Prototype/pull/847#pullrequestreview-3625801559

Also I can see something weird going on when hovering over the selected labels tiny, small, etc. Some letters seem to dissapear, but that might be out of scope for this one

C7/UIElements/NewGame/WorldSetup.cs
```diff
@@ -76,6 +77,60 @@ public partial class WorldSetup : Control {
...
 
+	// Maybe we refactor all this data later into a separate JSON/config file?
+	// I wasn't sure if it fit well into the MapGenerator.cs enum-style since it's a lot of integer data to store
+
+	private int sizeSelected = 2; // Correlates to standard size being default
+	private readonly Random sizeRandomizer = new Random();
+	private readonly List<WorldSize> sizeOptions = new List<WorldSize>
+	{
+		// Tiny, index 0
+		new WorldSize() {
```
stavrosfa, line 88: Where are these values for the WorldSize variants coming from?

zack-the-coding-actuary:
I can attempt to look into those items if you want to bundle that into this PR/resolving the issue. I've never used Godot before though and my skills are definitely more suited towards coding logic and features more so than UI and graphics. I know just enough right now to build and playtest.

C7/UIElements/NewGame/WorldSetup.cs
zack-the-coding-actuary: From the Civ 3 wiki + hardcoded game values: https://civilization.fandom.com/wiki/Map_(Civ3)

Albeit, they are indeed coded as "magic numbers" right now.

stavrosfa: I don't know if @TomWerner had something specific in mind when he opened the issue, besides making easier for modders to play around with the sizes, etc, and that's why he said that this should be saved in the json save file.
The other thing would be dynamic names for the buttons (Tiny, Small, etc) for all the World Sizes, vanilla & custom.
I would prefer if we wrote that to the json as well, but the current code doesn't break anything I tested so far with .biq, .sav and .json files.
If this were to be accepted, a TODO would have to go in there for a future refactor so that it can go in the json.

zack-the-coding-actuary: If you can point me to the right JSON file to save the configs in (or the directory where I should make a new config file) I'll happily refactor it. I agree that direction is better modular design.

stavrosfa: The json file is actually two files **c7-static-map-save.json** and **c7-static-map-save-standalone.json**
These files are used to launch a new game (for now, today Yegor opened a PR that merges them into one actually, but still valid), the first for the original version and the other one for the standalone with seperate graphics. I can't explain the whole thing here, but I would suggest you take some time, and try to understand how the creating/loading works.
There is a class called CreateGame.cs that you can have a look at and start to unravel the whole thing. Most of the code's variables/methods/classes are well named, so if you can't find something through code, you can search the text for what would be the most likely name, and chances are you are going to find it.
I would also suggest to use a debugger when you want to examine things in detail, and also have a look at the existing unit tests and even write your own if you want.
If you have any questions, discord is better for this kind of thing I believe

TomWerner: Yeah, we don't want to be hardcoding these.

This information should be present in the WSIZ or WCHR (world size/world characteristic) fields. We grab some data from them in ImportCiv3 right now: https://github.com/C7-Game/Prototype/blob/9ecbc5af0a8c2cfd7c14bc624a64802c975e6f97/C7Engine/C7GameData/ImportCiv3.cs#L1720-L1736

If we pull from there (and save it in the JSON files) then it's available to modders, existing civ3 mods that change the world sizes and such will work correctly, and it'll generally work nicely.

In case you're a printf debugging guy, you can use `dotnet test ../EngineTests/EngineTests.csproj -v n` to run the tests and see print statements you add, otherwise a debugger works great too. The WSIZ/WCHR classes should map to the civ3 editor pretty closely if you want to inspect the values from vanilla civ3.

TomWerner: One other printf debugging note, there's this nice [DumpObject](https://github.com/C7-Game/Prototype/blob/78d320c04f8fc10b2bfc2c38734b58f2a4790705/C7Engine/C7GameData/ImportCiv3.cs#L1791) function in `ImportCiv3.cs` that I've used with some of the civ3 structs before.

stavrosfa: Good to know!

In Rider, when running in Debug, you can right click on a variable and select Evaluate Expression, where you can run something like 

`System.Text.Json.JsonSerializer.Serialize(biq.Prto[4])`

and it will turn the object into a json, without having to modify the code, or if you have forgotten something, you can always run this as long as the var is in the current scope.

This probably doesn't cover private fields like DumpObject, but still quite helpful for larger objects

esbudylin:
Resolved by #897

Issue #424, opened 2023-09-18
Next City Name Cycles When Build City Dialog Opened but City Not Built


pcen:
Repro:
1. open build city dialog: default city name is cities[n]
2. close build city dialog instead of building city
3. open build city dialog: default city name is cities[n + 1]

The cause is that Player.GetNextCityName always increments cityNameIndex:

in C7GameData/Player.cs
```
public string GetNextCityName() {
    string name = civilization.cityNames[cityNameIndex % civilization.cityNames.Count];
    int bonusLoops = cityNameIndex / civilization.cityNames.Count;
    if (bonusLoops % 2 == 1) {
	name = "New " + name;
    }
    int suffix = (bonusLoops / 2) + 1;
    if (suffix > 1) {
	name = name + " " + suffix; //e.g. for bonusLoops = 2, we'll have "Athens 2"
    }
    cityNameIndex++;
    return name;
}

```

Instead, behaviour should increment cityNameIndex when a city is built. MsgCityBuilt is implement in #407 so once merged we can fix this behaviour easily

WildWeazel:
Technically this is faithful to Civ3 but I agree on fixing it :)

pcen:
lol good to know, I guess this should be categorised as an enhancement 😆

PR #852, opened 2026-01-06
Issue #424 - Fixed city name cycling

zack-the-coding-actuary:
City names will be reused now if player prompted a plant but cancelled. The proposed city name will now be the earliest entry in a civilization's city names that is not currently used by a city, following Civ 3 naming conventions for new loops over the cities (i.e. Athens, New Athens, Athens 2, New Athens 2, etc)

The property `int cityNameIndex` is no longer used in the `Player` class (or anywhere else from what I could tell), but I left in since there appear to be many references throughout the codebase assigning cityNameIndex when instantiating Player objects, and I didn't want to do anything destructive at this time.

### DIFF
--- a/C7Engine/C7GameData/Player.cs
+++ b/C7Engine/C7GameData/Player.cs
@@ -157,18 +157,34 @@ namespace C7GameData {
 			ResearchQueue.Enqueue(tech);
 		}
 
+		private IEnumerable<string> cityNameGenerator() {
+			List<String> cityNames = civilization.cityNames;
+			int loopCounter = 0;
+
+			// Perpetual generator expression to yield all city names lazily
+			while (true) {
+				foreach (string city in cityNames) {
+					if (loopCounter == 0) yield return city;
+					else if (loopCounter == 1) yield return $"New {city}";
+					else {
+						if (loopCounter % 2 == 0) yield return $"{city} {(loopCounter / 2) + 1}";
+						else yield return $"New {city} {(loopCounter / 2) + 1}";
+					}
+				}
+				loopCounter++;
+			}
+		}
+
 		public string GetNextCityName() {
-			string name = civilization.cityNames[cityNameIndex % civilization.cityNames.Count];
-			int bonusLoops = cityNameIndex / civilization.cityNames.Count;
-			if (bonusLoops % 2 == 1) {
-				name = "New " + name;
+			// Convert to hashset for faster lookups
+			HashSet<String> cityNameHashSet = cities.Select(city => city.name).ToHashSet();
+
+			foreach (string city in cityNameGenerator()) {
+				if (!cityNameHashSet.Contains(city)) return city;
 			}
-			int suffix = (bonusLoops / 2) + 1;
-			if (suffix > 1) {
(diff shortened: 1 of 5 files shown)


C7Engine/C7GameData/Player.cs
```diff
@@ -157,18 +157,34 @@ public void AddTechItemToResearchQueue(Tech tech) {
 			ResearchQueue.Enqueue(tech);
 		}
 
-		public string GetNextCityName() {
-			string name = civilization.cityNames[cityNameIndex % civilization.cityNames.Count];
-			int bonusLoops = cityNameIndex / civilization.cityNames.Count;
-			if (bonusLoops % 2 == 1) {
-				name = "New " + name;
+		private IEnumerable<string> cityNameGenerator() {
```
stavrosfa: It would be preferable the method name to be in Pascal case, to be more in line with c# conventions

C7Engine/C7GameData/Player.cs
```diff
@@ -157,18 +157,34 @@ public void AddTechItemToResearchQueue(Tech tech) {
...
+			List<String> cityNames = civilization.cityNames;
```
stavrosfa: Let's also use **string** instead of **String** here. There is also another instance of this below in the hashset declaration

C7Engine/C7GameData/Player.cs
```diff
@@ -157,18 +157,34 @@ public void AddTechItemToResearchQueue(Tech tech) {
...
-			int suffix = (bonusLoops / 2) + 1;
-			if (suffix > 1) {
-				name = name + " " + suffix; //e.g. for bonusLoops = 2, we'll have "Athens 2"
+		}
+
+		public string GetNextCityName() {
+			// Convert to hashset for faster lookups
+			HashSet<String> cityNameHashSet = cities.Select(city => city.name).ToHashSet();
+
+			foreach (string city in cityNameGenerator()) {
```
stavrosfa: This loop, down to where the methods returns, could be replaced by a single return statement using LINQ. Here is my take, if you want to figure it out yourself, don't click on the spoiler 😅 
<details>
  <summary>Spoiler</summary>
  
  cityNames is the cityNameHashSet in your code
  
  ```
  return cityNameGenerator().First(city => !cityNames.Contains(city));
  ```
    
</details>

I don't mind your version though, it just seems cleaner to me.

Also, if you want to keep it as is, I would throw an exception instead of returning an empty string, because it's never supposed to happen, and if it ever does we would want to know about it, and fix it, as it probably means something is wrong with the CityNameGenerator or the condition we want met

stavrosfa:
If cityNameIndex is unused I would remove it completely from the code + json save files

stavrosfa:
Also completely unrelated in terms of code, but I am sad to see this bug/feature go after 20 years of loving to hate it 😅

C7Engine/C7GameData/Player.cs
zack-the-coding-actuary: Yay I got it right!

stavrosfa (approved):

stavrosfa:
I am approving, I have also tested it locally around, seems to be ok, if anyone else has any comments, I will leave it here for a while, and then merge it

Issue #796, opened 2025-09-08
Fill in the missing details on the domestic advisor screen


TomWerner:
Here are the relevant TODOs:
 - https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7/UIElements/Advisors/DomesticAdvisor.cs#L162
 - https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7/UIElements/Advisors/DomesticAdvisor.cs#L284

We already track [maintenance costs](https://github.com/C7-Game/Prototype/blob/8b7643a1641ac43ca25ec0d6b1efc09df16c4d78/C7Engine/C7GameData/City.cs#L545), and we can calculate the city-based revenue and tax collector revenue pretty easily from methods in `City.cs`.

We don't yet have interest (need to implement the wall street wonder, which might be an interesting building to implement, you can look at some other examples like marketplaces to see an example of a building with custom logic), and don't yet have multi-turn civ-to-civ trade deals, so those can be ignored for now.

zack-the-coding-actuary:
I can grab this one!

PR #853, opened 2026-01-06
Issue #796 - Domestic advisor details

zack-the-coding-actuary:
This one was a bit of a doozy! Lots of updates going on. Created a new struct under Player.cs that can handle all the different revenue and expense sources to put directly into the domestic advisor screen, as well as refactoring the end-turn gold reconciliation just a touch so we're not doubling calculations.

Additionally, this update should be compatible with Wall Street's interest income as well as GPT deals with other players (once implemented).

I know I messed with a lot here so lmk if any changes should be made or rolled back. Thanks!


C7/UIElements/Advisors/DomesticAdvisor.cs
```diff
@@ -160,14 +159,14 @@ public void ShowAdvisor() {
 			treasury.Text = $"Treasury: {player.gold}";
 
 			// TODO: fill these in.
```
TomWerner: Looks like the TODO can be removed now?

C7Engine/C7GameData/Player.cs
```diff
@@ -8,6 +8,35 @@
 using static C7GameData.EraUtils;
 
 namespace C7GameData {
+
+	public struct PlayerCommerceBreakdown {
+		public int corrupted;
+		public int taxes;
```
TomWerner: Could you document a couple of these? In particular, taxes vs taxmenTaxes would be helpful.

C7Engine/C7GameData/Player.cs
```diff
@@ -8,6 +8,35 @@
...
+
+		public int Inflows() {
+			return corrupted + taxes + taxmenTaxes + beakers + happiness + fromOtherCivs + interest;
+		}
+
+		public int Outflows() {
+			return corrupted + beakers + happiness + toOtherCivs + maintenance + unitSupport;
+		}
+
+		public int Netflows() {
```
TomWerner, line 32: Is this Inflows - Outflow?

C7Engine/C7GameData/Player.cs
```diff
@@ -76,6 +105,8 @@ public int gold {
 			}
 		}
 
+		public bool hasWallStreet = false;
```
TomWerner: I think this is fine for now, but could you leave a TODO to track this in the Building class instead? https://github.com/C7-Game/Prototype/blob/25be85af374063773e647faf739dda33d9657f15/C7Engine/C7GameData/Building.cs#L33

That might make a good followup PR if you wanted to try doing something simple with ImportCiv3 - the relevant field in the BIQ struct is [this one](https://github.com/C7-Game/Prototype/blob/25be85af374063773e647faf739dda33d9657f15/QueryCiv3/BiqSections/Bldg.cs#L89), and there are several other examples of the building flags [here](https://github.com/C7-Game/Prototype/blob/25be85af374063773e647faf739dda33d9657f15/C7Engine/C7GameData/ImportCiv3.cs#L1274).

C7Engine/C7GameData/Player.cs
```diff
@@ -314,6 +345,62 @@ public override string ToString() {
...
+				fromOtherCivs = 0,
+				toOtherCivs = 0,
+				interest = 0,
+				maintenance = 0,
+				unitSupport = 0
+			};
+
+			int currentTurn = EngineStorage.gameData.turn;
+
+			// If current turn < 10 and player has no cities (hasn't planted yet), apply no expenses or income.
```
TomWerner: Where does this come from? Is it what the game does right now?

C7Engine/C7GameData/Player.cs
```diff
@@ -314,6 +345,62 @@ public override string ToString() {
...
+			foreach (City city in cities) {
+				CommerceBreakdown cityCommerce = city.CurrentCommerceYield();
+				result.corrupted += cityCommerce.corrupted;
+				result.taxes += cityCommerce.taxes;
+				result.beakers += cityCommerce.beakers;
+				result.happiness += cityCommerce.happiness;
+				result.maintenance += city.MaintenanceCosts();
+
+				// TODO: Hook this up to .biq or .sav specifications for tax collector income. Using default Conquests value as placeholder.
+				const int TAXMAN_VALUE = 2;
```
TomWerner: You can actually get this from the CitizenType class [here](https://github.com/C7-Game/Prototype/blob/25be85af374063773e647faf739dda33d9657f15/C7Engine/C7GameData/CitizenType.cs#L33-L35), so you shouldn't need to look at the specialist index at all.

C7Engine/C7GameData/Player.cs
```diff
@@ -314,6 +345,62 @@ public override string ToString() {
...
+					if (mtd.TurnsRemaining(currentTurn) > 0 && mtd.dealSubType == DealSubType.GoldPerTurn) {
+						if (mtd.dealDetails == DealDetails.Inbound) result.fromOtherCivs += mtd.goldPerTurn;
+						else if (mtd.dealDetails == DealDetails.Outbound) result.toOtherCivs += mtd.goldPerTurn;
+					}
+				}
+			}
+
+			result.unitSupport = TotalUnitsAllowedUnitsAndSupportCost().Item3;
+
+			if (hasWallStreet) result.interest = (int)Math.Floor(gold * 0.05);
```
TomWerner: Optional, maybe add a TODO here to load this constant from the rules. It doesn't seem to be in the BIQ file that I can see, but could be an interesting thing to be able to change.

TomWerner (changes requested):
Looks pretty good! I haven't playtested this yet but just a couple initial comments.

C7Engine/C7GameData/Player.cs
```diff
@@ -76,6 +89,8 @@ public int gold {
 			}
 		}
 
+		public bool hasWallStreet = false;
```
stavrosfa: I am not sure I like the idea of the Player knowing whether they have a Wall street building. It is too verbose, and there are certainly scenarios that this is completely irrelevant.

A method that checks for it would be more appropriate I think, or an inline check where it's used if it's just the one time, and also we don't have to store the state and having to update it. 

Check this here https://github.com/C7-Game/Prototype/blob/25be85af374063773e647faf739dda33d9657f15/C7Engine/C7GameData/Player.cs#L830 to see how the forbidden palace is retrieved when needed for an idea of what I am talking about.

Also have in mind that a public field will end up in the json save files too

C7Engine/C7GameData/City.cs
```diff
@@ -620,6 +620,10 @@ public void AddBuilding(Building building) {
 				year = 1, // TODO: Implement in-game year tracking
 				totalCulture = 0
 			});
+			
+			// Update player flag if Wall Street is built to accrue interest in gold calculations
+			// TODO: Add logic to make owner.hasWallStreet false if the city holding Wall Street is captured
+			if (building.name == "Wall Street") owner.hasWallStreet = true;
```
stavrosfa: Checking if the building name is "Wall Street" is again too verbose. If you think about it, we don't care what the building is, rather, what building (a small wonder) gives us the bonuses that Wall Street does, namely 5% on our treasury.

If in my mod tomorrow I decide that I have a new building called InterestMaker, I have to change the source code, and that's not good.

Have a look at https://github.com/C7-Game/Prototype/blob/25be85af374063773e647faf739dda33d9657f15/QueryCiv3/BiqSections/Bldg.cs#L89
and the file in general. It basically stores the traits of a building among other info. 

Then, the way to make this usable would be to add it as a flag to the building itself, that it "does" something.

So then my InterestMaker building can have a "TreasuryEarnsInterest" flag and the code stays intact.

Hope this makes sense

C7Engine/C7GameData/Player.cs
```diff
@@ -363,6 +386,23 @@ public PlayerCommerceBreakdown AggregateFlows()
...
+				{
+					if (mtd.TurnsRemaining(currentTurn) > 0 && mtd.dealSubType == DealSubType.GoldPerTurn)
+					{
+						if (mtd.dealDetails == DealDetails.Inbound) result.fromOtherCivs += mtd.goldPerTurn;
+						else if (mtd.dealDetails == DealDetails.Outbound) result.toOtherCivs += mtd.goldPerTurn;
+					}
+				}
+			}
+
+			if (hasWallStreet) result.interest = (int)Math.Floor(gold * 0.05);
```
stavrosfa: From what I read in the Civilopedia, it is 5% "with a maximum cap of 50 gold per turn", so the amount would need to be added here too.
But as I was going to say about the 0.05 which is a magic number, these values should probably go either in the Rules class here
https://github.com/C7-Game/Prototype/blob/25be85af374063773e647faf739dda33d9657f15/C7Engine/C7GameData/Rules.cs#L2
or somehow in the building itself, and of course in the save files

C7Engine/C7GameData/Player.cs
zack-the-coding-actuary: Yep! From a performance perspective, it's VERY marginally slightly better, but at the cost of readability and potential errors. I'll simplify it.

C7Engine/C7GameData/Player.cs
zack-the-coding-actuary: For the turn counter? Yes. I found that there were several calls to `EngineStorage.gameData.turn` throughout the codebase. However, it looks like it was also designed to be called directly from this point:

https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/C7Engine/EntryPoints/TurnHandling.cs#L148-L151

So I'll update to use that reference.

The mechanic of gold before planting a city? I actually just tested it in Civ 3 and all gold income/expenses are hardcoded to 0 if you have no cities (and you're not in a situation where you've been defeated yet, such as having another settler or a King unit). So I'll update the code to do that.

C7Engine/C7GameData/Player.cs
zack-the-coding-actuary: Ahh I wasn't sure if this was tracking regular commerce or the "uncorruptable" direct citizen commerce generation. Perfect that makes this easier.

zack-the-coding-actuary:
@TomWerner @stavrosfa I THINK I covered all your comments with this last commit, but let me know if I missed anything or if you spot any other problems. Thanks!

C7Engine/C7GameData/Player.cs
```diff
@@ -324,6 +353,63 @@ public override string ToString() {
...
+				foreach (CityResident cr in city.residents) {
+					// Split city income into "regular citizen" and "tax collector" buckets
+					result.taxes -= cr.citizenType.Taxes;
+					result.taxmenTaxes += cr.citizenType.Taxes;
+				}
+			}
+
+			foreach (var pr in playerRelationships.Values) {
+				foreach (var mtd in pr.multiTurnDeals) {
+					if (mtd.TurnsRemaining(currentTurn) > 0 && mtd.dealSubType == DealSubType.GoldPerTurn) {
```
stavrosfa: A minor thing here is that, multi turn deals with even 0 turns remaining are still active.
So you can either do mtd.TurnsRemaining(currentTurn) >= 0, or simply skip this and check just for the second condition (mtd.dealSubType == DealSubType.GoldPerTurn), because if there is such a deal, it means by default it's active.
TurnsRemaining was added with the intention to be used by AI and ourselves, to cancel a deal after expiration, or for example add bad reputation when a player declares war with some active deal that hasn't expired.

C7/Text/c7-static-map-save.json
```diff
@@ -76065,9 +76068,11 @@
     "startUnitType1": "Settler",
     "startUnitType2": "Worker",
     "scoutUnitType": "Scout",
-	"maxRankOfWorkableTiles": 2,
-	"maxRankOfBarbarianCampTiles": 2,
-    "defaultDealDuration": 20
+	  "maxRankOfWorkableTiles": 2,
+	  "maxRankOfBarbarianCampTiles": 2,
```
stavrosfa: I wouldn't ask for this to be changed by itself, but since there is also another change to be made, can you please fix the indentation here (and the other json) too?

stavrosfa:
@esbudylin I think you are the Lua expert here, should the flags for the building be added here too?
https://github.com/C7-Game/Prototype/blob/6419b0f069b4d5a5c945fad6ecc64e5afb6331c4/C7Engine/C7GameData/ImportCiv3.cs#L1307

and

https://github.com/C7-Game/Prototype/blob/6419b0f069b4d5a5c945fad6ecc64e5afb6331c4/C7/Lua/rules/civ3/buildings.lua

C7Engine/C7GameData/Player.cs
zack-the-coding-actuary: Ah okay so the `dealSubType` property already has a method associated with it that clears the enum assignment when `TurnsRemaining` goes below 0? Sweet, I missed that first time around but easy fix

esbudylin:
> I think you are the Lua expert here, should the flags for the building be added here too?

No, I don't think so. As far as I can see, this PR doesn't actually implement any behavior for this flag, so changes on the Lua side aren't needed.

C7Engine/C7GameData/Player.cs
stavrosfa: No you didn't miss it, it has not been implemented yet. When it is, it will simply remove the whole multi turn deal entry from the relationship. That's what I meant when I said if the deal is present, it's automatically active.

C7Engine/C7GameData/Player.cs
```diff
@@ -324,6 +353,63 @@ public override string ToString() {
...
+						if (mtd.dealDetails == DealDetails.Inbound) result.fromOtherCivs += mtd.goldPerTurn;
+						else if (mtd.dealDetails == DealDetails.Outbound) result.toOtherCivs += mtd.goldPerTurn;
+					}
+				}
+			}
+
+			result.unitSupport = TotalUnitsAllowedUnitsAndSupportCost().Item3;
+
+			if (interestBuildings > 0) result.interest = interestBuildings * Math.Min((int)(gold * rules.TreasuryInterestRate), rules.MaxInterest);
```
stavrosfa, line 406: @esbudylin That's what cought my eye, and saw some similarities with the lua files

C7Engine/C7GameData/Player.cs
zack-the-coding-actuary: We should make a note or TODO to change the container for multi-turn deals from a List to a Queue, since those will ideally be coming in chronological order and we'd be popping the first ones in. What do you think?

stavrosfa: I don't understand why we would need a Queue for this. What practical advantage does it give us, and why should we care what comes first or last, etc?

zack-the-coding-actuary: Well multi-turn deals commence and resolve in chronological order, yes? E.g. if you initiate three deals on turns 4, 6, and 8, they'll each resolve on turns 24, 26, and 28, respectively.

So when we implement the inter-turn logic, it's something like:

`while (pr.multiTurnDeals.Peek().TurnsRemaining(currentTurn) < 0) {
    pr.multiTurnDeals.Dequeue();
}`

Instead of removing elements from a List, which rebuilds the entire list per removal. It's mostly semantics since any given player likely won't have more than ~15 MTD's outstanding ever, so rebuilding the list multiple times isn't a big performance difference, but the Queue makes removing expired deals O(1) each.

stavrosfa: Let's assume that it somehow makes a huge difference in performance, and Queue is 100000 times faster.
Even then, I don't think that the Queue is not the appropriate structure for this.

> Well multi-turn deals commence and resolve in chronological order, yes?

Yes, and no. A luxury/resource per turn deal, sure, will be resolved after 20 turns (almost) automatically. But a peace treaty for example, is always there, and the duration is always 0 (unless we have been to war and are in the 20 immediate turns after peace).
Also, in the original game, there are many times (just from my playing experience, I don't have any concrete evidence) the AI won't terminate a Right of passage for example, even after 20 turns. Same with Gold per turn, sometimes the deal is dangling there, untill something else triggers the AI to terminate it.
Another thing is modularity. Maybe we decide to integrate some logic, where different deals have different durations. Or we want to negotiate with the AI for different duration times, like we do for the gold/gpt amounts

All these things would be a nightmare to deal with a queue, we would be limiting/making it super hard on ourselves to maintain.

If it were an actual performance problem, I would try something like a Swapback Array of sorts. 
If you don't know what that is, take a look at this (go a bit backwards to get some context, although I recommend watching the whole thing)

https://youtu.be/WwkuAqObplU?t=2717

Small edit, to say, I wouldn't directly go to a swapback array.
I would go
Simple Array
SwapbackArray
Managed SwapbackArray to manually shrink/grow the array size to get the best performance out of it

zack-the-coding-actuary: Ahh that's a really good point on the auto-renewing deals like peace treaty and RoP (I agree that sometimes it just continues without renewal, but I'm not sure the trigger for that).

Yeah, I agree the semantics of the Queue fall apart when you're considering deals that could renew in perpetuity, and it's also annoying for handling variable MTD lengths (PriorityQueue/min-heap could work for that but still isn't ideal for the perpetual deals).

Thanks for the insight!

C7/Text/c7-static-map-save-standalone.json
```diff
@@ -73492,9 +73495,11 @@
     "startUnitType1": "Settler",
     "startUnitType2": "Worker",
     "scoutUnitType": "Scout",
-	"maxRankOfWorkableTiles": 2,
+	  "maxRankOfWorkableTiles": 2,
 	"maxRankOfBarbarianCampTiles": 2,
-    "defaultDealDuration": 20
+  "defaultDealDuration": 20,
+    "treasuryInterestRate": 0.05,
+    "maxInterest": 50
```
TomWerner: There's some sort of weird whitespace change here, could you fix that?

C7/Text/c7-static-map-save.json
```diff
@@ -76065,9 +76068,11 @@
     "startUnitType1": "Settler",
     "startUnitType2": "Worker",
     "scoutUnitType": "Scout",
-	"maxRankOfWorkableTiles": 2,
+	  "maxRankOfWorkableTiles": 2,
```
TomWerner: same here

TomWerner (approved):
Very nice, thanks for the PR! Just a couple of notes.

I checked a couple of saves in civ3 vs openciv3 and the commerce numbers didn't line up exactly.

In one early game civ it looks like we aren't calculating the capital city's income properly, so that's unrelated to this PR (probably something related to despotism penalty/being on a river/civ trait/etc).

In a late game save I turned the sliders to zero and still saw negative numbers for happiness and science - I wonder if we're incorrectly counting the gold from specialists (entertainers/scientists) in our per-city accumulation. Might be worth checking.

Either way I think this PR is good to go, feel free to leave a TODO on the happiness/science thing and merge if you want. The corruption numbers matched up perfectly which was cool!

PR #855, opened 2026-01-10
Fix SaveTest when certain scenario filenames are present

### LINKED ISSUE #849: Unit test fails when running custom .BIQ scenario file in Scenarios
Unit test https://github.com/C7-Game/Prototype/blob/9ecbc5af0a8c2cfd7c14bc624a64802c975e6f97/EngineTests/GameData/SaveTest.cs#L272
fails when a custom scenario with a number prefix is present in the directory (ex. 1vsAll.biq)

Instead of iterating through all the scenarios with a number prefix here https://github.com/C7-Game/Prototype/blob/9ecbc5af0a8c2cfd7c14bc624a64802c975e6f97/EngineTests/GameData/SaveTest.cs#L290 for the time being, we should hardcode the scenarios that we want to test and as the comments suggest these scenarios are 

> // currently only test 1 Mesopotamia.biq -> 9 WWII in the Pacific.biq:

> // skip 7 Sengoku - Sword of the Shogun.biq for now because biq parsing fails

bmgxyz:
# Changes

- Replace the existing logic in `SaveTest.cs` for enumerating scenario files with an explicit list of scenario filenames
  - Scenarios whose filenames aren't in the list are skipped
  - Scenarios in the list whose files are missing are also skipped
  - Always skip scenario 7 because parsing is currently broken

Closes #849 

# Test Plan

1. Create an empty file in `$CIV3_HOME/Conquests/Conquests/` with a name that starts with a number, such as `1vsAll.biq`
1. Check out `Development` and run the `EngineTests`
1. Verify that `EngineTests.GameData.SaveTests.LoadAllConquestScenarios` fails to parse the empty file created above; this confirms that the test tried to load a scenario file that it shouldn't have
1. Checkout this PR's branch and re-run the tests
1. Verify that all tests now pass

### DIFF
--- a/EngineTests/GameData/SaveTest.cs
+++ b/EngineTests/GameData/SaveTest.cs
@@ -278,17 +278,7 @@ public class SaveTests {
 		string is_on_github = System.Environment.GetEnvironmentVariable("CI");
 		if (is_on_github != null) { return; }
 
-		// Only bother running one turn of the newer scenarios, just to keep the
-		// tests faster.
-		CheckScenariosInCiv3Subfolder("Conquests/Conquests", runOneTurn: true);
-		CheckScenariosInCiv3Subfolder("Conquests/Scenarios", runOneTurn: false);
-	}
-
-	private void CheckScenariosInCiv3Subfolder(string subfolder, bool runOneTurn) {
-		string conquests = Path.Join(Civ3Location.GetCiv3Path(), subfolder);
-		DirectoryInfo directoryInfo = new DirectoryInfo(conquests);
-		// skip all scenarios whose file names aren't explicitly listed here
-		string[] scenarioNamesToTest = {
+		string[] singleplayerScenarios = {
 			"1 Mesopotamia.biq",
 			"2 Rise of Rome.biq",
 			"3 Fall of Rome.biq",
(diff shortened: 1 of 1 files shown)


stavrosfa:
Thanks, It's working as expected!

However, the current test runs for both **Conquests/Conquests** and **Conquests/Scenarios** paths here 
https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/EngineTests/GameData/SaveTest.cs#L283
and the scenarios have different names, so it's silently failing for **Conquests/Scenarios** because the names don't match exactly.

Perhaps add these in the scenarioNamesToTest array and then have an Assert that checks the saveFiles count?

Regex would be another solution, but then it's not as maintainable, or even user friendly.

Another thing that could be an enchancement, that is not in the original request, but I just saw as I was going through the code now is that the json files the test produces override each other
https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/EngineTests/GameData/SaveTest.cs#L370
so it would be nice to have unique names for these, perhaps instead of "conquest_" we could use the subfolder param?

If you have the time for this as well it would be great to have this in this PR, if not that's fine as well

zack-the-coding-actuary:
Another potential thing. The folks who play live MP Civ 3 usually have their default MP Conquests scenarios name mangled by their own custom scens that pull assets from different libraries than the base ones (e.g. Tethurkan, Medieval Japan, etc)

Not sure if that's a concern at this time though, I think I might be the only one with that problem who would actually be running unit tests.

stavrosfa:
Yeah, not sure why the MP scenarios are even tested at the moment, but unless we come to an issue like we did with the custom scenario 1vsAll.biq , I think we have to replicate the existing behaviour just to be on the safer side.

I have some ideas on how we could make this more robust/consistent in future, and perhaps figure out why we test the MP files too, but for now this will do

Git history points to @TomWerner for adding the MP files, maybe you can enlighten us on this a bit more

TomWerner:
> Git history points to @TomWerner for adding the MP files, maybe you can enlighten us on this a bit more

Looks like [this commit](https://github.com/C7-Game/Prototype/commit/4031bfaed5aef0695a35936836280af7616cd95e)? I think I remember there being some issue that was present in one set of scenarios but not the other, but I don't remember for sure.

stavrosfa:
Yeah that seems to be the one, thanks! I guess that's good enough for us to leave the tests in then

bmgxyz:
> However, the current test runs for both **Conquests/Conquests** and **Conquests/Scenarios** paths here
> 
> https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/EngineTests/GameData/SaveTest.cs#L283
> 
> and the scenarios have different names, so it's silently failing for **Conquests/Scenarios** because the names don't match exactly.
> 
> Perhaps add these in the scenarioNamesToTest array and then have an Assert that checks the saveFiles count?

Good point, I moved the list of filenames up one scope and split it into singleplayer and multiplayer. These are then passed back down, and an assert in the inner function checks the number of loaded files against the length of the list.

> Another thing that could be an enchancement, that is not in the original request, but I just saw as I was going through the code now is that the json files the test produces override each other
> 
> https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/EngineTests/GameData/SaveTest.cs#L370
> 
> so it would be nice to have unique names for these, perhaps instead of "conquest_" we could use the subfolder param?

Added a basename parameter to the function that determines the first parts of each filename. The two groups are written separately now instead of overwriting each other.

esbudylin (approved):

PR #859, opened 2026-01-19
Configure a FileDataIniParser to handle duplicate keys

### LINKED ISSUE #856: INI duplicate entry issue
When a .INI file has a duplicate entry, the game doesn't crash, but it doesn't load the unit art and also doesn't behave like a unit (can't move/fight, etc)

As an example, Tank has 2 DEAD entries under [Sound Effects].

This looks like a library issue that has also stopped being updated for several years now.
https://github.com/rickyah/ini-parser

Pasting (part of) the exception here so you don't have to replicate

```
IniParser.Exceptions.ParsingException: Duplicated key 'DEAD' found in section 'Sound Effects while parsing line number 0 with value '' - IniParser version: 2.5.2.0 while parsing line number 78 with value 'DEAD=' - IniParser version: 2.5.2.0
 ---> IniParser.Exceptions.ParsingException: Duplicated key 'DEAD' found in section 'Sound Effects while parsing line number 0 with value '' - IniParser version: 2.5.2.0
   at IniParser.Parser.IniDataParser.HandleDuplicatedKeyInCollection(String key, String value, KeyDataCollection keyDataCollection, String sectionName)
   at IniParser.Parser.IniDataParser.AddKeyToKeyValueCollection(String key, String value, KeyDataCollection keyDataCollection, String sectionName)
   at IniParser.Parser.IniDataParser.ProcessKeyValuePair(String line, IniData currentIniData)
   at IniParser.Parser.IniDataParser.ProcessLine(String currentLine, IniData currentIniData)
   at IniParser.Parser.IniDataParser.Parse(String iniDataString)
   --- End of inner exception stack trace ---
   at IniParser.Parser.IniDataParser.Parse(String iniDataString)
   at IniParser.StreamIniDataParser.ReadData(StreamReader reader)
   at IniParser.FileIniDataParser.ReadFile(String filePath, Encoding fileEncoding)
   at IniParser.FileIniDataParser.ReadFile(String filePath)
   at AnimationManager.getINIData(String pathKey) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7\Animations\AnimationManager.cs:line 66
   at AnimationManager.getUnitINIData(String unitTypeName) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7\Animations\AnimationManager.cs:line 112
   at AnimationManager.getUnitFlicFilepath(UnitPrototype unit, AnimatedAction action) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7\Animations\AnimationManager.cs:line 117
   at AnimationManager.LoadAnimation(UnitPrototype unit, AnimatedAction action) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7\Animations\AnimationManager.cs:line 178
```

What could we do here?
Find another library that is actively maintained? I looked around, it doesn't look promising...
Create and maintain our own fork of this, fixing issues ad hoc (meh) ?
Write our own tiny parser (hmm) ?
Modify through code known problematic source files (ew no) ?

stavrosfa:
Configure a FileDataIniParser to allow and override duplicate keys in .ini files, for cases like the Tank.ini that has a duplicate key DEAD in its sound effects.

Closes #856 

I also commented out part of Civ3UnitSprite class that used a FileDataIniParser as the class itself seems to no longer be in use. Potentially we can completely remove it.


C7Engine/Util.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace C7Engine;
+
+public static class Util {
+	private static FileIniDataParser fileIniDataParser;
+
+	public static FileIniDataParser GetFileIniDataParser() {
+		if (fileIniDataParser != null) return fileIniDataParser;
+
+		FileIniDataParser parser = new FileIniDataParser();
+		parser.Parser.Configuration.AllowDuplicateKeys = true;
```
TomWerner, line 15: Could you leave a comment explaining why we would set these?

ConvertCiv3Media/Civ3UnitSprite.cs
```diff
@@ -67,6 +67,7 @@ public class Civ3UnitSprite {
 		public Civ3UnitSprite(Civ3UnitSprite civ3UnitSprite) {
 		}
 
+		/*
```
TomWerner: I'd just delete any of this that isn't used. We can always bring it back from  version control if needed.

TomWerner (approved):

ConvertCiv3Media/Civ3UnitSprite.cs
stavrosfa: The whole file is unused, should I just delete it as a whole?

TomWerner (approved):
Feel free to merge!

PR #861, opened 2026-01-21
Expanding the relationship API

stavrosfa:
This PR is a first pass at expanding the relationship API to make it easier to implement actual per turn deals.

I wanted to implement the Gold per turn deal, and I did, but in the process I found that I needed to alter the Relationship class to make it easier to work with. So I decided to do this first, kind of by itself.

So took this chance to also remove the boolean atWar, as I am not a huge fan of turning off and on properties to check if we are at war, so I 'd rather wrap this around a method that checks it. 
Now we do this by checking if 2 civs have a relationship, but no multi turn deals, which in turn means that these 2 civs don't have a peace deal, and since they know each other, they are at war.

I have also found it quite confusing to work with stuff like **p1.playerRelationships[p2].something** so I wrapped what I could around a more (I think) comprehensive API that hides all the annoying things from the end user (us) and once they are set, we don't have to rack our brains everytime we use it.

That being said, it's a lot to cover, but I would love your feedback at this point (like that I overused the static keyword and microsoft will have me pay some kind of subscription to keep using it this month 😛)

I am going to try and write some unit tests for this as well, as it touches many areas, but after manually testing it with multiple scenarios, I would say most of it is ok.

Let me know what you think.

### DIFF
--- a/C7Engine/C7GameData/ImportCiv3.cs
+++ b/C7Engine/C7GameData/ImportCiv3.cs
@@ -502,14 +502,15 @@ namespace C7GameData {
 			i = 0;
 			foreach (QueryCiv3.Sav.LEAD leader in savData.Lead) {
 				List<int> contacts = leader.GetContact();
-				List<bool> warStatus = leader.GetWarStatuses();
 				List<int> refuseContactForTurns = leader.GetRefuseContactForTurns();
 				for (int j = 0; j < contacts.Count; ++j) {
 					if (contacts[j] > 0) {
 						QueryCiv3.Sav.LEAD_LEAD relationship = savData.ReputationRelationship[i][j];
 						save.Players[i].playerRelationships.Add(save.Players[j].id.ToString(), new PlayerRelationship() {
-							atWar = warStatus[j],
 							warDeclarationCount = relationship.WarDeclarationCount,
+							// I don't think there is a way to figure this out for .sav or .biq files
+							// so by default we set this to 0 for these games
+							warDeclarationWithRoPActiveCount = 0,
 							wasSneakAttacked = relationship.WasSneakAttacked == 1,
 							refuseContactUntilTurn =
 								refuseContactForTurns[j] > 0 ?
--- a/C7Engine/C7GameData/ImportCiv3.cs
+++ b/C7Engine/C7GameData/ImportCiv3.cs
@@ -750,17 +750,19 @@ namespace C7GameData {
 			}
 
 			foreach (SavePlayer savePlayer in save.Players) {
-				int other = 0;
 				log.Information($"- - - - - - - - - - - - - - - - - - {savePlayer.civilization} - - - - - - - - - - - - - - - - - - ");
-				foreach (PlayerRelationship pr in savePlayer.playerRelationships.Values) {
-					other++;
-					foreach (MultiTurnDeal mtd in pr.multiTurnDeals) {
-						string against = mtd.dealSubType == DealSubType.MilitaryAlliance || mtd.dealSubType == DealSubType.TradeEmbargo ? $"(against: {save.Players.First(p => p.id == mtd.againstPlayer).civilization}({mtd.againstPlayer}))" : "";
+				foreach (KeyValuePair<string, PlayerRelationship> pr in savePlayer.playerRelationships) {
+					if (pr.Value.multiTurnDeals.Count == 0) {
+						log.Information($"{savePlayer} is at war with {save.Players.First(c => c.id.ToString() == pr.Key)}");
+						continue;
+					}
+					foreach (MultiTurnDeal mtd in pr.Value.multiTurnDeals) {
+						string against = mtd.dealSubType == DealSubType.MilitaryAlliance || mtd.dealSubType == DealSubType.TradeEmbargo ? $"(against: {save.Players.First(p => p.id == mtd.againstPlayer)})" : "";
 						string gpt = mtd.dealSubType == DealSubType.GoldPerTurn? $"({mtd.goldPerTurn} gold per turn)" : "";
 						string rpt = mtd.dealSubType == DealSubType.LuxuryPerTurn || mtd.dealSubType == DealSubType.ResourcePerTurn ? $"({mtd.resourcePerTurn} per turn)" : "";
-						log.Information($"{savePlayer.civilization}({savePlayer.id}) " +
+						log.Information($"{savePlayer} " +
 										$"has {mtd.dealSubType} " +
-										$"with {save.Players[other].civilization}({save.Players[other].id}) " +
+										$"with {save.Players.First(c => c.id.ToString() == pr.Key)} " +
 										$"for another {mtd.TurnsRemaining(save.TurnNumber)} turns {mtd.turnStartDeal}-{mtd.turnEndDeal}" +
 										$" {against} {gpt} {rpt} {mtd.dealDetails}");
 					}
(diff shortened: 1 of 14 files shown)


C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -502,14 +502,15 @@ private void ImportSavLeaders() {
...
 				List<int> refuseContactForTurns = leader.GetRefuseContactForTurns();
 				for (int j = 0; j < contacts.Count; ++j) {
 					if (contacts[j] > 0) {
 						QueryCiv3.Sav.LEAD_LEAD relationship = savData.ReputationRelationship[i][j];
 						save.Players[i].playerRelationships.Add(save.Players[j].id.ToString(), new PlayerRelationship() {
-							atWar = warStatus[j],
 							warDeclarationCount = relationship.WarDeclarationCount,
+							// I don't think there is a way to figure this out for .sav or .biq files
+							// so by default we set this to 0 for these games
+							warDeclarationWithRoPActiveCount = 0,
```
TomWerner, line 513: Does civ3 treat "declaring war with your troops in the enemy civ's borders" differently than "declaring war with rop active"?

C7Engine/C7GameData/PlayerRelationship.cs
```diff
@@ -21,6 +34,159 @@ public class PlayerRelationship {
...
+
+		if (!left.playerRelationships.TryGetValue(right.id, out var pr)) return false;
+
+		return pr.multiTurnDeals.Count == 0;
+	}
+
+	public static bool IsInAnyWar(Player player, List<Player> players) {
+		if (players
+			// any player that is not barbarians, is not us,
+			.Any(other => !other.isBarbarians && other.id != player.id
```
TomWerner: Is this equivalent to `players.Any(other => !other.isBarbarians && AtWar(player, other)`?

C7Engine/C7GameData/PlayerRelationship.cs
```diff
@@ -21,6 +34,159 @@ public class PlayerRelationship {
...
+
+		// increment the times the aggressor has declared war on the defender while there is an active RoP
+		if (HaveActiveRightOfPassage(aggressor, defender)) {
+			defender.playerRelationships[aggressor.id].warDeclarationWithRoPActiveCount++;
+		}
+
+		// update whether the defender was sneak attacked
+		defender.playerRelationships[aggressor.id].wasSneakAttacked = sneakAttack;
+
+		// TODO: do we need the same for the aggressor? Perhaps a few turns less what the defender's is?
```
TomWerner: Yeah, I think that's reasonable - you could firm up the TODO, "we need to do the same for the aggressor ..."

C7Engine/C7GameData/PlayerRelationship.cs
```diff
@@ -21,6 +34,159 @@ public class PlayerRelationship {
...
+		if (mtd.dealDetails != DealDetails.Exchange)
+			throw new Exception("This is not a valid two way deal. Perhaps you intended to use RegisterOneWayDeal() instead.");
+
+		// add the deal for this player
+		left.playerRelationships[right.id].multiTurnDeals.Add(mtd);
+
+		// add the deal for the other player
+		right.playerRelationships[left.id].multiTurnDeals.Add(mtd);
+		// right.playerRelationships[left.id].multiTurnDeals.Add(new MultiTurnDeal(mtd.dealType, mtd.dealSubType,
+		// 	mtd.dealDetails, mtd.goldPerTurn, mtd.resourcePerTurn, mtd.dealDuration, mtd.turnStartDeal, mtd.againstPlayer));
```
TomWerner: Could you clean up the commented out code?

C7Engine/C7GameData/PlayerRelationship.cs
```diff
@@ -21,6 +34,159 @@ public class PlayerRelationship {
...
+	}
+
+	/// <summary>
+	/// This automatically ends all deals except peace when they go over the initial agreed upon duration
```
TomWerner: Is it worth a TODO to add a popup if an AI/Human deal expires? I feel like that happens for some deal types in civ3

C7Engine/EntryPoints/TurnHandling.cs
```diff
@@ -17,6 +17,12 @@ internal static void OnBeginTurn() {
 			GameData gameData = EngineStorage.gameData;
 			log.Information("\n*** Beginning turn " + gameData.turn + " ***");
 
+			// TODO: We could make this more versatile.
+			// For example unless we have a good reason, as a human, receiving luxuries, gpt,
+			// or having an active RoP, doesn't hurt us.
+			log.Information("Checking to terminate any deals past their due duration");
+			CheckForObsoleteDeals(gameData.players, GetTurnNumber());
```
TomWerner: Should this happen player by player, rather than all at once? I can imagine this mattering for things like resource trades, where depending on the ordering another civ may or may not have resources available for new trades.

TomWerner (approved):

TomWerner:
Very nice! I didn't patch + playtest this, but it seems like a reasonable way to deal with this stuff. I like how this PR unifies a war/peace, and the static functions definitely make this more readable.

C7Engine/C7GameData/ImportCiv3.cs
stavrosfa: I am not sure, but I feel like it does make a difference with the other AI, besides the one you are at war with, if you break some multi turn deal like RoP, alliances, MPP and stuff like that. Just to be clear, I am not aware of such a field existing in the biq.

Maybe we don't need/want such granular control, perhaps if we keep in a dictionary or something what we have broken and against whom would be enough for the AI to calculate some kind of score?

Should I add this as a TODO?

C7Engine/C7GameData/PlayerRelationship.cs
stavrosfa: Actually yeah, thanks! You can probably guess the order in which these were implemented 😅

C7Engine/C7GameData/PlayerRelationship.cs
stavrosfa: You are right, I had totally forgotten about that, will do. 
I think this is mostly about resources/luxuries? I can't remember a popup about an expired RoP for example, but sometimes(or is it always, not sure) it will open the deal window  to renegotiate the expired deal, right? So i think this is actually 2 TODOs ?

I can see in the preferences a "Alaways renegotiate deals" option. So does this mean it has a random factor if it's off?

C7Engine/EntryPoints/TurnHandling.cs
stavrosfa: Yeah that's a really good point. And the current implementation has a bug where UnRegisterMultiTurnDeal() only works for the "left" player, because the mtd object passed is their end of the deal, and not the object the "right" player has.
So for the other player the deal lingers untill it's their turn, where they are then the "left" player. 

Even though there is a bug in my code, is this the correct behaviour? Or should we actually remove it from both when even 1 party terminates it?
I am afraid that if we do that, we will end up taking 1 round of the deal away from the second player, because it's getting yanked before it actually ends for them..

Perhaps this is what you meant too, besides not looping for every player, and check at their respective turn start instead.

I 'll drop a question in the discord channel if anyone has any information on this.

C7Engine/C7GameData/ImportCiv3.cs
TomWerner: A TODO seems reasonable, yeah. Someday we'll want to think about how to manage trade reps and such.

stavrosfa:
@TomWerner (sorry for the tags), does this warrant any further review or should it get merged?

TomWerner:
> @TomWerner (sorry for the tags), does this warrant any further review or should it get merged?

Looks good!

PR #862, opened 2026-01-27
Inflow (Wealth) implementation (Experimental/Proposal)

stavrosfa:
This PR is a product of a discussion we had in discord regarding the implementation of "Wealth". Check it out [here](https://discord.com/channels/909680526498082837/909857733388406844/1463980975477362781)

There is a lot of boiler plate code because a new type of producable items was introduced.
I ended up calling it **Inflow**, not that it's a good name, but I spent at least 2 hours searching for a proper term, in the end I just went with something.

This type/category is meant to host items like "Wealth". The wealth producable is fully implemented. I have included 2 other examples with dummy code, I hope you will get the idea. 

The point is to have stuff like wealth, that add to the city (and perhaps in the future globally), so in my examples you can see another item called "Cultivation" that is boosting the city's culture based on its current value, and another called "Expertise" that boosts science based on the city's beakers. So it's like Wealth but for Culture and Science.

These are fully exposed to the lua layer so it can do whatever it's instructed, so no need to hardcode anything in C#. We could add more restrictions like after what tech each is available and so on, this is I guess a more minimalistic example as I have simply hardcoded CanProduce() to true.

The triggers that call Lua eventually is in the Player and City classes. All 3 triggers send an integer so that Lua can use to calculate what it should do. At first I was passing the city object, but I had some infinite loop issues where I was calling from lua the method that was calling it in C#, so again, for now and for the sake of simplicity I pass an int to base the calcualtions on. This can change in the future.

If it's any easier for you to understand, I kind of tried to follow the way Terraform/TerraformImprovement is implemented and integrated with Lua.

FInal thing, this is not the "final" implementation, there are many things to be done for this to be actually merged, this is more of a POC.

Let me know what you think or if you have any questions.

@WildWeazel I hope you will like this!

Cheers!

### DIFF
--- a/C7/Lua/rules/civ3/inflows.lua
+++ b/C7/Lua/rules/civ3/inflows.lua
@@ -6,76 +6,119 @@ local inflows = {}
 
 -- to match an item in a list based on a predicate
 local function any(list, predicate)
-    for _, v in ipairs(list) do
-        if predicate(v) then
-            return true
-        end
+  for _, v in ipairs(list) do
+    if predicate(v) then
+      return true
     end
-    return false
+  end
+  return false
 end
-    
--- context is [ int input, List<Tech> techs ]
--- this is the actual implementation we would do for Wealth, for conquests
+  
+local function doubles_wealth_production(tech)
+  return tech.DoublesWealthProduction == true
+end
+
+-- context is [ Player player, City city ]
+-- this is the actual (minimal) implementation we would do for Wealth, for conquests
 local function extra_commerce_calculation(context)
-    local useful_shields = context.input
-    local known_techs = context.techs
-    local double_effect = any(known_techs,
-    function(x) 
-                return x.DoublesWealthProduction == true
-            end
-            )
-    local ratio = rules().ShieldCostPerGold
-    return math.max(1, useful_shields / (double_effect and (ratio / 2) or ratio))
+  local player = context.player
+  local city = context.city
+  
+  local useful_shields = city.CurrentProductionYield().useful
+  local known_techs = player.GetKnownTechs()
+  local double_effect = any(known_techs, doubles_wealth_production)
+  local ratio = double_effect and (rules().ShieldCostPerGold / 2) or rules().ShieldCostPerGold
+  
+  return math.max(1, useful_shields / ratio)
 end
 
--- example
-local function extra_culture_calculation(context)
-    local city_culture = context.input
-    return math.max(1, city_culture/2)
-end
+-- Any and all of the table values below should return an integer
+-- that will be added to the respective base value.
+--
+-- for example commerce will be added to the overall commerce income, culture to the current culture per turn.
+--
+-- maintenance, unitsupport and corruption are the ones that will be subtracted by their current base value
+-- so if you want MORE corruption/maintenance/unit support, the number should be negative
 
--- example
-local function extra_science_calculation(context)
-    local beakers = context.input
-    return math.max(0, beakers/10)
-end
-    
 inflows.result = {
   wealth = {
-      commerce = function(context)
-          return extra_commerce_calculation(context)
-      end,
-      culture = function(context)
-          return extra_culture_calculation(context)
-      end,
-      science = function(context)
-          return extra_science_calculation(context)
-      end,
+    commerce = function(context)
+      return extra_commerce_calculation(context)
+    end,
   },
-  -- example
-  cultivation = {
-      commerce = function(context)
-          return extra_commerce_calculation(context) * 2
-      end,
-      culture = function(context)
-          return extra_culture_calculation(context) * 2
-      end,
-      science = function(context)
-          return extra_science_calculation(context) * 2
-      end,
-  },
-  -- example
-  expertise = {
-      commerce = function(context)
-          return extra_commerce_calculation(context) * 3
-      end,
-      culture = function(context)
-          return extra_culture_calculation(context) * 3
-      end,
-      science = function(context)
-          return extra_science_calculation(context) * 3
-      end,
-  },
+  -- add new inflow(s) as the example below.
+  -- The respective json entry with the yieldCalculation having the path to this new item
+  -- should be added in the json save file, under inflows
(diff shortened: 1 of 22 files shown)


C7/Lua/rules/civ3/inflows.lua
```diff
@@ -0,0 +1,81 @@
...
+    return false
+end
+    
+-- context is [ int input, List<Tech> techs ]
+-- this is the actual implementation we would do for Wealth, for conquests
+local function extra_commerce_calculation(context)
+    local useful_shields = context.input
+    local known_techs = context.techs
+    local double_effect = any(known_techs,
+    function(x)
```
TomWerner: Could this lambda be pulled out to a named function? The formatting here otherwise makes this a little weird to read

C7/Lua/rules/civ3/inflows.lua
```diff
@@ -0,0 +1,81 @@
...
+local function extra_commerce_calculation(context)
+    local useful_shields = context.input
+    local known_techs = context.techs
+    local double_effect = any(known_techs,
+    function(x) 
+                return x.DoublesWealthProduction == true
+            end
+            )
+    local ratio = rules().ShieldCostPerGold
+    return math.max(1, useful_shields / (double_effect and (ratio / 2) or ratio))
```
TomWerner: Sorry for the lua nits, feel free to ignore this

But would doing `local ratio = double_effect and (rules().ShieldCostPerGold / 2) or rules().ShieldCostPerGold` make sense?

C7/Lua/rules/civ3/inflows.lua
```diff
@@ -0,0 +1,81 @@
...
+end
+
+-- example
```
TomWerner: For this and the one below, what would this be an example of? Could you document the idea in the code?

C7/Lua/rules/civ3/inflows.lua
```diff
@@ -0,0 +1,81 @@
...
+          return extra_commerce_calculation(context)
+      end,
+      culture = function(context)
+          return extra_culture_calculation(context)
+      end,
+      science = function(context)
+          return extra_science_calculation(context)
+      end,
+  },
+  -- example
```
TomWerner: ditto with these examples - I know the PR has some details, but if we document here too it avoids having to go through git blame

C7Engine/C7GameData/City.cs
```diff
@@ -594,11 +613,17 @@ public int GetCulture() {
 			return perPlayerCulture[owner];
 		}
 
-		public int GetCulturePerTurn() {
+		public int GetCulturePerTurn(bool bonus = true) {
```
TomWerner: what's this bonus parameter?

C7/Lua/rules/civ3/inflows.lua
```diff
@@ -0,0 +1,81 @@
...
+            end
+            )
+    local ratio = rules().ShieldCostPerGold
+    return math.max(1, useful_shields / (double_effect and (ratio / 2) or ratio))
+end
+
+-- example
+local function extra_culture_calculation(context)
+    local city_culture = context.input
+    return math.max(1, city_culture/2)
```
TomWerner: Should these default to doing nothing?

TomWerner (changes requested):

TomWerner:
Cool! Implementing wealth is nice, though I think we'd want the culture/science modifiers off by default.

C7Engine/C7GameData/City.cs
stavrosfa: Sorry, that was a leftover from a test and it was quicker to do it like this. It's been removed.

C7/Lua/rules/civ3/inflows.lua
stavrosfa: I have made this look more like it should now, and added a comment to document the usage of this. let me know if it's enough, or I should add more. 

Since it makes sense to me as I wrote it, I might be missing the bigger picture.

C7/Lua/rules/civ3/inflows.lua
stavrosfa: Yeah, it was there purely as an example instead of trying to explain it with words

stavrosfa (commented):
Thanks. I 've pushed a commit that should resemble more the final product than this.

stavrosfa:
I have added some extra functionality (corruption, happiness, etc).

I think there is an issue with how we generally calculate these values, I 've opened an issue [here](https://github.com/C7-Game/Prototype/issues/864), but the new implementation is wokring on top as it should. When this is fixed, we should be 100% there.

On the code side, I decided to change the ScriptContext to include the Player and the city, as a simple integer was too simple and limiting.

The risk to recursively call lua that calls c# is still tehre, but I have tried to add **some** guardrails for lua to not call c# methods that directly call themselves, by splitting the methods that calculate the base value and the ones that take the lua result into account, and also add the [MoonSharpHidden] annotation to those that call lua directly.
This is still not enough, because lua could call the c# method that calls the c# method that calls lua, but I am not sure how to properly handle that. 

We could add some kind of enum to pass around c# and lua to let the game know which layer is calling these methods, and when it's lua we should skip some parts, but it's clanky.
Or simply have an call caps wanring in the lua files "be careful not to call methods that call themselves" ?!

Let me know if you have any feedback on this.

TomWerner (approved):
looks good, thanks!

This definitely opens up some interesting possibilities. Since something like additional culture per turn would be pretty powerful, it could be balanced by lots of pollution or unhappiness or something. Neat!

stavrosfa:
> This definitely opens up some interesting possibilities. Since something like additional culture per turn would be pretty powerful, it could be balanced by lots of pollution or unhappiness or something. Neat!

Yeah I have a TODO regarding pollution and other things, but couldn't integrate them since I don't they are implemented yet. Feel free to suggest some more I haven't thought of!

I pushed some minor refactoring to reduce duplication, but the logic is the exact same, and added a few TODOs that I wanted to add before but forgot.

I guess this is good to merge?

PR #870, opened 2026-02-07
Add macOS build/sign/release script

fragmede:
Handles the full pipeline: build, export, sign, notarize, package, release.


TomWerner:
Thanks for the PR!

A couple quick questions:
 - How would this be used? I"m not super familiar with the Mac building process
 - The script seems very general purpose, do we actually need all those options, or would something more focused work?

PR #871, opened 2026-02-09
Adjust tile enter check to allow bombardment as well

ajhalme:
I got OpenCiv3 running and noticed that the Catapult unit was behaving strangely when attacking a city. Even though I'd been at war with an opponent for some time, I was still being prompted to declare war whenever attacking with a Catapult unit.

I suspect bombardment isn't fully enabled yet, but I tracked this specific issue down to a bug in the engine code that checks whether the moving unit can enter an occupied tile. This code checks for `attack > 0`, which in the case of a bombarding unit is not true.

This fix pushes the "tile entering aggression" logic into the Unit Prototype, and adds a relevant test to allow bombarding units to attack occupied tiles.

### DIFF
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -541,7 +541,7 @@ namespace C7GameData {
 				foreach (MapUnit other in tile.unitsOnTile) {
 					if (other.owner != owner) {
 						if (!other.owner.IsAtPeaceWith(owner))
-							return allowCombat && unitType.attack > 0;
+							return allowCombat && unitType.CanAttack();
 						return false;
 					}
 				}
--- a/C7Engine/C7GameData/UnitPrototype.cs
+++ b/C7Engine/C7GameData/UnitPrototype.cs
@@ -108,6 +108,10 @@ namespace C7GameData {
 			return categories.Contains("Sea");
 		}
 
+		public bool CanAttack() {
+			return attack > 0 || bombard > 0;
+		}
+
 		// TODO: Consider golden ages when determining whether a unit is obsolete.
 		// If a golden age has not yet been triggered and a unit can trigger one,
 		// it shouldn't be marked as obsolete, even if its upgrade is available.


stavrosfa:
Hi, thanks for the PR!

If I understand correctly, you are trying to attack with a catapult, not by bombarding, but by moving the unit on the enemy tile, as you would with a Swordsman for example.
I don't think this is how Catapults (or any other unit with 0 attack) work.

I tried to replicate it with a Trebuchet in the original game, and I got this pop up message

image

I believe only defensive bombard is impemented to a degree, when you are being attacked, but not when directly attacking (bombarding) another unit.

ajhalme:
Hmm, you are right - moving a catapult on a tile occupied by the opponent doesn't bombard in the original. Just got my Steam install working properly and was able to confirm myself. It's been a while. But I'm sure there's a game I've played where siege units, certainly range=1 units, work in this way by interpreting tile navigation as bombardment.

Without this behaviour, it does mean that the siege units are for the time being not operational at all. No way to action a bombardment, as far as I can tell. Maybe I can have a go at implementing the whole bombard mechanic.

The current behaviour of prompting the player to declare war every time when trying to do this catapult-ramming is also not great. As such, the other thing I could try is to implement this unit popup message. If that mechanism isn't in place yet, I can think about the general case at the same time.

I'll abandon this PR.

Thanks for the swift reply!

stavrosfa:
>  the siege units are for the time being not operational at all

Yeah, as I said, they are only opearational in terms of defensive bombardment. If you try to attack a tile that has a catapult on it, you will see it.

> Maybe I can have a go at implementing the whole bombard mechanic

Sure, but it might be more involved that you want at this point(code + UI + probably some lua). Perhaps a pure logic task to get familiarized with the project (or at least a part of it) is more suitable at this point.

The pop up message you mentioned might be a good one to start. Check how the "This unit has already moved" temporary pop up is implemented.

In any case, welcome on board, you can also check for open issues marked as "Good first issue" to start with, or if you have any questions or want to discuss what you would want to work on, Discord is a good place to ask!

PR #872, opened 2026-02-10
Use ReadOnlySpan<byte> comparison with WAV loading

mcraiha:
No need to create strings when doing just simple text comparison. 

Also improve exception texts in case someone actually tries to load incorrect files.


stavrosfa (approved):
Thanks, I tested this and seeems to work as expected.

The weird thing is that I tried to break it using the old code as well as this one, the exceptions never seem to be thrown.

I am approving, but someone else more familiar with this functionality should approve as well and merge this.

TomWerner:
Yeah, seems reasonable - thanks for the contribution!

PR #873, opened 2026-02-11
Fix standalone startup handling on macOS and add build instructions doc

mizuirorivi:
basis

Fixes a standalone startup issue on macOS exports where the app could skip the mode-selection flow or fail to resolve C7-packaged assets correctly. This updates main menu initialization to avoid creating classic-only buttons before Civ3 assets are available, resolves C7 asset lookups relative to the export base directory, and adds a concise macOS build guide in doc/build.md.

### DIFF
--- a/C7/UIElements/MainMenu/MainMenu.cs
+++ b/C7/UIElements/MainMenu/MainMenu.cs
@@ -43,13 +43,15 @@ public partial class MainMenu : Node {
 		LoadScenarioDialog.SetDirectoryForLoading(@"Conquests/Scenarios");
 		LoadScenarioDialog.GoToScenarioSetupAfterLoading = true;
 
-		// If couldn't find the button textures when ButtonContainer._Ready, then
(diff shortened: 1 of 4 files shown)


TomWerner:
This is great, thank you for the contribution!

If you download the mac export from the CI run does that work? Or do we need to integrate some of the things from your mac build instructions into our CI?

TomWerner (approved):

mizuirorivi:
Thanks! 
I tested this locally from a clean build and the standalone flow now works as expected on macOS (the mode-selection screen appears, and “Play in standalone mode” proceeds correctly).  

For CI, I don’t think additional workflow changes are needed for this fix. 
The macOS workflow already checks out submodules and appends Assets/Text/Lua into the zip artifact, which are the key requirements for this path fix.  
I’ll also download the mac artifact from this PR run and confirm behavior there, then report back.

PR #874, opened 2026-02-11
Added a property "Name" in MapUnit

stavrosfa:
To accomodate for custom names such as in the WWII in the Pacific scenario, and future implementation of renaming units.

The fields in the json files are actually not needed since I am checking for empty or null values for this field and assign the Unit Type name if that's the case, but I left them to be as clear as possible.


TomWerner (approved):
Neat, looks good to me! I loaded WWII and the names showed up

PR #875, opened 2026-02-12
Standardise TemporaryPopup usage, add invalid navigation popups

ajhalme:
Following up on discussion in #871, I've had a go at adding popup notifications to invalid navigation by "non-combat units" as the siege units are referred to in the original game.

I cleaned up the popup label render so that the text offset calculation is done in one location instead at every call site. I added text centering and some transparency to the label, to make the popup a bit nicer and more like the original. (I made a clone of the StyleBox for tooltips, which work better without transparency.)

Implementing the actual invalid navigation detection proved a bit more challenging than I expected. The method `CanEnterTile(..)` is used in different contexts to check for slightly different things, which made adding side effects a bit tricky. A naive implementation would make it so that the pathing and AI calculations would also raise popups. I ended up generalizing `CanEnterTile(..)` with an additional parameter to enable raising warnings when navigating from the `move(..)` context. The whole `CanEnterTile(..)` pattern probably needs a bit more thought - passing in boolean flags to change the behaviour of the function is a bit of a code smell.

Finally, while adjusting TemporaryPopup usages, I noticed that the existing MsgShowTemporaryPopup call at `MaybeAwardForestClearingShields(..)` was being called even during non-human turns, resulting in seemingly pointless  popup renders somewhere out of camera, deep in other player's territory. I've changed this so shield value popups are rendered only for human players. I want to say that this change had an observable impact on next turn load performance, but I didn't measure.

### Original game
non-combat_cannot_attack
only_combat_units_capture

### Existing popup mechanism
non-combat_cannot_attack_openciv3
only_combat_units_capture_openciv3

### Revised popup mechanism
non-combat_cannot_attack_openciv3_new
only_combat_units_capture_openciv3_new

### DIFF
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -74,6 +74,10 @@ namespace C7GameData {
 			return IsLandUnit() && unitType.defense > 0;
 		}
 
+		public bool IsNonCombatUnit() {
+			return unitType.attack == 0;
+		}
+
 		public bool HasRank() {
 			return this.unitType.attack > 0 || this.unitType.defense > 0;
 		}
@@ -500,6 +504,14 @@ namespace C7GameData {
 		// Like above, but allows specifying that we want to handle the case
 		// where a war declaration could be made before the move.
 		public bool CanEnterTile(Tile tile, bool allowCombat, bool allowWarDeclaration) {
+			return CanEnterTile(tile, new TileProbe() {
+				AllowCombat = allowCombat,
+				AllowWarDeclaration = allowWarDeclaration
+			});
+		}
+
+		// Generalized check to see whether a given tile is accessible to the unit in a given context.
+		public bool CanEnterTile(Tile tile, TileProbe probe) {
 			if (this.owner.isHuman && !this.owner.HasExploredTile(tile))
 				return true;
 
@@ -521,10 +533,25 @@ namespace C7GameData {
 			if (this.IsLandUnit() && !tile.IsLand())
 				return false;
 
+			if (IsNonCombatUnit()) {
+				if (HasHostileCity(tile, owner)) {
+					if (probe.RaiseNotice) {
+						new MsgShowTemporaryPopup($"Only combat units can capture cities and improvements.", tile).send();
+					}
+					return false;
+				}
+				if (HasHostileUnits(tile, owner)) {
+					if (probe.RaiseNotice) {
+						new MsgShowTemporaryPopup($"Non-combat units may not attack.", tile).send();
+					}
+					return false;
+				}
+			}
+
 			// If we allow declaring war on this move, then it doesn't matter if
 			// there are units belonging to another player on the tile.
 			// TODO: unbreakable alliances
-			if (allowWarDeclaration) {
+			if (probe.AllowWarDeclaration) {
 				return true;
 			}
 
@@ -541,7 +568,7 @@ namespace C7GameData {
 				foreach (MapUnit other in tile.unitsOnTile) {
 					if (other.owner != owner) {
 						if (!other.owner.IsAtPeaceWith(owner))
-							return allowCombat && unitType.attack > 0;
+							return probe.AllowCombat && unitType.attack > 0;
 						return false;
 					}
 				}
@@ -550,11 +577,23 @@ namespace C7GameData {
 			// Check for cities belonging to other civs.
 			if (tile.cityAtTile != null && tile.cityAtTile.owner != owner) {
 				if (!this.owner.isHuman || this.owner.tileKnowledge.isActiveTile(tile))
-					return allowCombat;
+					return probe.AllowCombat;
 				return false;
 			}
 
 			return true;
+		}
+
+		private static bool HasHostileUnits(Tile tile, Player player) {
+			foreach (MapUnit other in tile.unitsOnTile) {
+				if (player != other.owner && !player.IsAtPeaceWith(other.owner))
+					return true;
+			}
+			return false;
+		}
+
+		private static bool HasHostileCity(Tile tile, Player player) {
+			return tile.HasCity && !player.IsAtPeaceWith(tile.cityAtTile.owner);
 		}
 
 		/// <summary>
@@ -568,7 +607,7 @@ namespace C7GameData {
 		public async Task<bool> move(TileDirection dir, bool wait = false) {
 			(int dx, int dy) = dir.toCoordDiff();
 			Tile newLoc = EngineStorage.gameData.map.tileAt(dx + location.XCoordinate, dy + location.YCoordinate);
-			if ((newLoc != Tile.NONE) && CanEnterTile(newLoc, true) && (movementPoints.canMove)) {
+			if ((newLoc != Tile.NONE) && CanEnterTile(newLoc, TileProbe.MoveProbe()) && (movementPoints.canMove)) {
 				facingDirection = dir;
 				wake();
 
@@ -855,6 +894,16 @@ namespace C7GameData {
 			// unit.availableActions.Add("rename");
 
 			return result;
+		}
+	}
+
+	public class TileProbe {
+		public bool AllowCombat { get; set; }
+		public bool AllowWarDeclaration { get; set; }
+		public bool RaiseNotice { get; set; }
+
+		public static TileProbe MoveProbe() {
+			return new TileProbe() { AllowCombat = true, RaiseNotice = true };
 		}
 	}
 }
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -74,10 +74,6 @@ namespace C7GameData {
 			return IsLandUnit() && unitType.defense > 0;
 		}
 
-		public bool IsNonCombatUnit() {
-			return unitType.attack == 0;
-		}
-
 		public bool HasRank() {
 			return this.unitType.attack > 0 || this.unitType.defense > 0;
 		}
@@ -354,7 +350,7 @@ namespace C7GameData {
 						// TODO: Defender retreat behavior requires some more work. There's an issue for it here:
 						// https://github.com/C7-Game/Prototype/issues/274
 						Tile retreatDestination = defender.location.neighbors[attacker.facingDirection];
-						if ((retreatDestination != Tile.NONE) && defender.CanEnterTile(retreatDestination, false)) {
+						if ((retreatDestination != Tile.NONE) && defender.CanEnterTile(retreatDestination, TileProbe.PathProbe())) {
 							await defender.move(attacker.facingDirection, true);
 							result = CombatResult.DefenderRetreated;
 							break;
@@ -497,19 +493,6 @@ namespace C7GameData {
 			}
 		}
 
-		public bool CanEnterTile(Tile tile, bool allowCombat) {
-			return CanEnterTile(tile, allowCombat, allowWarDeclaration: false);
-		}
-
-		// Like above, but allows specifying that we want to handle the case
-		// where a war declaration could be made before the move.
-		public bool CanEnterTile(Tile tile, bool allowCombat, bool allowWarDeclaration) {
-			return CanEnterTile(tile, new TileProbe() {
-				AllowCombat = allowCombat,
-				AllowWarDeclaration = allowWarDeclaration
-			});
-		}
-
 		// Generalized check to see whether a given tile is accessible to the unit in a given context.
 		public bool CanEnterTile(Tile tile, TileProbe probe) {
 			if (this.owner.isHuman && !this.owner.HasExploredTile(tile))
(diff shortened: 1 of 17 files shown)


TomWerner (changes requested):
Thanks for the contribution!

I tested this out with the "quick start" game and walked my worker down 9ish tiles to the aztecs. When I tried to attack their city or the worker a war declaration popup was shown, and I accepted it. The popups you added then showed up correctly, which was cool!

However I wasn't able to get them to show up again after that. Similarly if I declared war first then tried attacking with my worker the popup didn't show. Is there something wrong with the probing logic?

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -74,6 +74,10 @@ public bool CanDefendOnLand() {
 			return IsLandUnit() && unitType.defense > 0;
 		}
 
+		public bool IsNonCombatUnit() {
```
stavrosfa: Do we really need this? Can't we do !HasRank() instead (directly below)? Even if we only want to check for attack and we do need this, the term Combat might not be the best option to use here. Something like IsAttackingUnit, or CanInitiateCombat would be more clear imo. What do you think?

Also fyi, the term HasRank is a bit silly (I can say this because I wrote it while working in a different context), I have renamed it to IsCombatUnit() in something I am currently working on, so this would be more clear in the future.

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -550,13 +577,25 @@ public bool CanEnterTile(Tile tile, bool allowCombat, bool allowWarDeclaration)
...
 				if (!this.owner.isHuman || this.owner.tileKnowledge.isActiveTile(tile))
-					return allowCombat;
+					return probe.AllowCombat;
 				return false;
 			}
 
 			return true;
 		}
 
+		private static bool HasHostileUnits(Tile tile, Player player) {
```
stavrosfa, line 572: Doesn't this and the next method make more sense as Tile (extension?) methods instead of being in MapUnit?

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -857,4 +896,14 @@ public List<UnitAction> GetAvailableActions() {
 			return result;
 		}
 	}
+
+	public class TileProbe {
```
stavrosfa: Super minor nit, I would make this a struct since we won't be mutating the values of the properties. The set also could be turned into an "init".
The logic is cool nontheless, I dislike passing booleans as arguments where I can avoid it too.

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -500,6 +504,14 @@ public bool CanEnterTile(Tile tile, bool allowCombat) {
 		// Like above, but allows specifying that we want to handle the case
 		// where a war declaration could be made before the move.
 		public bool CanEnterTile(Tile tile, bool allowCombat, bool allowWarDeclaration) {
```
stavrosfa: We could get rid of this completely, I don't think it adds any extra value, we could start passing the TileProbe argument everywhere and get rid of the booleans.

stavrosfa (commented):
Thanks for this!

A few things we can consider for the future that are beyond the scope of this PR.

1. When I click multiple times I get multiple overlapping pop ups, without really being able to see them, except for the top one. Maybe we can only keep one pop up  per tile at any given time.
2. We should not be able to declare war with workers or artillery units, or any non-attacking unit.
3. When we raise these pop ups, we should be focusing the camera on that tile and staying there until the pop up goes away, otherwise we are missing on what's happening if it's not directly in our visible area. This is more of an animation issue and how we handle the unit's actions, etc, but perhaps this piece can be implemented by itself at some point

C7Engine/C7GameData/MapUnit.cs
ajhalme: Removed, replaced usage with `!HasRank()`.

I did have a quick look at what functions were available. I was looking for `IsCombatUnit()`, but I didn't see it and then wasn't sure if that's a real category within the game. I think there's some opportunity to tighten up the code with a few of these property helper functions.

Feels very reasonable to say that any unit that can either attack or defend is a *combat unit*, so I'm in favour of renaming `HasRank()` to `IsCombatUnit()`.

C7Engine/C7GameData/MapUnit.cs
ajhalme: Yes, I was thinking about pushing this inside `Tile` because the method relies on `Tile` types internal state. But then `Tile` would have to know about `Player` types ability to be at peace (or not) with other players. In a way this logic kind of belongs in both...and neither.

An extension method would work here, but I didn't see that pattern used that much.

But really, I feel `MapUnit` is exactly the type where `Tile` and `Player` meet, so having the logic right here is quite nice. With a private method, this logic composition doesn't need to go further than this file.

I spotted a few cases where a few of these local helpers might make some of the nested logic easier to read. Building out some public extension methods could work as well.

C7Engine/C7GameData/MapUnit.cs
ajhalme: Changed to a struct. Editor even suggested `private init` after I cleaned up the usage, but this is good enough for now.

C7Engine/C7GameData/MapUnit.cs
ajhalme: I considered pushing the method signature change everywhere, but I didn't want to touch all the call sites.

Given you liked the pattern, I did push this change through. There weren't that many callers after all. And as this can be a hot path, I figured might as well streamline it all. Plus, the expanded TileProbe pattern is quite tidy and helped me find the bug.

ajhalme:
Thanks both for taking a look.

> I wasn't able to get them to show up again after that. Similarly if I declared war first then tried attacking with my worker the popup didn't show. Is there something wrong with the probing logic?

I think I got this issue fixed now. This was due to the various calling contexts of `CanEnterTile` and the behaviour differences during the `GoTo` planning and `Move` execution. With the contexts now explicitly laid out in `TileProbe`, I could see that I should be returning `true` even for non-combat units, when calling CanEnterTile from contexts that *aren't* `Move`.

The context-dependent logic of `CanEnterTile` is rather mind-bending and could maybe benefit from a rethink, but maybe the `TileProbe` variants make it a bit more manageable and extensible now.

> When I click multiple times I get multiple overlapping pop ups, without really being able to see them, except for the top one. Maybe we can only keep one pop up per tile at any given time.

The original game has an elegant solution for this: the popups are in a stack that builds upwards as new popups get added in. A popup stack in either the tile or unit object might work nicely. Could be a nice first task for someone.

> We should not be able to declare war with workers or artillery units, or any non-attacking unit.

Yeah, should be fairly straightforward. Maybe as part of a broader cleanup pass of `MapUnit`.

> When we raise these pop ups, we should be focusing the camera on that tile and staying there until the pop up goes away, otherwise we are missing on what's happening if it's not directly in our visible area. This is more of an animation issue and how we handle the unit's actions, etc, but perhaps this piece can be implemented by itself at some point.

Hmm, maybe. Could also be quite jarring if you have a bunch of these in a row. I have noticed a few times when fortifying a unit that the move animations for other units run first and the unit in focus / middle of the screen is in a kind of a limbo waiting for the other units.

I wonder if there are many "positive" popups like shields from cutting down a forest, or is it mostly just from attempted illegal actions? If it's mostly "negatives", those probably occur when the user is instructing a unit in the visible section of the map, so any centering might be disorienting.

I do recall seeing this kind of centering happening quite regularly in the original games.

C7Engine/C7GameData/MapUnit.cs
stavrosfa: > but I didn't see that pattern used that much.

Don't let that stop you, if you think something is better/cleaner this way, go ahead.

What you have is not bad in any way, I am just thinking that the `Tile` is the object that has the property `HasHostileUnits` or `HasHostileCity` and it makes sense in my head and if I was trying to see if this API exists so I can use it in my code, I would seach for it in `Tile`.
If it's an extension you can also read this like, `tile.HasHostileUnits(playerTryingToMove)` over the simpler `HasHostileUnits(tile, playerTryingToMove)`


Also, I have something in what I am working on similar to `HasHostileUnits`, and I see now that I have it  in `Tile`, so what I am trying to say is, think in terms of what could be useful for other use cases rather than this specific thing you are working on.

You can leave this as is for now, I am not trying to get you to rewrite anything 😃

stavrosfa:
I was thinking about the positive ones to be honest, rather than the negative, which as you say are mostly triggered when the bad action is happening in the visible area of the map.
There are stuff like, "We Love the King Day", or "The Statue of Zeus produced an Ancient Cavalry" etc, plus all other future pop ups we can implement that are not in the original game.

The whole centering of the camera could also be a preference setting, so if I want to see them I can turn it on, and if you don't want to see it you could turn it off.

TomWerner (approved):
Looks reasonable to me!

I confirmed the behavior I noticed earlier was fixed, and let the game run in observer mode and confirmed war was still declared. (Observer mode definitely seems to have gotten slower, we'll have to do some benchmarking at some point)

Feel free to merge once stavrosfa posts any responses!

stavrosfa:
> Observer mode definitely seems to have gotten slower

How come? Because of the extra checks in CanEnterTile()? I don't see anything else that would potentially cause any slowness, but even that seems like an unlikely culprit.

Can you give me some steps to reproduce?

ajhalme:
@TomWerner - Do you suspect an issue in this PR, or do you mean that running the observer mode has gotten slower *for a while now*?

--

I didn't notice the observer mode in the codebase - handy feature! Would be neat if one could also set the Quickstart AI seed. I haven't looked at the AI code much, but I seemed to be getting different games from the same setup.

(For anyone wondering, observer mode is `Ctrl+Alt+Shift+O` at turn end; it prompts for how many turns to let the game run on autopilot.)

A quick online search found [this excellent tutorial on Godot profiling](https://www.youtube.com/watch?v=eafdNo1kMA8), so I ran the profiler for both the Development branch I have (commit `e0c7cb340443fe8c97a382fecdc34d1beeee41fd`) and this `unit-popup` branch and after the default 100 turns got the results below on my trusty 10-year-old Dell XPS 9350 (Godot 4.6; Linux Mint 22.2 (~Ubuntu 24.04 & patches), Linux 6.8.0-100-generic). Nothing conclusive, it's a bit apples to oranges, but I don't think the changes I'm introducing here are making the observer mode meaningfully worse.

The results suggest that, as expected, the processing time seems to increase with the number of turns and the amount of complexity generated by the players. The visuals are posing no trouble to the engine. I was seeing 25-30x peak time spent on "process time" vs "physics frame time". I wouldn't read much into the frequency of the profiler spikes - those seemed to fluxuate a lot for both runs.

The tutorial shows how to declare custom monitors. That could be something to explore. One could also try running the engine alone in some kind of a headless mode as an interesting performance test. With good instrumentation, that could yield insights on where resources are being spent.

monitors_dev
monitors_unit-popup
profile_dev
profile_unit-popup
visual_dev
visual_unit-popup

TomWerner:
> @TomWerner - Do you suspect an issue in this PR, or do you mean that running the observer mode has gotten slower for a while now?

Sorry, I should have been clearer - I think observer mode is slower now than it was back in say March, not anything to do with your PR. Thanks for the profiling links, that'll come in handy. We definitely can run the engine in headless mode, [this test](https://github.com/C7-Game/Prototype/blob/e0c7cb340443fe8c97a382fecdc34d1beeee41fd/EngineTests/GameData/SaveTest.cs#L157) plays out the first 50 turns of a game. I bet we're just doing some overly expensive AI stuff.

TomWerner:
(It's also possibly my poor little Chromebook was having a slow day and I was just imagining things)

stavrosfa:
Yeah, i think the pathfinding in general got more expensive (and I am partly responsible). There is definitely room for improvement there.

Nice work @ajhalme, thanks for taking the time to profile this and share the insight, appreciate it.

Alright, is this one good to merge then? I don't know if @ajhalme can merge this themselves, but I could do it, it's just a button press really.

ajhalme:
Good to go as far as I'm concerned. I don't see a button to merge.

stavrosfa:
I think I just stumbled upon what might be slowing the observer mode down.

When I am in observer mode, and my units (the human units) are in the visible aera of my screen, this error is flooding the terminal continuously (which I think is writing in the Godot sink, same as it would if we wrote `GD.Print("something")`, hence the slowdown)

ERROR: Condition "p_texture.is_null()" is true.
   at: draw_texture_rect_region (scene/main/canvas_item.cpp:855)

When I am not in observer mode I don't get this error.

Again, this is unrelated to this PR, I was testing in the Development branch

ajhalme:
If you are on Godot 4.4, you might be missing out on better error backtracing, introduced in Godot 4.5.

On Godot 4.6 I'm getting this kind of stuff in my observer mode. Something off with HillsLayer draw?

image

stavrosfa:
You are actually right! We don't load the `jungleVolcanoTexture` on the hills layer which was definetly causing the error I was actually talking about.
Thanks a ton!

I am running the game through Rider, so I don't have (almost) never Godot open.

PR #881, opened 2026-02-14
Nationality for MapUnits and adjective for Civilization

stavrosfa:
Added property Nationality for MapUnit, distinguishing between native and captured units.
Added adjective for Civilization
Updated some description texts

Another bite size chunk that can be merged by itself, before I open a PR for the capturing units mechanism.

### DIFF
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -14,6 +14,7 @@ namespace C7GameData {
 	public class MapUnit {
 		public ID id { get; internal set; }
 		public string name { get; internal set; }
+		public Civilization nationality { get; set; }
 		public UnitPrototype unitType { get; set; }
 		public Player owner { get; set; }
 		public Tile previousLocation { get; internal set; }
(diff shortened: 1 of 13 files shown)


C7Engine/C7GameData/MapUnit.cs
```diff
@@ -83,19 +83,31 @@ public bool CanBeActive() {
 			return !this.IsBusy() && this.movementPoints.canMove;
 		}
 
+		public bool IsSlave() {
```
TomWerner: Maybe `IsCaptured`? That seems like it generalizes better to artillery and such (and I think is how it's displayed in the military advisor)

TomWerner (approved):
Cool! The captured workers being slower is a nice functionality gap to close

C7Engine/C7GameData/MapUnit.cs
stavrosfa: How about to `IsCaptive` ? `IsCaptured` seems like the logical term, but I associate it with the result of a method, like another unit tried to capture it, and it succeeded, not the state of the unit, if that makes any sense.

TomWerner: `IsCaptive` sounds great, thanks!

PR #882, opened 2026-02-15
Load jungleVolcanoTexture

stavrosfa:
We weren't loading this texture at all, causing some error flooding which in turn potentially also slowed the game down.

Thanks to @ajhalme for picking up on it


TomWerner:
Nice, thanks!

PR #883, opened 2026-02-16
On city destroy, remove ownership from "fringe" tiles outside natural city border

### LINKED ISSUE #854: Border Recalculation when a city is destroyed/captured
When a city (city A) is captured (and currently destroyed by default), a neighboring city (city B) that has a connecting tile, meaning a tile that is there because of Law 7 or 8 https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/C7Engine/C7GameData/GameData.cs#L151
remains in the city that can't actually justify owning it now that city A is gone.

We do calcuate the tiles after capture here
https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/C7Engine/C7GameData/GameData.cs#L183

but we need to recalculate the affected tiles, and remove ownership of tiles that we can't justify.

To be cleared, the tile is not justified because 1. the owning city hasn't expanded, 2. its existance is solely because of border laws.

Also, when removing tile ownership from CityA, here
https://github.com/C7-Game/Prototype/blob/1cc393311530907bf687398cf7c7b0006d5192ee/C7Engine/C7GameData/GameData.cs#L179
I am not sure that's enough and we don't end up with leftover tile ownership that we can't just see, I haven't tested for that, it's merely a thought.

But in any case, we should go through with the fix for the tiles with the destroyed city too, because the dangling tile could have been in the destroyed city's ownership just as well that GetTilesWithinBorders() wouldn't be able to retrieve.

Here is an screenshot to better visualize the scenario

Image

After another civ captured my Old City (expanded once), the tile that belongs to Utrecht marked with X should have been removed but remains in Utrecht's ownership.

ajhalme:
Fix for #854.

When a city is destroyed (the default capture mechanism for the time being) a special routine clears out the `owningCity` property from tiles controlled by the city. The current implementation is limited to removing city ownership from tiles within the "natural" culture/rank based city boundary.

If my reading of the laws is correct, it should be enough to simply remove *all* ownership from the tiles around the "natural" city border, as 1) the laws relate to "edge neighbours" of "naturally" owned tiles, and 2) the update mechanism will restore any valid ownership that got removed.

The point is that we *want* to recompute "from scratch" the tile ownership around the destroyed city. Instead of trying to be clever about removing ownership precisely and dealing with corner cases, just take out all tile ownership that *could* be affected, and let the update heal the state. If there's a corner case missed by this scheme, the fix should be done in the update mechanism.

Relevant lore on CFC: *[The Eight Laws of Border Dynamics](https://forums.civfanatics.com/threads/the-eight-laws-of-border-dynamics.106882/)*

-- 

Test save file:
[854_border_recompute.json](https://github.com/user-attachments/files/25329878/854_border_recompute.json)

Screens:
borderline_1_before
borderline_2_current
borderline_3_new

### DIFF
--- a/C7Engine/C7GameData/GameData.cs
+++ b/C7Engine/C7GameData/GameData.cs
@@ -189,7 +189,7 @@ namespace C7GameData {
 				foreach (var fringeTile in tile.GetEdgeNeighbors().Where(x => !borderTileIds.Contains(x.Id))) {
 					if (fringeTile.HasCity) // skip encountered cities, just in case 
 						continue;
-					
+
 					fringeTile.owningCity = null;
 				}
 			}
--- a/C7Engine/C7GameData/GameData.cs
+++ b/C7Engine/C7GameData/GameData.cs
@@ -176,9 +176,19 @@ namespace C7GameData {
 
 		public void UpdateTileOwnersOnCityDestruction(City city) {
 			city.location.owningCity = null;
-
-			foreach (Tile tile in city.GetTilesWithinBorders()) {
+			
+			var borderTiles = city.GetTilesWithinBorders();
+			var borderTileIds = borderTiles.Select(x => x.Id).ToHashSet();
+			
+			foreach (Tile tile in borderTiles) {
 				tile.owningCity = null;
+
+				// Aggressively remove ownership from edge neighbors around the city outside the natural border.
+				// This clears out tile ownership due to Law VII or Law VIII.
+				// Regular tile ownership update will re-assign ownership where needed.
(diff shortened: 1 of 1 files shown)


TomWerner (approved):
Thanks!

stavrosfa (approved):
Thanks, looks good to me!

If the workflows are failing because of the formatting like in this case, you need to run `dotnet format C7.sln whitespace` and commit again. 
I am trying for it to be the last thing I do before I push, this and run the unit tests, just keep it in mind.

PR #884, opened 2026-02-16
player relationships: Remove the message to the engine

### LINKED ISSUE #877: MessageToEngine weird behaviour
Very recently I started noticing that workers where completing mines and road in pretty much 0 turns. I traced this behaviour back to this https://github.com/C7-Game/Prototype/pull/861
Before this merge, the worker behaviour is fine. This is weird because I didn't change anything even remotely close to how workers behave.

After hours of looking through this, I have narrowed it down to this line here
https://github.com/C7-Game/Prototype/blob/e0c7cb340443fe8c97a382fecdc34d1beeee41fd/C7Engine/AI/PlayerAI.cs#L31

if I comment this line out, the workers behave as usual.

I tried to comment out all the behaviour inside this message (which had nothing to do with workers), still couldn't fix it.

Then I tried to send the message twice. From the logs I can see that now stuff were called twice as much. Then I completely replaced the message I send, with this, which does nothing

```
    public class SomeMessage : MessageToEngine
    {
        public override void process()
        {
        }
    }
```

and I got the same bad behaviour. 

Which leads me to believe that there is something wrong with the Message Engine itself, and the number of messages it processes, but alas, I can't figure it out.

Any help is welcome.

Steps to reproduce:
Launch a new game, build your city, press A for the worker, and see that they complete a mine or a road or cutting trees, in the same turn.

TomWerner:
Both places where we created the message already could access the gamedata directly, and this seems to be one of the two things that was slowing down observer mode on my chromebook (the other being animations not being disabled, somehow).

I suspect there was some weirdness with repeatedly trying to reacquire the mutex we were already holding or something.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -320,7 +320,10 @@ public partial class Game : Node {
 		EngineStorage.ReadGameData((GameData gameData) => {
 			log.Information("Starting player turn");
 
-			new MsgCheckObsoleteDeals(controller).send();
+			// TODO: Before we call this method to automatically end obsolete deals, we could make this more versatile.
+			// For example unless we have a good reason, as a human, receiving luxuries, gpt,
+			// or having an active RoP, doesn't hurt us.
+			PlayerRelationship.CheckForObsoleteDeals(controller, gameData.players, gameData.turn);
 
 			// If the player can now pick a new government, force them to do so.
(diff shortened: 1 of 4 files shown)


stavrosfa (approved):
Cool, I was trying to be smart and have the "implementation" in one place, and it bit me 😓 

Thanks for having a look, feel free to merge!

PR #885, opened 2026-02-16
Ensure that animations never run in observer mode

TomWerner:
This avoids any issues with the "shift" key toggling animations colliding with the "ctrl+alt+shift+o" shortcut for observer mode.

### DIFF
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -166,7 +166,7 @@ namespace C7GameData {
 		}
 
 		public async Task animateAsync(MapUnit.AnimatedAction action, AnimationEnding ending = AnimationEnding.Stop) {
-			if (EngineStorage.animationsEnabled) {
+			if (EngineStorage.animationsEnabled && !EngineStorage.gameData.observerMode) {
 				var msg = new MsgStartUnitAnimation(this, action, ending);
 				msg.send();
 


stavrosfa (commented):
Do we want to bake the observer mode not having animations in the code? I think this is more of an Animation issue and that all the units are moving at the same time.

That being said, I don't mind this, but I'd rather put it in the `animateAsync` method in MapUnit and Tile with something like

`if (EngineStorage.animationsEnabled && !EngineStorage.gameData.observerMode)`

~or in one single place in the `AnimationTracker` `startAnimation` method
`if (EngineStorage.gameData.observerMode) return;`~
this doesn't actually work, I suspect it doesn't pass the if `(EngineStorage.animationsEnabled)` check in MapUnit

If that works for you, I think it would be preferable, let me know what you think.

Perhaps also a TODO to have a future look at it if we "fix" the animations would be useful.
Now that I tthink of, another TODO for an configuration/preferences option to choose if animations are enabled/disabled in obvserver mode would come in handy someday

esbudylin:
Have you tried other ways of dealing with this input collision? 

You can try to add an action for toggling observer mode to the Godot InputMap instead of manually handling the key combination for it. AFAIK, the engine should be able to resolve the collision and won't trigger two actions for the same key stroke. https://docs.godotengine.org/en/stable/tutorials/inputs/input_examples.html#inputmap

stavrosfa:
I had the exact opposite experience with Godot, I have left a comment [here ](https://github.com/C7-Game/Prototype/blob/546a91ad91bee4d944de775c331fc67ed4d112cf/C7/Game.cs#L629)as well regarding this.

[This](https://github.com/C7-Game/Prototype/blob/546a91ad91bee4d944de775c331fc67ed4d112cf/C7/Game.cs#L629) also seems to suggest the same way, check for pressing `T` for example, and if `Shift` is pressed do X otherwise do Y. 

Am I getting this wrong?

I don't think the best way to handle inputs is manually, but it seems to be the only way.

esbudylin:
@stavrosfa Hm, I've the reread documentation for Input.IsActionJustPressed, and it seems that by default it doesn't take modifier keys (Shift, Alt, Control) into account. It allows to pass a flag to toggle this behavior https://docs.godotengine.org/en/4.4/classes/class_input.html#class-input-method-is-action-just-pressed 

Maybe calling this method with `exactMatch` flag set to `true` can fix this issue?

stavrosfa:
I tried to do

```
if (Input.IsActionJustPressed(action, true)) {
ProcessAction(action.ToString());
} else if (Input.IsActionJustReleased(action, true)) {
ProcessOnReleaseAction(action.ToString());
}
```

where the `true` boolean I am passing is the `exactMatch `argument

and it doesn't seem to make a difference with the current setup.

If I release Shift last from the keyboard, it enables the animations. If I release another key last, the animations stay disabled.

Not exactly a scientific way to try this, but hey 😅 

EDIT: Do I need to pass the `CTRL + SHIFT + ALT + O` as an action though? I think it won't matter because the release action is still the Shift key, so it will do the same thing..

TomWerner:
> That being said, I don't mind this, but I'd rather put it in the animateAsync method in MapUnit and Tile with something like

> if (EngineStorage.animationsEnabled && !EngineStorage.gameData.observerMode)

Yeah, that seems reasonable. Done

stavrosfa (approved):
There is also this here

https://github.com/C7-Game/Prototype/blob/546a91ad91bee4d944de775c331fc67ed4d112cf/C7Engine/C7GameData/Tile.cs#L754

I think we would want that too, if anything, for consistency

Feel free to merge this btw

PR #886, opened 2026-02-16
Nationality for MapUnits and adjective for Civilization

stavrosfa:
The same as PR #881, it seemed easier to create a new branch and push here, rather than rebase after the latest extensive changes


TomWerner (approved):
I didn't playtest this but it looks just like the old one

PR #887, opened 2026-02-16
Civ primary and secondary colors

stavrosfa:
Implemented primary and secondary colors, with some functionality to try and avoid duplicates.

Also moved the shader caching, as well as the new color calculations to a new Util class as they are specific to the gameplay and not the pure loading of files/colors

### DIFF
--- a/C7/Textures/PlayerTextureUtil.cs
+++ b/C7/Textures/PlayerTextureUtil.cs
@@ -1,0 +1,65 @@
+using System.Collections.Generic;
+using C7Engine;
+using C7GameData;
+using System.Linq;
+using Godot;
+
+namespace C7.Textures;
+
+public static class PlayerTextureUtil {
+	private static Dictionary<int, ShaderMaterial> materialCache = [];
+	private static Dictionary<ID, int> colorsInUseCache = [];
+
+	public static int GetPlayerColor(this Player player) {
+		if (!colorsInUseCache.ContainsKey(player.id)) {
+			InitializeCivColors(EngineStorage.gameData);
+		}
+		return colorsInUseCache[player.id];
+	}
+
+	private static int LoadCivColor(Player player) {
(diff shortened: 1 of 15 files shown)


C7/Textures/PlayerTextureUtil.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Linq;
+using Godot;
+
+namespace C7.Textures;
+
+public static class PlayerTextureUtil {
+	private static Dictionary<int, ShaderMaterial> materialCache = [];
+	private static Dictionary<ID, int> colorsInUseCache = [];
+
+	public static int GetPlayerColor(this Player player) {
```
TomWerner, line 20: Could you document what the returned int is? Is it the key in the material cache?

C7/Textures/PlayerTextureUtil.cs
```diff
@@ -0,0 +1,65 @@
...
+			InitializeCivColors(EngineStorage.gameData);
+		}
+		return colorsInUseCache[player.id];
+	}
+
+	private static int LoadCivColor(Player player) {
+		if (colorsInUseCache.TryGetValue(player.id, out int colorIndex))
+			return colorIndex;
+
+		if (colorsInUseCache.ContainsValue(player.primaryColorIndex)) {
```
TomWerner, line 47: Would this be an accurate comment?

```
// If the player's primary color is in use, fall back to their secondary color.
// This may already be in use, but that's ok; we don't have a third
// color for each civ to fall back to.
```

TomWerner: And maybe document how this interacts with the second run in InitializeCivColors?

TomWerner (approved):

stavrosfa:
I added some documentation, let me know if it makes sense

TomWerner:
Perfect, thanks!

PR #889, opened 2026-02-16
TileKnowledge: Optimize GetTilesVisibleToUnit

TomWerner:
This speeds up the time for the first 100 turns in observer mode from ~40 seconds to ~35 on my chromebook, and brings the printf benchmark from ~230 ms to ~40.

Optimizations:
 - avoid `TileDirection.GetValues(typeof(TileDirection))` + a dictionary lookup and just iterate over the dictionary directly
 - use a list instead of a hashmap, and size it to hold the maximum number of tiles
 - (probably not super important) pulled out some variables so they aren't accessed repeatedly in the inner loop

### DIFF
--- a/C7Engine/C7GameData/AIData/TileKnowledge.cs
+++ b/C7Engine/C7GameData/AIData/TileKnowledge.cs
@@ -82,7 +82,8 @@ namespace C7GameData {
 					// tiles away. We can't see the hill that is SW+SW of us
 					// because we are blocked. To implement this we need to
 					// prevent 90 degree "turns", which we handle by ensuring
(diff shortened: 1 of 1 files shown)


C7Engine/C7GameData/AIData/TileKnowledge.cs
stavrosfa: Can the `a+b` in the documentation be updated to reflect the renaming of the variables?

stavrosfa (approved):
I run this and it seems to almost cut the calculation times in half from ~24ms to ~12ms on my machine, good stuff!

PR #890, opened 2026-02-16
Put a 5 minute timeout on CI test runs and improve some debugging aspects

TomWerner:
Getting the hang dumps and info logging would have been very helpful with the recent stuck tests.

### DIFF
--- a/EngineTests/GameData/SaveTest.cs
+++ b/EngineTests/GameData/SaveTest.cs
@@ -12,6 +12,7 @@ using C7GameData.Save;
 using Newtonsoft.Json.Linq;
 using QueryCiv3;
 using Xunit;
+using Serilog;
 
 namespace EngineTests.GameData;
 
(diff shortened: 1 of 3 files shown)


EngineTests/GameData/SaveTest.cs
stavrosfa, line 169: Why do we need this in the unit tests in the first place? It looks to me like the behaviour of the game is somehow tied to the UI?
And why don't check for `MsgCityCreated` or other UI messages for example?

stavrosfa (approved):
The PR looks good to me, its purpose anyway, I am no expert in this area.

I only had one question that is kind of irrelevant to this PR, I am just seeing this for the first time.

Feel free to merge this as far as i am concerned Tom

EngineTests/GameData/SaveTest.cs
TomWerner: It is slightly tied to the UI - it pauses between turns to let the player hit "next turn". There's also a UI link when the AI opens a diplomacy popup, but that's only for human players. I wanted to make sure that wasn't involved, hence throwing an exception if we see one.

TomWerner:
Thanks!

PR #891, opened 2026-02-18
Unit disband utilizing the Lua layer

stavrosfa:
In short these are the highlights for this PR

1. Implemented the disband mechanic, and separated from unit removal without reward.
2. LuaMethod annotation
3. Wrote c# methods to specifically be utilized by the Lua layer
4. Added option to cache delegates instead of having to parse them each time we call a lua function

This PR introduces more than the disband mechanic itself. Its implementation is straight forward, but its rewards calculations are done completely on the Lua layer.
I chose not to add the function call for the reward to the json because it's gameplay specific and used only in one specific place, and instead added a hook for a custom implementation. Let me know your thoughts on that.

After some thinking and experimentation, especially after implementing the Inflow calculation, I added a few methods that would make the Lua behaviour more customizable in general. Now, it's left to the modder to decide if they want to give back shields, or for example gold, or both, in any scenario. I think this is the way to move forward, expose more of these methods like, AddTech, or AddBuilding, etc, so we hardcode as little as possible to the c#.
You will see above these methods that I have added a new Attribute `LuaMethod`, which as I explain is there (for now) purely as visual aid. I didn't go and add this in the other methods I know are used by lua already, as I would like your thoughts on this first.

Another thing is the `ActionToEngineMsg`, I wanted to have the disband action as an async method, because it contains an animation. 
I had some problems with the current structure of  `ActionToEngineMsg` and Rider suggested to add another cunstructor, and it snowballed from there. This is the only thing I am not sure what it does 100% and if it has any side effects, but I playtested a bunch, and didn't see anything weird.

The other thing is, I added an option to cache the delgate connected to the lua function path, so that we don't have to parse the lua script everytime. I googled a bunch for this as well, aside from playtesting, if it had any unwanted effects, and the conclusion seems to be that's it's fine.

The whole implementation was triggered because I saw a TODO a few lines below `TechCostFor` in the `GameData` class which says

> TODO: See this this whole equation can be configurable

I was about to add a bunch of rules to make the disband reward mechanic friendlier than it would have been if it a straight up c# implementation, but I think I came up with a better plan.

Last side note, I think we need to clean up some classes, have some java like setters/getters, private variables, and reusable constructors, both for clarity and safety but also so that Lua can safely call them.

Again, let me know your thoughts, or if you have any questions.

PS I am working my way through what I actually started off to do initially, which is the capture mechanism of units, but I don't want to end up submitting a 50 file PR with all the surrounding changes I have come across along the way.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -810,15 +810,16 @@ public partial class Game : Node {
 
 		if (currentAction == C7Action.UnitDisband) {
 			if (CurrentlySelectedUnit == null || CurrentlySelectedUnit == MapUnit.NONE) {
+				log.Warning("Trying to disband null or NONE unit");
 				return;
 			}
 			popupOverlay.ShowPopup(
 				new ConfirmationPopup(
-					$"Disband {CurrentlySelectedUnit.name}? Pardon me but these are OUR people. Do \nyou really want to disband them?",
+					$"Disband {CurrentlySelectedUnit.name}? Pardon me but these are OUR people.\nDo you really want to disband them?",
 					"Yes, we need to!",
 					"No. Maybe you are right, advisor.",
 					() => {
-						new ActionToEngineMsg(() => CurrentlySelectedUnit?.disband()).send();
+						new ActionToEngineMsg(async () => await CurrentlySelectedUnit.Disband()).send();
 					}),
 				PopupOverlay.PopupCategory.Advisor);
 		}
--- a/C7/Lua/game_modes/base-ruleset.json
+++ b/C7/Lua/game_modes/base-ruleset.json
@@ -7188,7 +7188,8 @@
     "defaultDealDuration": 20,
     "treasuryInterestRate": 0.05,
     "maxInterest": 50,
-    "shieldCostPerGold": 4
+    "shieldCostPerGold": 4,
+    "shieldRateForDisbanding": 0.25
   },
   "techs": [
     {
--- a/C7/Lua/rules/civ3.lua
+++ b/C7/Lua/rules/civ3.lua
@@ -7,5 +7,6 @@ return {
   --]]
   terraforms = require "civ3.terraforms",
   terrain_improvements = require "civ3.terrain_improvements",
-  inflows = require "civ3.inflows"
+  inflows = require "civ3.inflows",
+  gameplay = require "civ3.gameplay",
 }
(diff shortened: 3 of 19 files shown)


ajhalme:
Looks good to my untrained eye. I'm sure having Lua hooks to customise mechanic behaviour outside the engine will prove useful.  :+1: 

--

On a loosely related note, I wonder about the organisation of the Lua directory. If the idea is to make it easy to replace civ3 elements and even mechanics with modded versions, should the boundaries between the base c7 parts, the civ3 parts, and future mod parts be drawn more clearly?

So, for example, instead of `Lua/rules/civ3/..`, `Lua/texture_configs/civ3/..`, `Lua/texture_configs/c7.lua`, and `Lua/game_modes/standalone.lua` (which is kind of like `Lua/game_modes/c7.lua`, no?), one could instead have different "flavours" up top: `Lua/civ3/rules/..`, `Lua/civ3/texture_configs/..`, `Lua/c7/texture_configs/..`, `Lua/c7/c7_game.lua`, and so on

I'm also thinking about an override mechanism where I could choose to "apply" all of `Lua/c7/..` content, but then override one specific file or even a single Lua element in my own `Lua/mymod/texture_configs/..`. If one mod is effectively one extra top level folder in the Lua directory, that might be appealing from a mod management point of view. I can even imagine a simple UI where one selects which mods to apply, and in what order: composable mods.

stavrosfa:
> I'm sure having Lua hooks to customise mechanic behaviour outside the engine will prove useful. 👍

I am a bit afraid that we will end up needing to have ALL of the behaviours exposed to Lua, so it can be unimaginably moddable, and i don't know if it will end up hurting performance, or at this point, someone might as well modify their version of c# directly. I don't know, I don't have the end vision in mind, I am might be exaggerating.

The lua organization sure can be improved, I want to think that we do this regardless as we go on. The part you mention `Lua/game_modes` is super fresh, it was merged just yesterday.

For a short answer about composable mods, this is what [the PR I mention](https://github.com/C7-Game/Prototype/pull/848) is all about. It's too much for me to write here. I suggest if you want to continue this conversation, let's do this over discord, there is already a thread discussing some of this [here](https://discord.com/channels/909680526498082837/1457726459895484598) you might want to go over.

esbudylin (approved):

PR #892, opened 2026-02-19
Add river delta textures

### LINKED ISSUE #107: Rivers - Add Delta Support

ajhalme:
Fixes #107, missing river delta textures. See also Draft PR #433 about a tile map variant, linked to #407.

There are two river textures in the Civ3 Art assets, a regular set and one for river deltas, to be drawn when the river meets the sea. The delta textures are not currently being loaded in.

This patch adds the textures, and applies according to a simple heuristic. Given the close relationship between the two tile sets and how they are deployed, the implementation reduces to a simple check to see whether the tile drawing is happening next to water tiles or not.

[Sample save file](https://github.com/user-attachments/files/25402664/107_river_deltas.json)

Given how closely related to the regular river tiles the delta tiles are, regardless of how the tile map effort proceeds, #433 can probably be closed off.

--

river_delta_trio
riverdelta_old
riverdelta_new
riverdelta_original

Note the non-delta river segments in the delta texture:
tilemap_riverdelta

### DIFF
--- a/C7/Lua/texture_configs/civ3/terrain.lua
+++ b/C7/Lua/texture_configs/civ3/terrain.lua
@@ -86,6 +86,8 @@ terrain.marsh = {
 
 terrain.river = TERRAIN .. "mtnRivers.pcx"
 
+terrain.river_delta = TERRAIN .. "deltaRivers.pcx"
+
 terrain.tnt = TERRAIN .. "tnt.pcx"
 
 terrain.jungle = {
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -362,43 +362,80 @@ public partial class RiverLayer : LooseLayer {
 	public static readonly Vector2 riverSize = new Vector2(128, 64);
 	public static readonly Vector2 riverCenterOffset = new Vector2(riverSize.X / 2, 0);
 	private ImageTexture riverTexture;
+	private ImageTexture riverDeltaTexture;
 
 	public RiverLayer() {
+		// Both textures are 4x4 grids of equal dimensions
+		// The delta texture has four proper deltas textures and 12 other river segments
 		riverTexture = TextureLoader.Load("terrain.river");
+		riverDeltaTexture = TextureLoader.Load("terrain.river_delta");
 	}
 
 	public override void drawObject(LooseView looseView, GameData gameData, Tile tile, Vector2 tileCenter) {
-		//The "point" is the easternmost point of the tile for which we are drawing rivers.
-		//Which river graphics to used is calculated by evaluating the tiles that neighbor
-		//that point.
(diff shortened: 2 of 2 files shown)


stavrosfa (approved):
Thanks! I tested it with a few maps, looks good to me.

stavrosfa:
Hmm, I think I found something.

Editor
image

vs

OpenCiv3
image

I think we need to also check that it's the "ending" of the river too, so only one neighbor can have a river if it's a delta

ajhalme:
Revised the RiverLayer a bit more.

Existing logic assumes that rivers are symmetrical across the tile edge, but this doesn't seem to be the case. For example, if one tile has a NW river, the tile NW from it doesn't necessarily have a river to the SE. This might be an implementation issue deeper in the river tile properties handling code. I suspect that rivers that start in a point, but don't flow anywhere (i.e., the (0,0) tile, or "oasis") get lost somewhere in the logic

Having looked at the rivers in C3C editor a fair bit, I'm fairly confident that there isn't a single tileset that is consistently being applied in a given tile context. I suspect there's some arithmetic variance being added in. I left the finer tileset selection as a TODO, as I didn't want to implement something that would mainly just make modding more complicated.

It seemed like there might have been elevation-based logic that determines which tileset to use, but I don't think that's the case. (I was hoping the texture name `mtnRivers` would have been a reference to _mountains_.)

This pattern (oasis tiles from both textures) is both not consistent aross similar tile situations, and unlikely to be random:
river_patterns

Some more before and after shots - "Donut Island" and "Mixed Terrain":
donut_island_oc3_old
donut_island_oc3_new
donut_island_c3c_ed
mixed_terrain_openciv3_old
mixed_terrain_openciv3_new
mixed_terrain_c3c_ed

stavrosfa:
Rivers are definitely not done yet, both in terms of generation and gameplay.

I didn't know there were 2 different tile sets, but it seems indeed that they follow some kind of pattern. I think it has more to do with map generation rather than a random arethmetic variance. Or perhaps the variance comes from what the tile/texture the river is sitting on is.

From a quick google search I found these 2 threads that might be useful in figuring out which is used and why

https://forums.civfanatics.com/threads/deltarivers-pcx-anyone-know-when-its-used-other-than-for-deltas.675402/
https://apolyton.net/forum/civilization-series/civilization-iii/122361-civilization-iii-bix-biq-file-format/page4

The only difference I see is where the red arrow points, we have a Delta, whereas in the Editor it shows up as a regular river (parhaps it's the beginning of the river?). Ours looks more like what I would expect though 😅 
Comp

In any case, I think this is ok to merge.

PR #893, opened 2026-02-19
Fix/unit movement

stavrosfa:
I think the bug was two-fold.

First all the issues after many many replays, seems to be that non combat units (Worker, Scout, Settler) tried to enter a tile where there was either a Barbarian or a Barbarian camp. The camps weren't accounted for.

Secondly, there were 2 early returns, but they weren't dependant on HasHostileCity() or HasHostileUnits() but rather, whether we raised a notice to the player. 

If you have any ideas on how we could unit test this effectively, let me know.

Right now I have tested this just by playing 10-15 games in a row in observer mode for 100 turns each, and I didn't crash. Previously every 3-4 games max, I had this issue.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -949,7 +949,7 @@ public partial class Game : Node {
 
 					// If we couldn't enter this tile without a war declaration,
 					// record which civ we need to declare war on.
-					if (!unit.CanEnterTile(tile, TileProbe.CombatProbe())) {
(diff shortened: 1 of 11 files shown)


esbudylin (approved):

stavrosfa:
I condensed the different options down to the 4 distinct cases we have, with somewhat more clear names.

I used this during my investigation for the bug to make it easier to understand at a glance what each probe does, seemed like a good thing to have overall.

esbudylin (approved):

PR #895, opened 2026-02-20
 Disable test parallelization; fix setup for PlayerRelationshipTest

esbudylin:
First, this PR disables running tests in parallel. By default, XUnit executes tests from different classes in parallel (however, tests from the same class are not executed in parallel against each other). Docs: https://xunit.net/docs/running-tests-in-parallel

For our project, this can be problematic, since some tests access and overwrite the `EngineStorage.gameData` global. Parallel execution is a likely cause of tests hanging. `SaveTest` contains a while loop that waits for engine messages while simulating the game. If `PlayerRelationshipTest` overwrites the engine’s `gameData`, the expected message will never arrive and the test will wait indefinitely.

Second, this PR fixes `PlayerRelationshipTest` initialization. When I was merging #848, I mistakenly put test setup logic inside a fixture. This was a blunder, since fixtures are created once per test class. What I need is to set up the test data per test method. To fix this, I moved the initialization logic into the test constructor, which is executed before each test: https://xunit.net/docs/shared-context.html

### DIFF
--- a/EngineTests/GameData/PlayerRelationshipTest.cs
+++ b/EngineTests/GameData/PlayerRelationshipTest.cs
@@ -1,43 +1,23 @@
 using System;
 using C7Engine;
 using C7GameData;
-using C7GameData.Save;
 using Xunit;
 using static C7GameData.PlayerRelationship;
 
 namespace EngineTests.GameData;
 
-public class GameDataFixture : IDisposable {
-	public SaveGame SaveGame { get; }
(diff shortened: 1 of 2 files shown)


stavrosfa (approved):
Thanks! I tested it locally, seems to be ok. Perhaps a little slower, but nothing extreme.

I think in a way we might be abusing the `EngineStorage.gamedata` during gameplay too.

PR #897, opened 2026-02-23
Add dynamic world size selectors in New Game

### LINKED ISSUE #800: Allow creating games with different world sizes

ajhalme:
Fixes #800, creating games with different world sizes. A continuation of #850 (and #847) , incorporating the suggested improvements around data-driven UI elements.

This patch extends the base ruleset with standard world sizes extracted from a C3C save file with DumpObject. I added `isDefault` as a way to indicate which one is the default world size.

On the UI side, I've replaced the previously disabled set of world size buttons with a dynamically generated set of `Civ3MenuButtons` for a nice uniform look. The default set is based on the base ruleset as loaded into the New Game context as a save file.

As a bit of polish, both the world size controller and the Opponent selection controller now have ScrollContainers.

world_size_selector

### DIFF
--- a/C7/UIElements/NewGame/WorldSetup.cs
+++ b/C7/UIElements/NewGame/WorldSetup.cs
@@ -2,12 +2,13 @@ using Godot;
 using System;
 using C7Engine;
 using C7Engine.Lua;
+using C7GameData;
 using C7GameData.Save;
 using Serilog;
 
 [Tool]
 public partial class WorldSetup : Control {
-	private static ILogger log = LogManager.ForContext<WorldSetup>();
+	private ILogger log = LogManager.ForContext<WorldSetup>();
 
 	[Export] TextureRect background;
 
@@ -57,13 +58,6 @@ public partial class WorldSetup : Control {
 	[Export] TextureRect billion4Large;
 	[Export] TextureRect billion5Large;
 
-	[Export] CheckButton tinySize;
-	[Export] CheckButton smallSize;
-	[Export] CheckButton standardSize;
-	[Export] CheckButton largeSize;
-	[Export] CheckButton hugeSize;
-	[Export] CheckButton randomSize;
-
 	[Export] TextureButton confirm;
 	[Export] TextureButton cancel;
 
@@ -74,6 +68,12 @@ public partial class WorldSetup : Control {
 	WorldCharacteristics.Age age = WorldCharacteristics.Age.Billion_4;
 	WorldCharacteristics.Temperature temp = WorldCharacteristics.Temperature.Temperate;
 	WorldCharacteristics.Climate clim = WorldCharacteristics.Climate.Normal;
+
+	private WorldSize _worldSize = WorldSize.Generic();
+
+	private int GameSeed => int.Parse(seedInput.Text);
+
+	private SaveGame _saveGame;
 
 	// Called when the node enters the scene tree for the first time.
 	public override void _Ready() {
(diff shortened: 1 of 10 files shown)


stavrosfa:
Looks fine at first glance, I 'll have a deeper look at the whole thing tomorrow, but I just wanted to give you some quick feedback on this.

I assume you have anchored the container to the top left corner of the screen/game/tscn (red arrow), whereas it should be anchored to the top left of the background image (green arrow). So when scaling the game window, we have the effect you see in the image.

You probably didn't do this by hand, so I am guessing that it's a hierarchy issue solvable by parenting the element to the image. I'd have a look at how the other elements are parented to make sure.

Screenshot_55

Also, I don't know how I feel about `isDefault` since (again, from having a quick look) it doesn't seem to be used anywhere. If the only reason, future or present, is for it to indicate that "this is the default", a comment is probably more than enough.

But let me know if you have other plans for it :)

C7/UIElements/NewGame/WorldSetup.cs
```diff
@@ -266,14 +266,51 @@ public override void _Ready() {
...
+
+	private void InitMapSizes() {
+		var worldSizeButtons = GetNode("%WorldSizeButtons");
+		var worldSizeButtonsContainer = worldSizeButtons.GetNode<VBoxContainer>("%WorldSizeButtonsContainer");
+
+		var defaultSizeButton = worldSizeButtons.GetNode<Civ3MenuButton>("%DefaultSizeButton");
+		var randomSizeButton = worldSizeButtons.GetNode<Civ3MenuButton>("%RandomSizeButton");
+		var worldSizeButtonGroup = defaultSizeButton.ButtonGroup;
+
+		var sizeRandom = GameSeed == -1 ? new Random() : new Random(GameSeed);
```
esbudylin: Wouldn't this conditions (`GameSeed == -1`) always be true? You call this method from `_Ready`, so it will always execute before the GameSeed is set. In general, I'm not sure if it's necessary to tie the random selection of a world size to the game seed.

ajhalme: Changed to always use a fresh `Random()`.

Dealing with random things can be annoying when testing, so I figured I'd try to tie the random to the seed, if given, but you are right that this would not cut it. One could probably change the logic so that pressing the random sets a flag, and actual random selection is done only at game init time in `CreateGame()`. But, as you say, it probably isn't even necessary/correct to tie this random to the game seed.

ajhalme:
Adjusted the hierarchy so that the world sizes scroller is under the background.

I've been working with Godot 4.6, but switched to 4.4.1 in the middle for this UI stuff, as I noticed Godot was changing the scene files a lot as. Godot 4.6 adds unique ids to all nodes. Blending two Godot's made the scene file and the editor go crazy. I hit several issues trying to get Godot 4.4.1 to re-create the node structure I initially created in the 4.6 version. For some reason the editor was desperate to try and embed the background texture into the scene, which then caused the scene file size to blow up, which then broke the editor. I was able to trigger this embedding just by trying to resize the new scroller and container.

Anyway, the scene node tree should be fine for now. I removed some of the pieces I had initially created in the UI and simplified the button creation as well.

The `isDefault` is used in the world size button generator to select one of the world sizes as the default choice. (The blue button highlighting, `isPressed`.) One fewer click if you just want to try something on the Player Page, say, or one could also use it to fix a preferred or custom world size for debugging, etc.

ajhalme:
Also added `C7/Textures/PlayerTextureUtil.cs.uid`, which seems to be missing from branch `Development`.

C7/UIElements/NewGame/WorldSetup.cs
```diff
@@ -266,14 +266,52 @@ public override void _Ready() {
...
+			// Dynamically create a new button for each world size in the game 
+			foreach (var ws in _saveGame.WorldSizes) {
+				var worldSizeButton = new Civ3MenuButton
+				{
+					Text = ws.name,
+					textPosition = Civ3MenuButton.TextPosition.TextRightOfIcon,
+					FontSize = 0,
+					ButtonGroup = worldSizeButtonGroup,
+					ToggleMode = true,
+					ButtonPressed = ws.isDefault
```
stavrosfa, line 304: Here, if I add another custom entry in World Sizes and also make it the default, in the UI it will only have the last one selected based on the json entry order. That's good because we want only one selected at a time, but can we document this in the code? I wish we could do this in the json, but we can't.
But, if I just start a new game, it won't actually take my Custom map dimensions (160x160) and will generate a 100x100 map. I think it's because even if the button looks to be selected, it's not actually, and somehow has the Standard map selected behind the scenes.
You can verify this, by adding a new entry in the json (I added one as the very last entry) with some custom size and make this also default, start the game, click  `New Game` -> click on next, next, and you are in the game. Now press CTRL+G (to show the grid) and then CTRL+SHIFT+ALT+G (to show coordinates) and see that it didn't take in your custom size.

C7/UIElements/NewGame/WorldSetup.cs
```diff
@@ -266,14 +266,52 @@ public override void _Ready() {
...
-		randomSize.Visible = false;
+		_saveGame = GameModeLoader.Load(GamePaths.GameModesDir, GamePaths.GameMode);
+
+		InitMapSizes();
+	}
+
+	private void InitMapSizes() {
+		var sizeRandom = new Random();
+
+		var worldSizeButtonsContainer = GetNode<VBoxContainer>("Background/WorldSizeButtonsScroller/WorldSizeButtonsContainer");
```
stavrosfa: Can this be Node be a reference that we assign in Godot instead of hardcoded (like above with `[Export]`) ? That way we won't have to worry about renaming it/changing the hierachy.

C7Engine/C7GameData/WorldSize.cs
```diff
@@ -0,0 +1,29 @@
...
+		public int width;
+		public int height;
+
+		public int optimalNumberOfCities;
+		public int techRate;
+
+		public int distanceBetweenCivs;
+		public int numberOfCivs;
+
+		public bool isDefault;
```
stavrosfa, line 14: I am leaving a comment here, so I don't bloat the others.

This is actually cool, I didn't catch it at first.

This pertains to this functionality, but I think is beyond the scope of this PR, but first, we should do this for every button panel we have, and second, as an fyi, I opened an [issue ](https://github.com/C7-Game/Prototype/issues/898) that is present to all these button panels, where we can seeminglly select multiple buttons.

C7/UIElements/NewGame/WorldSetup.cs
ajhalme: Bah, I forgot to wire the apply for the retrieved world size object. The `isDefault` setting should work now.

C7/UIElements/NewGame/WorldSetup.cs
ajhalme: Changed accordingly.

stavrosfa (approved):
I think this is good to merge. Leaving it up for a while in case anyone else wants to chime in

PR #901, opened 2026-02-26
Spawn barbarian sea units on sea tiles

### LINKED ISSUE #860: Barbarian galleys spawn position
Barbarian sea units (Galleys) spawn on the same tile as the camp https://github.com/C7-Game/Prototype/blob/85e080916e106c4dd2c392bca451fd3e3862159f/C7Engine/EntryPoints/TurnHandling.cs#L135

We should be spawing them on a sea tile instead, that is also not a lake and that is not already occupied by another civ's unit.

If you are unsure as to how to get that sea tile, you can use methods like
https://github.com/C7-Game/Prototype/blob/85e080916e106c4dd2c392bca451fd3e3862159f/C7Engine/C7GameData/Tile.cs#L645

or 

https://github.com/C7-Game/Prototype/blob/85e080916e106c4dd2c392bca451fd3e3862159f/C7Engine/C7GameData/Tile.cs#L685

ajhalme:
Fixes #860, Barbarian galley spawning.

I refactored the barbarian spawning logic and pushed it into its own `BarbarianInteractions` class, to line up with some of the prior art in `/Entrypoints`.

Spawning new units in the game should probably be centralised somewhere. I pushed the spawning into GameData, as the City class routine for adding units can probably be made to reuse it there.

barb_galleys

### DIFF
--- a/C7Engine/EntryPoints/BarbarianInteractions.cs
+++ b/C7Engine/EntryPoints/BarbarianInteractions.cs
@@ -48,9 +48,9 @@ public class BarbarianInteractions {
 		if (unitType.IsLandUnit())
 			return camp;
 
-		// Spawn sea units on a sea tile, but not in a lake or on a tile occupied by another player
+		// Spawn sea units on a coast tile, but not in a lake or on a tile occupied by another player
 		bool CanSpawnSeaUnits(Tile t) =>
-			t.IsWater() && !t.isFreshWater && t.unitsOnTile.TrueForAll(u => u.owner == player);
+			t.IsCoast() && !t.isFreshWater && t.unitsOnTile.TrueForAll(u => u.owner == player);
 
 		if (unitType.IsSeaUnit()) {
 			// Check first two ranks around camp for a suitable tile, or bail with a null 
--- a/C7Engine/EntryPoints/BarbarianInteractions.cs
+++ b/C7Engine/EntryPoints/BarbarianInteractions.cs
@@ -25,7 +25,9 @@ public class BarbarianInteractions {
 		foreach (Tile camp in spawningCamps) {
 			UnitPrototype unitType = SelectBarbarianUnitType(gameData.barbarianInfo, camp);
 			Tile tile = SelectSpawnTile(barbPlayer, camp, unitType);
-			gameData.SpawnUnit(barbPlayer, unitType, tile);
+			if (tile != null) {
+				gameData.SpawnUnit(barbPlayer, unitType, tile);
+			}
 		}
 
 		return barbariansToSpawn;
@@ -47,9 +49,13 @@ public class BarbarianInteractions {
 			return camp;
 
 		// Spawn sea units on a sea tile, but not in a lake or on a tile occupied by another player
+		bool CanSpawnSeaUnits(Tile t) =>
+			t.IsWater() && !t.isFreshWater && t.unitsOnTile.TrueForAll(u => u.owner == player);
+
 		if (unitType.IsSeaUnit()) {
-			return camp.FindInRing(1,
-				t => t.IsWater() && !t.isFreshWater && t.unitsOnTile.TrueForAll(u => u.owner == player));
+			// Check first two ranks around camp for a suitable tile, or bail with a null 
+			return camp.FindInRing(1, CanSpawnSeaUnits)
+				?? camp.FindInRing(2, CanSpawnSeaUnits);
 		}
 
 		// Default to camp 
--- a/C7Engine/EntryPoints/BarbarianInteractions.cs
+++ b/C7Engine/EntryPoints/BarbarianInteractions.cs
@@ -1,0 +1,58 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using C7GameData;
+
+namespace C7Engine;
+
+public class BarbarianInteractions {
+	public static int SpawnBarbarians(GameData gameData) {
+		Player barbPlayer = gameData.players.Find(player => player.isBarbarians);
+
+		// A random 5% of camps will spawn a unit each turn. Shuffle the
+		// camps to make this random.
+		int barbariansToSpawn = (int)Math.Ceiling(GameData.rng.Next(gameData.map.barbarianCamps.Count) / 20.0);
+		List<int> tileIndicies = Enumerable.Range(0, gameData.map.barbarianCamps.Count).ToList();
+		GameData.rng.Shuffle<int>(CollectionsMarshal.AsSpan(tileIndicies));
+
+		// Sample barbarian camps
+		var spawningCamps = tileIndicies
+			.Select(i => gameData.map.barbarianCamps[i])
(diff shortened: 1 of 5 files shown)


ajhalme:
As an aside, I was experimenting with a testing idea where one could specify a map in ASCII, *Dwarf Fortress style*, and then use some generators to come up with a full testing setup.

Something like:
```
	var tiles = AsciiMapBuilder.Build(@"
	     ...%~====
	     ..@%~====
	     ...%~====
	");
```
where `.` is plains, `@` is a city, `%` is coast, `~` is sea, and `=` is ocean.

I kinda got something to work, but just getting the tiles is still some ways away from enough context for stuff like testing `BarbarianInteractions.SelectSpawnTile(..)`.

C7Engine/EntryPoints/BarbarianInteractions.cs
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	public static Tile SelectSpawnTile(Player player, Tile camp, UnitPrototype unitType) {
+		// Spawn land units at the camp
+		if (unitType.IsLandUnit())
+			return camp;
+
+		// Spawn sea units on a sea tile, but not in a lake or on a tile occupied by another player
+		if (unitType.IsSeaUnit()) {
+			return camp.FindInRing(1,
```
stavrosfa: What happens if the predicate fails to find a suitable tile because for example is occupied by an other civ's units, which returns null? Should we continue with the spawning of the unit?

stavrosfa (commented):
Good stuff!

Yeah, we spawn units in 3-4 different places, we should make this more centralized

stavrosfa:
I actually saw the newer comment after the review. 

This is a really cool idea! The are many times where I wanted to test something and I don't want to have a whole map, or having to do it manually. 
An editor where we could export a json with exactly what we need would be helpful, but that's not something that's gonna happen any time soon. So, if this can support a variety of map elements, it can fill this gap in the mean time.

If I were you, I would open a discussion in discord and/or put together something more concrete, open a pull request as a proposal, get the other dev's opinions, and move from there.

C7Engine/EntryPoints/BarbarianInteractions.cs
ajhalme: If we think of the barbarian camp as a city, spawning galleys there should be fine - so this whole PR is somewhat unnecessary.

But if the thinking is that the camp specifically *isn't* a city, it perhaps doesn't have the means to host a naval vessel, then bailing out with a null tile and not spawning anything feels reasonable. This doesn't need to be more complicated than that.

It's your issue - you tell me!

ajhalme:
Yeah, I've been missing a save file editor as well.

Claude had a nice idea about using [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902) as the engine of the editor. Recording changes to the JSON as a series of patches would make editor workflows (undo/redo) quite simple to implement. JSON Patch is also very forgiving for changes in the data model, as it targets specific paths and ignores other things.

And with JSON as the data format, it was quite simple for Claude to throw together a little HTML prototype. After a bit of iteration, I ended up with this: 
[openciv3_save_editor_proto_v0.4.html](https://github.com/user-attachments/files/25561669/openciv3_save_editor_proto_v0.4.html)

stavrosfa:
What I had in mind was something even simpler. I meant, if you have a look at [this unit test](https://github.com/C7-Game/Prototype/blob/eedeac5dd7a51369a06030f95a06e950c3e0b140/EngineTests/AI/Pathing/EdgeWalkerTest.cs#L15), we need 3-4 tiles for the test (or the galley spawn from this PR would also be a good example)
So, something like you described using ASCII, or a super minimal json just having the tile info.

A fully fledged editor is another story 😅 

Also, I toyed with the idea last week, to have a Debug view during the game, if we run it with a particular flag, so we can have , let's call them "Tools", that give real time feedback for stuff like, "what's the cost of going on a mountain", or, "does this path have a river crossing?"

example
Screenshot_63

So I would eventually want to be able to either execute a command in a console (CSGO style if you are familiar), for example 
`add_gold(Player-1, 100)`

or have a button or something to do this.

For changing small stuff, I find it tedious to save, open the save, fix what I need, reload, etc.

The only thing that I am missing and want to have, is conditional compilation with preprocessors, as for example  I would do in Unity:

```
#if UNITY_EDITOR

    code I only want to run in a certain mode

#endif
```

C7Engine/EntryPoints/BarbarianInteractions.cs
stavrosfa: >If we think of the barbarian camp as a city

I have though about this in the past, and I think the best option is no (at least for now). A camp is nothing but a "special" tile, that produces a very particular set of units, and yields gold if any other civ steps on it.

> then bailing out with a null tile and not spawning anything feels reasonable. 

I think that's the best option. The alternative would be to look even further from the immediate set of tiles (ring 1), and spawn it there (ring 2,3 etc)

I hadn't thought of this when I created the issue, it just seemed like a good issue for someone to get familiar with some part of the code.

Let's go with not spawning anything for now if the tile is null, we can revisit in the future if need be.

ajhalme: Added ring 2 and a bailout with null.

I figured it could be a nice dynamic, a valid strategy, to try and get a naval blockade going. Three ranks would be a lot, but two is a nice challenge.

Here my armada wasn't enough to cover the rank 2 tile north of the camp:
armada

ajhalme: Also noticed that the galleys are not aiming their broadsides against one another, but rather shoot arrows 90 degrees in the wrong direction.

ajhalme:
Dev tools and a debug console could both be really handy.

C7Engine/EntryPoints/BarbarianInteractions.cs
stavrosfa: Cool, I am fine with that. Since you added this though, shouldn't we add an extra check for the tile to be only a coast tile, and not a sea/ocean tile?

>Also noticed that the galleys are not aiming their broadsides against one another, but rather shoot arrows 90 degrees in the wrong direction.

Yeah, I had a quick look, there is an unit ability "Rotate Before Attack", which is not implemented on our side as far as I can see 
https://github.com/C7-Game/Prototype/blob/eedeac5dd7a51369a06030f95a06e950c3e0b140/QueryCiv3/BiqSections/Prto.cs#L115

A units' direction in a fight right now is handled like this
https://github.com/C7-Game/Prototype/blob/eedeac5dd7a51369a06030f95a06e950c3e0b140/C7Engine/C7GameData/MapUnit.cs#L304

We are also missing the "half" step an attacking unit makes before the attack.

ajhalme: Narrowed the spawnable tile check to coast only.

stavrosfa (approved):
Sorry for the long back and forth, this looks ready to me now. Again, leaving it up for a while for anyone else to chime in.

Issue #218, opened 2022-04-08
Teach the AI to attack barbarian camps


QuintillusCFC:
Playing through 150+ turns, it became clear that the AI should learn how to clear out barbarian camps.

With random exploration in Babylon, the AI would de facto clear out camps because enough random movement around their home area eventually cleared out the barb camps.

Now that the AI explores to reveal tiles, and then retreats to defend cities, barbarian camps can remain the entire game.

This will also be a case where priorities have to be balanced - do we send out another explorer, or clear that barb camp?

PR #903, opened 2026-03-01
Add barbarian activity levels

ajhalme:
This patch adds barbarian activity levels, in a few closely related pieces:
- New UI controls for setting barbarian activity level in New Game, based on Civ3
- Wire barbarian activity level to game state, save games, loading
- Extend barbarian unit spawn engine logic to allow activity level parameterisation
- Extend existing barbarian AI logic with an overarching strategy based on activity level
- Revise barbarian AI to more cleanly separate observation and decision logic, to potentially enable more nuanced tactics

If the thinking was to make barbarian behaviour more customisable, I feel extending this framework with a "Custom" activity level or strategy should be doable. I haven't included the standard activity levels as game options in the same as world sizes, though this could be doable. Making it all fully dynamic so that one could have complete control over barbarian activity level options in a mod/scenario is probably doable, but quite involved.

Longer term, it would be interesting to expose the game state / the observation and decision points as hooks somehow so that one could wire-in a completely separate external AI process.

I haven't playtested this anger, the parameterisation is just a first stab: *Sedentary* seems to work as intended, and with the proposed double digit spawn rate and reduced exploration, Raging can be really something even with simple unit-based barbarian tactics. I've seen Raging camps hold their own well over 200 turns against AI players until AI players crash their economy and the game!

---

*From the Civ3 manual*

### Barbarian
You can also set the level of barbarian activity in the game.
**No Barbarians:** Players who really hate barbarians can choose to play in this ideal barbarian-free world.
**Sedentary:** Barbarians are restricted to their encampments. The surrounding terrain is free of their mischief.
**Roaming:** Barbarian settlements occasionally appear, but less frequently and in smaller numbers than at higher levels. This is the standard level of barbarian activity.
**Restless:** Barbarians appear in moderate up to significant numbers, at shorter intervals than at lower levels.
**Raging:** You asked for it! The world is full of barbarians,and they appear in large numbers.
**Random:** This option randomly selects a Barbarian setting.

--

Civ3 UI:
og_new_game_slice

OpenCiv3:
openciv3_newgame_barbs


ajhalme:
Loosely connected to #396 for #139. Implementing the tribe idea, having units attack together, will be key to making the different barb activity levels feel right.

It seems #218 is not an issue any more.

C7Engine/AI/BarbarianStrategy/BaseStrategy.cs
```diff
@@ -0,0 +1,96 @@
...
+
+		var orientation = await Orient(player, unit);
+		var plan = await Decide(player, unit, orientation);
+		var result = await Act(player, unit, plan);
+
+		// TODO: store result
+	}
+
+	/// <summary>
+	/// Wake the unit if a foreign unit or the borders of a civ are in sight.
```
TomWerner, line 36: This might be a nice use of `GetTilesVisibleToUnit` in `TileKnowledge.cs`, which is currently private, but it would make barbarians on mountains be able to "see" further for example

TomWerner (commented):
I didn't have time to play test this (or really look in that much detail, sorry) but it looks very cool!

C7Engine/EntryPoints/BarbarianInteractions.cs
```diff
@@ -33,6 +40,27 @@ public static int SpawnBarbarians(GameData gameData) {
 		return barbariansToSpawn;
 	}
 
+	/// <summary>
+	/// 
+	/// </summary>
+	private static float DetermineSpawnRate(BarbarianActivity activity) {
```
stavrosfa, line 48: Question, just because I am curious. Where are these values coming from? Do you see any value in extracting them to the json so they can be parameterized?

Also same question for the strategies, where we do for example
`GameData.rng.Next(100) < 50`

stavrosfa (approved):
Another comment here, because I couldn't find an appropriate place for it.

Shouldn't we disable barb camp spawning if the option is None?

https://github.com/C7-Game/OpenCiv3/blob/28c483c491758c8cc434b4fe997fd4e32a9d3186/C7Engine/MapGenerator.cs#L122

https://github.com/C7-Game/OpenCiv3/blob/28c483c491758c8cc434b4fe997fd4e32a9d3186/C7Engine/MapGenerator.cs#L1298

Another TODO might be to have more/less camps based on the difficulty and/or barbarian level.

All in all, good stuff!

C7Engine/AI/BarbarianStrategy/BaseStrategy.cs
stavrosfa: Which we could extract to be configurable in the json perhaps

ajhalme: Applied GetTilesVisibleToUnit and left a comment about configurability.

ajhalme:
> Shouldn't we disable barb camp spawning if the option is None?
> [Might we have] more/less camps based on the difficulty and/or barbarian level.

Added this in, there was actually a TODO for this.

I was remembering that barbarian camps double as "goody huts", but that can't be right.

It might even be that scaling the number of camps is the primary element of the activity level. The description does suggest behaviour changes as well, but hard to say.

C7Engine/EntryPoints/BarbarianInteractions.cs
ajhalme: Current activity level parameters aren't based on anything except casual play testing.

I left some comments/TODOs to highlight that more work is needed if we want to make the barbs behave the way they do in Civ3.

For this, my target was simply to have some kind of a gradient of barbarian behaviour so that there's a reason to include the missing UI elements. Making things configurable would probably be the preferred option over trying tease out the perfect parameters. The AI in the original system might of course have worked in a completely different way, so there aren't really "true" parameters for any of this.

stavrosfa:
Sorry, I saw the TODO, but in my mind it didn't include the "no camps" part, I thought it was a TODO for "calculate how many" when it would have been implemented

C7Engine/EntryPoints/BarbarianInteractions.cs
stavrosfa: That's fine, I was just curious whether you pulled this info out of some source I haven't seen.

I think it's ok for now as a base, behaviour can be tweaked and refined as we go.

I don't think that will ever capture the true original AI mechanics 100%, and that's good in my opinion

Issue #899, opened 2026-02-24
Custom map sizes crash the game


stavrosfa:
A few examples I tried, that completely crash the game.

Perhaps not always, but some times a factor might be the number of players and the distance between them. All these entries were copies of the Huge map, only changing the height/width. But I tried to lower these too, for some it worked, for others still throwing an exception and crashing.

> 50x50

```
ERROR: Unhandled exception
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at C7Engine.GameSetup.AddPlayer(SaveGame save, Civilization civ, Boolean isHuman) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7Engine\GameSetup.cs:line 93
   at C7Engine.GameSetup.PopulatePlayers(SaveGame save) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7Engine\GameSetup.cs:line 69
   at C7Engine.GameSetup.Populate(SaveGame save) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7Engine\GameSetup.cs:line 36
   at PlayerSetup.<>c__DisplayClass23_0.<CreateGame>b__0() in C:\Users\Stavros\Documents\C7\C7-Prototype\C7\UIElements\NewGame\PlayerSetup.cs:line 230
   at System.Threading.Thread.StartCallback()
   at: void Godot.NativeInterop.ExceptionUtils.LogUnhandledException(System.Exception) (/root/godot/modules/mono/glue/GodotSharp/GodotSharp/Core/NativeInterop/ExceptionUtils.cs:132)
Unhandled exception. System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at C7Engine.GameSetup.AddPlayer(SaveGame save, Civilization civ, Boolean isHuman) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7Engine\GameSetup.cs:line 93
   at C7Engine.GameSetup.PopulatePlayers(SaveGame save) in C:\Users\Stavros\Documents\C7\C7-Prototype\C7Engine\GameSetup.cs:line 69
[...]

PR #904, opened 2026-03-01
Fixes for custom map issues

ajhalme:
Fixes/mitigations for #899, custom map pains.

First issue is due to map gen bailing out too early due to extreme custom maps failing to honour civ distancing rules even after initial halving mitigation. This patch allows progressively closer and closer civ placement, down to a hard floor of 3.

Second issue is due to odd-valued map dimensions tripping up some of the tiling / map gen logic. Instead of dealing with stride issues and tile indexing maths, a simple mitigation is to adjust the world characteristics and and shrink world size by one relative to the spec, so dimensions become even-valued. Odd-valued maps should be doable, but support for such probably isn't a high priority.

Final fix is to unrelated to the reported issue. `TileToDraw` in `MapView` seems to be some sort of texture sorting based performance enhancement. The tile comparison logic inside the mechanism generously allows for null comparison tiles on the right hand side, but doesn't handle left hand side null tiles. Worse still, the current error handling means that the right hand side wins every one of these degenerate comparisons, resulting in chaos in things like sorting algorithms. I was hitting some poor performance / freezing due to infinite looping. The fix cleans up the comparison with consistent null handling.

### DIFF
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -47,13 +47,9 @@ public partial class TerrainLayer : LooseLayer {
 		}
 
 		public int CompareTo(TileToDraw other) {
-			// "other" might be null, in which case we should return a positive value. CompareTo(null) will do this.
-			try {
-				return this.tile.ExtraInfo.BaseTerrainFileID.CompareTo(other?.tile.ExtraInfo.BaseTerrainFileID);
-			} catch (Exception) {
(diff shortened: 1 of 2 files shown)


TomWerner (approved):

TomWerner:
Thanks!

PR #905, opened 2026-03-03
Fix QuickStart

ajhalme:
Looks like Quick Start broke during the base game / save file consolidation effort.

Git bisect suggests it was one of these commits:
19aada80da27a36d8e81dbea90ebc89da175ba5f
42c369674f31968c36c5ff2dd81949157e2076cc

This patch adds a dedicated class that does some default inits, enough to launch a Quick Start game.

### DIFF
--- a/C7/UIElements/NewGame/QuickStartSetup.cs
+++ b/C7/UIElements/NewGame/QuickStartSetup.cs
@@ -1,0 +1,49 @@
+using Godot;
+using System;
+using System.Linq;
+using C7GameData;
+using C7Engine;
+using C7Engine.Lua;
+using Serilog;
+
+public partial class QuickStartSetup : Node {
+	private static ILogger log = LogManager.ForContext<ScenarioSetup>();
+
+	public static void Init(GlobalSingleton globalState) {
+		log.Information("Setting up a QuickStart game");
+
+		globalState.ResetLoadGameFields();
+
+		var save = GameModeLoader.Load(GamePaths.GameModesDir, GamePaths.GameMode);
+
+		globalState.WorldCharacteristics = new WorldCharacteristics(save) {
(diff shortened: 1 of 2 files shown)


stavrosfa:
That's tottaly fine as a temp setup, I don't know if @esbudylin is working on something similar at the moment, or it's fine to merge

C7/UIElements/NewGame/QuickStartSetup.cs
```diff
@@ -0,0 +1,49 @@
...
+			age = WorldCharacteristics.Age.Billion_4,
+			climate = WorldCharacteristics.Climate.Normal,
+			temperature = WorldCharacteristics.Temperature.Temperate,
+			worldSize = WorldSize.Generic(),
+			mapSeed = new Random().Next(),
+		};
+
+		globalState.SaveGame = save;
+
+		Civilization player =  save.Civilizations.FirstOrDefault(x => x.name == "Netherlands") ?? save.Civilizations[1];
```
esbudylin, line 31: Since these strings ("Netherlands" and "Regent") are also used as default values in the `PlayerSetup` class, should we extract them into constants?

esbudylin (approved):
LGTM, with the exception of this minor comment. Feel free to merge.

C7/UIElements/NewGame/QuickStartSetup.cs
ajhalme: Yeah, I considered constants, but actually, what we want to do is make Quick Start conform to the original.

I was reading the manual again and noticed that Quick Start is supposed to run a game with the same parameters as what was set by the user when a New Game was last creates. So this setup information should be stored in a file somewhere. This file can then be initially set with some defaults, which may or may not not match the New Game defaults.

I'll open a separate issue for this behaviour, I think it would be useful for testing.

PR #912, opened 2026-03-06
Tile improvement fix

stavrosfa:
I had some issues when trying to load a json save where some tiles had railroads. As far as I can tell, this happens, because when we save the overlays, if there is a railroad, we don't save the road that invisibly exists underneath it.

So then, when loading, we try to add railroad to a tile that doesn't seem to have a road.

This does not happen with .biq and .sav files because when debugging I saw that we import both a road and a railroad overlay when laoding the tile info

So now, the only place where we check if we can add an improvement, is during gameplay and not when loading a save. I think it also makes  sense from a cleaner code standpoint, to have a dedicated method to add an improvement, rather than also checking if we can add it in the same method.

I don't think that we need to imitate what the .biq and .sav files do and save both improvements, as we can upgrade and downgrade on the fly.

### DIFF
--- a/C7Engine/C7GameData/Terraform.cs
+++ b/C7Engine/C7GameData/Terraform.cs
@@ -89,6 +89,8 @@ public class Terraform {
 		Effect?.Invoke(new(player, tile, this));
 
 		if (Improvement != null) {
+			if (!tile.overlays.CanAdd(Improvement))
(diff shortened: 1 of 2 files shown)


ajhalme (approved):
Seems reasonable.

PR #915, opened 2026-03-09
Unit art implementation that also supports variations and pedia icons

stavrosfa:
This PR tackles something that was discussed a few months ago on discord, Unit art.

This is by no means how the final product should look like, it's rather a step towrads it.

In short this PR does the following:

1. Parses the PediaIcons.txt for the game/scenario, extracts the Unit art, the unit's icon indexes, and pedia entries, and creates an `Art` object that contains all the art for each unit.
2. This includes variations for Eras, the captured worker textures, and Leader/Army textures. Workers/Settlers change their textures dynamically if we enter the next Era (usually Industrial).
3. The only thing that is not covered are Scientific Leaders, but I have left some relevant TODOs.
4. The Science Advisor is also updated as a consequence of this to show the unit icons.

I tested this with .sav, .biq and .json, everything seemed to be in order

Let me know what you think on the proposed scheme for the art. 

As I said, this doesn't need to remain as is in the future, as we can incorporate the INI details, custom animation details, etc, and finally even detach the whole Art json from the main json and have it as a separate one, but I think it's a good first step.

Also, this PR has a dependency on https://github.com/C7-Game/Assets/pull/3

### DIFF
--- a/C7/Lua/texture_configs/civ3/unit_icons.lua
+++ b/C7/Lua/texture_configs/civ3/unit_icons.lua
@@ -8,13 +8,42 @@ local unit_icons = {
   },
 }
 
-function unit_icons:map_object_to_sprite(unit_prototype)
-  if (unit_prototype:GetType().Name ~= "UnitPrototype") then
+local function isKeyInDictionary(dict, targetKey)
+  if(dict == nil) then
+    return false
+  end
+  for key, _ in pairs(dict) do
+      if key == targetKey then
+          return true
+      end
+  end
+end
+  
+-- Context - ItemContext (UnitPrototype proto, Player player)
+function unit_icons:map_object_to_sprite(context)
+  local proto = context.proto
+  local player = context.player
+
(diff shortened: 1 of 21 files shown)


C7/Lua/texture_configs/civ3/unit_icons.lua
```diff
@@ -8,13 +8,42 @@ local unit_icons = {
   },
 }
 
-function unit_icons:map_object_to_sprite(unit_prototype)
-  if (unit_prototype:GetType().Name ~= "UnitPrototype") then
+local function isKeyInDictionary(dict, targetKey)
```
esbudylin: If dict behaves like a regular Lua table, this function can be replaced by a one-liner: `dict and dict[targetKey]`

stavrosfa: Great, thanks! I am still very new to Lua and I am probably going more verbose than needed.

PR #916, opened 2026-03-11
Add a basic minimap

ajhalme:
Add a basic minimap:
- Render map tiles as pixels in an image (with a fixed palette for now)
- Muted terrain colours, super-saturated civ colours
- Support for click to center (!)
- Render viewport edge
- Supports zoom
- Supports different map sizes via stretching

Lots of TODOs around configurability, and I'm sure the colours could use some work, but it's a start. Already super handy.

Should be trivially parallelizable.

Tested briefly with civ3 and standalone modes.

minimap_large
minimap_standalone
minimap_tiny


stavrosfa (changes requested):
Brilliant stuff! It definetly needs some work, but it's a great feature to have!

I 'll dump here some comments in general.

First and most importantly, the go-to functionality is broken. When I press G and then click on the map, it moves the map rather than acepting the click.

On the visual side of things that could be future TODOs or if you want to tackle some more things in this PR:
1. Add the frame texture for the minimapp, this is the easy part now :P
2. I see cities in the original are a 3x3 pixel square to make them more visible. Now, sometimes they are 2x2, 2x3, etc.
3. The sea tiles that are inside someones territory are not shaded with the civ's color, but simply as the sea color.
4. The territory seems to have a bit of transparency, blending more naturally with the underlying map.
5. Right/Left clicks should be disabled when over the minimap (I think left click is ok, right click is not) 
Image

5. Dragging the square in the minimap doesn't move the map
6. Dragging on the map to move, should stop when over a UI element (like the minimap). This is implemented for the Lower Right box if you want to hava a look.
7. The square also doesn't support wrapping. If I move the map manually and wrap around, the square remains squashed.
8. When it's on the top/bottom edges of the map (that usually don't wrap around), it shouldn't get squashed
9. Fog of war tiles in the minimap (original), seem to have an extra layer of gradient, going from total black to semi-transparent black, then to whatever tile color they represent.
10. The square seems to be a bit thicker at times than it should
11. Sometimes it also seems that its size is not consistent.
12. We need to stop using gD.GetFirstHumanPlayer() and start passing the player controller as an argument to pretty much all the UI elements.
13. Perhaps render the minimap texture in layers? The terrain is unlikely to change, so we can have this as a base, then add forests, cities, etc on top? We could have a hook to recalculate what we need if there are changes. This would also enable us to quickly toggle between a geographic view of the map and a more full one (even add more as we go)

These are my thoughts so far

ajhalme:
> First and most importantly, the go-to functionality is broken. When I press G and then click on the map, it moves the map rather than acepting the click.

Hmm, not 100% sure what the intended functionality is here. Are you pressing G and then clicking the *minimap* and expecting the Goto to set the target there automatically, without centering the screen, or?

I tried to fix issue (5) by disabling the right click, maybe that changes the GoTo behaviour as well.

Changed the case order so sea tiles get drawn before civ owned tiles, which I think fixes (3).

Also had a go at enabling the frame texture, even made a variant of the one used in the standalone (see https://github.com/C7-Game/Assets/pull/4).

The layer suggestion is interesting, I'll give that a go. And fixing wrap should fix a few of the issues listed.

The minimap content is stretched by nature, so getting it pixel perfect with that technique might be a hassle. There might be a way to make it more stable by selecting the dimensions carefully, like only even widths allowed or something, etc.

--

Some more screens with the frames:

Screenshot at 2026-03-19 00-20-27
Screenshot at 2026-03-19 00-21-38

stavrosfa:
> Hmm, not 100% sure what the intended functionality is here. Are you pressing G and then clicking the minimap and expecting the Goto to set the target there automatically, without centering the screen, or?

No, imagine there is no minimap, I was trying to move a unit with the go-to command and clicking on a tile on the regular map, that's what was broken.

I briefly tried this PR, the go-to seems fixed, so we are good on that!

Issue 3 seems ok
Issue 5, now even the right click goes through as far as I can tell, and it moves the map, not the minimap :/

>The minimap content is stretched by nature, so getting it pixel perfect with that technique might be a hassle. There might be a way to make it more stable by selecting the dimensions carefully, like only even widths allowed or something, etc.

Not sure what you are referring to here

As I said in my previous comment, these are suggestions and TODOs, I don't expect you to solve them all or something, if everything is working, or rather, if nothing is broken, we can merge this at any point, add a bunch of TODOs, and come back to it later, so, no pressure

ajhalme:
Refactored the minimap drawing to multiple components and the map itself to multiple draw layers.

Also implemented basic bounds logic to handle drawing the minimap across the edges of the map.

Screenshot at 2026-03-24 01-17-24

ajhalme:
Added some color blending for transparency effects (city, fog over) -- the blend parameter need some tunic.

Also added basic minimap panning.

stavrosfa (approved):
I think this is a very good first iteration on this feature. I don't know if you plan on work more on this (since you are pushing updates), but if not, I think we can merge this at this point.

Some obvservations on the current state.

The white rectangle although now wrapping left and right, if I go on the bottom or top, it also wraps there, which it shouldn't in a map that doesn't wrap on the Y.

In some cases it's hard to distinguish the cities from the terrain or the borders and the fog, hence my earlier comment on cities having a fixed 3x3 size (or something like that) to make sure they "pop" and are easy to spot.

Example:
Image

Some more obvservations from what the original game does. 

I think they render a larger texture which then they scale down, the pixels in the original are very fine, whereas ours looks more blocky.
I think this is the same texture they use for the replay when you retire/win/lose etc, but as I said, scaled down.

If we toggle the Geography in the original, there is another bg layer, almost completely transparent, that helps blend the 2 layers fog (that we have now) with the overall background.

ajhalme:
Yeah, the viewport indicator wrapping needs to be driven by the game map's wrapping rules. I implemented the general case where we wrap both ways. It should be fairly simple to selectively enable each wrap, but it's not urgent to sort out: the main map itself will not wrap, so the minimap won't fully wrap either. It just spills over and around a bit.

I tried 3x3 cities and 2x2, but I couldn't get good results because of the uneven stretching. Drawing on a bigger texture and the scaling it down sounds like an interesting idea to try.

And yeah, the layer stuff I implemented is a quite primitive. I did get decent results by simply lerping the "overlay" colour to the pixel colour that was already there, when layering, but it the layer scheme can definitely be improved.  I even considered actual layers, having each "mini-layer" work on its own Image object, with some kind of flattening at the end. But I suspect this would be annoying to implement and might perform poorly as well.

I'll merge this in, I think it's a useful base feature to have as is. Note that C7-Game/Assets/pull/4 needs merging as well, or another custom frame to bring the minimap to the  standalone game.

PR #917, opened 2026-03-12
Implement culture groups for civilizations

stavrosfa:
As the title says.

All .biq and .sav files seem to have hardcoded the 6 culture group values, but on our end should be fairly easily expandable.

Right now these don't really have any effect in the game, but when we start loading culture specific building textures for example, we will start seeing some meaningfull changes.

### DIFF
--- a/C7Engine/C7GameData/Civilization.cs
+++ b/C7Engine/C7GameData/Civilization.cs
@@ -11,6 +11,11 @@ namespace C7GameData {
 		Female,
 	}
 
+	public class CultureGroup() {
+		public int index { get; init; }
+		public string name { get; init; }
+	}
+
 	public class Civilization {
 		public enum Trait {
 			Militaristic,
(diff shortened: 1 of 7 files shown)


C7Engine/C7GameData/Civilization.cs
```diff
@@ -39,6 +44,22 @@ public Civilization(string name) {
 		public int secondaryColorIndex;
 		public Gender leaderGender;
 
+		[JsonIgnore]
+		public CultureGroup cultureGroup { get; private set; }
```
ajhalme, line 48: What uses are there for the culture group? Is it just potential trade/reputation schemes and culturally linked start?

Just wondering if the extra de/serialization machinery is worth it vs just using `cultureGroupKey` in game  logic, together with the game knowing what culture groups there are.

stavrosfa: The short answer is textures. Check the PediaIcons.txt, there are building textures that are generic (mostly Great Wonders), but also culture and era specific.

The cultureGroupKey is used because we might want to add more fields/info in the CultureGroup iteself, and we wouldn't want to be copy pasting the same information across many civs. It simply a reference, we do the same with experienceLevel on the (EDIT) MapUnit or uniqueUnit currently in Civilization.

I hope that covers your question

ajhalme (approved):

PR #920, opened 2026-03-23
Biq and SAV enhancements

stavrosfa:
This PR in short:
Add version info for .biq and .sav files.
Add a few more flags in the GAME structs (both .biq and .sav) 
Some comments and observations on the data structures 
Fix bug(?) where if the player number is 0 in the biq file, certain dynamic parts have also 0 length

I came across this .sav file, where it was clear that some data had either uninitialized values (usually in the form of -842150451 which is 0xCDCDCDCD in hex), or had plain wrong values, but the save progressed normally in the original civ3 game.

What I tracked it down to was that, even with an unmodified conquests.biq file, if you start a game choosing Quick Start certain values in the binary where "missing", notably the dynamic parts in the .biq GAME struct, which contain a) the ids of the civs in game and b) their alliance statuses.

That caused the whole reading of the binary to be offset by some amount, which I think is constant regardless of settings (4 bytes * 31 players (ids) + 4 bytes * 31 players (alliances)), resulting in junk data after that point in the corresponding GAME struct.

It also seems that the game at this point also assigns some default values, regardless of what it's loading from, so for example, the plague name might be "Plague" in the conquests.biq, or even absent if it's a PTW file, but upon inspection after loading, the plague name has the value "Black Death", which seems to be hadcoded in the game exe itself (found using Ghidra).

Tha flags/comments and the version are bonuses that I came across/added during the investigation.

Ah, almost forgot, [Quintillus' editor source code](https://sr.ht/~adj/Civ_III_Scenario_Editor/) was a big help to be able to identify biq and sav versions, as well as the [decompression tool ](https://forums.civfanatics.com/resources/cross-platform-sav-biq-decompressor-command-line.22852/)

### DIFF
--- a/QueryCiv3/Biq.cs
+++ b/QueryCiv3/Biq.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using QueryCiv3.Biq;
 
 namespace QueryCiv3 {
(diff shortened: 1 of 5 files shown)


ajhalme (approved):
Looks very reasonable.

Issue #898, opened 2026-02-23
UI Buttons can all be selected simultaneously


stavrosfa:
If I click on a button and then drag my mouse away, I can select multiple options at once..
Pretty much as theimages show

<img width="526" height="407" alt="Image" src="https://github.com/user-attachments/assets/03c532a6-77bb-4905-bcd6-ea7c262d40aa" />
<img width="383" height="502" alt="Image" src="https://github.com/user-attachments/assets/5ef0d726-c8d1-4107-b7f0-a3029fdfb0e3" />
<img width="1003" height="218" alt="Image" src="https://github.com/user-attachments/assets/25c69be4-0d07-4d62-be3e-1bc8e71e4272" />
<img width="391" height="370" alt="Image" src="https://github.com/user-attachments/assets/498fdb90-ec21-4023-b903-95be6036a206" />

ajhalme:
Fixed with #921.

PR #921, opened 2026-03-25
Fix Civ3MenuButton visual multi-select issue

ajhalme:
Fix for #898.

Removing the alternative route to "effective press" results in the visual update doing the right thing. We want to only change to the "active" button texture when we have a true click, which in ["active on release"](https://docs.godotengine.org/en/4.4/classes/class_basebutton.html#property-descriptions) `ActionMode` is what we want.

I can't think of a case where the current behaviour is desired, but if we want Civ3MenuButton to change state already on the press down, we probably want to change things up so that the `ActionMode` ("active on press" vs "active on release") is made use of.


stavrosfa (approved):
Thanks, that seems to work well!

PR #922, opened 2026-03-26
Integrate time

stavrosfa:
This PR introduces the following

Read time data from the .biq/.sav files
Implement a mod friednly way to display the time to the player. In the main c# code we only calculate the "raw" turn number (e.x. -4000, when we display 4000 BC)
Besides the standard time units (years, months, weeks), days and hours are added. How the are displayed is as modable as the rest of the tim units.
Some tests to make sure the raw conversions are working properly.

### DIFF
--- a/C7/UIElements/LabelExtensions.cs
+++ b/C7/UIElements/LabelExtensions.cs
@@ -1,7 +1,7 @@
 using Godot;
 
 public static class LabelExtensions {
-	public static void SetTextAndCenterLabel(this Label label, string text, float additionalXOffset = 0) {
+	public static Label SetTextAndCenterLabel(this Label label, string text) {
 		//For the centered labels, we anchor them center, with equal weight on each side.
 		//Then, when they are visible, we add a left margin that's negative and equal to half
 		//their width.
@@ -10,6 +10,16 @@ public static class LabelExtensions {
 		label.HorizontalAlignment = HorizontalAlignment.Center;
 		label.AnchorLeft = 0.5f;
 		label.AnchorRight = 0.5f;
-		label.OffsetLeft = -1 * (label.Size.X / 2.0f) + additionalXOffset;
+		label.OffsetLeft = -1 * (label.Size.X / 2.0f);
+		return label;
+	}
+
+	public static Label AddXOffset(this Label label, float xOffset) {
+		label.OffsetLeft += xOffset;
+		return label;
+	}
+	public static Label AddYOffset(this Label label, float yOffset) {
+		label.OffsetTop += yOffset;
+		return label;
 	}
 }
--- a/C7/UIElements/LabelExtensions.cs
+++ b/C7/UIElements/LabelExtensions.cs
@@ -1,7 +1,7 @@
 using Godot;
 
 public static class LabelExtensions {
-	public static void SetTextAndCenterLabel(this Label label, string text) {
+	public static void SetTextAndCenterLabel(this Label label, string text, float additionalXOffset = 0) {
 		//For the centered labels, we anchor them center, with equal weight on each side.
 		//Then, when they are visible, we add a left margin that's negative and equal to half
 		//their width.
@@ -10,6 +10,6 @@ public static class LabelExtensions {
 		label.HorizontalAlignment = HorizontalAlignment.Center;
 		label.AnchorLeft = 0.5f;
 		label.AnchorRight = 0.5f;
-		label.OffsetLeft = -1 * (label.Size.X / 2.0f);
+		label.OffsetLeft = -1 * (label.Size.X / 2.0f) + additionalXOffset;
 	}
 }
--- a/C7Engine/C7GameData/TimeOptions.cs
+++ b/C7Engine/C7GameData/TimeOptions.cs
@@ -1,0 +1,220 @@
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Globalization;
+using System.Text.Json.Serialization;
+using C7Engine;
+using Serilog;
+
+namespace C7GameData;
+
+public enum TimeUnit {
+	Years,
+	Months,
+	Weeks,
+	Days,
+	Hours,
+}
+
+public sealed class TimeOptions {
+	public TimeUnit baseUnit { get; init; } = TimeUnit.Years;
+	public string negativeLabel { get; init; } = "BC";
+	public string positiveLabel { get; init; } = "AD";
+	public int startYear { get; init; } = -4000;
+	public int startMonth { get; init; } = 1;
+
+	public int startWeek { get; init; } = 1;
+	// custom
+	public int startDay { get; init; } = 1;
+	// custom
+	public int startHour { get; init; } = 12;
+	public int[,] timeScale { get; set; } = new int[,] { { 25, 25, 40, 50, 100, 100, 100, 50000 }, { 50, 40, 25, 20, 10, 5, 2, 1 } };
+
+	public int turnLimit { get; init; } = 540;
+
+	// the sum of all the turns included in the timeScale array
+	// e.x. in a standard game of 540 turns, the totalTurns amounts to 440 turns
+	// after that we default to 1 timeUnit/per turn
+	private int totalTurns = -1;
+
+	[JsonIgnore]
+	public int currentYear { get; private set; } = -1;
+	[JsonIgnore]
+	public int currentMonth { get; private set; } = -1;
+	[JsonIgnore]
+	public int currentWeek { get; private set; } = -1;
+	[JsonIgnore]
+	// custom
+	public int currentDay { get; private set; } = -1;
+	[JsonIgnore]
+	// custom
+	public int currentHour { get; private set; } = -1;
+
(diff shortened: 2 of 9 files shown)


C7/UIElements/LabelExtensions.cs
```diff
@@ -1,7 +1,7 @@
 using Godot;
 
 public static class LabelExtensions {
-	public static void SetTextAndCenterLabel(this Label label, string text) {
+	public static void SetTextAndCenterLabel(this Label label, string text, float additionalXOffset = 0) {
```
ajhalme: [Suggestion] Giving `SetTextAndCenterLabel` an offset parameter feels a bit off to me. Maybe the function should be renamed "SetTextAndPositionLabel" or something here.

Alternatively, we could go for a [fluent-style](https://en.wikipedia.org/wiki/Fluent_interface) approach here. If we change SetTextAndCenterLabel so that it returns the `Label`, and add a separate static method for adjusting the offset, we could have something like:

`myLabel.SetTextAndCenterLabel("My New Text").AddToXOffset(7f)`;

C7Engine/C7GameData/TimeOptions.cs
```diff
@@ -0,0 +1,220 @@
...
+
+public enum TimeUnit {
+	Years,
+	Months,
+	Weeks,
+	Days,
+	Hours,
+}
+
+public sealed class TimeOptions {
```
ajhalme, line 19: [Suggestion] A reasonable implementation. I wonder if a more object-oriented approach would be more extensible. Instead of several functions that do different things depending on `baseunit`, we could have `baseunit`-based instances and multiple implementations of the same methods.

ajhalme (approved):
Looks reasonable to me. Left a few suggestions.

C7Engine/C7GameData/TimeOptions.cs
stavrosfa: I understand what you are saying theoretically, but I am not sure what you mean in this context. Make each TimeUnit a class, with a start & current year?

Also these functions which I assume you are talking about like SetTimeUnitCurrent() or GetStartingPoint(), are doing such minor tasks, that having 5 implementations of each seems a bit excessive, no? Kinda sounds like premature optimization, with no clear benefit.

I am all about making it extensible, but right now, I am not seeing anything that would hold us back from extending it (which I also doubt we will), or perhaps I am not really getting what you mean by "extending".
If you want to give me a rough example, please do.

PS
One thing I wanted to do, was to have the TimeUnit enum be extensible by lua, so that we wouldn't be bound to the compiled c# code, but as far as I could understand, that's not possible with enums, so we are in bad luck there.

C7/UIElements/LabelExtensions.cs
stavrosfa: Yeah, that was kind of hacky on my part.

This function should probably be abandoned altogether for either something more Godot native, like using anchors etc, or a UI label that handles all that for us.

Pixel offsets in code is something I don't really like anyway, and there are people who want us to support larger resolutions, so I think this approach is faulty anyway, it was just my ocd wanting the labels to be more center than they were 😅

C7Engine/C7GameData/TimeOptions.cs
ajhalme: I was thinking `startYear/startMonth/startWeek/startDay/startHour` --> `startMoment` , and so on.

Stepping back a bit, it seems like the Lua integration is really doing all the heavy lifting here. All the code on the C# side is generic. I don't see any reason to distinguish between hours and days on the C# side.

If in Lua code, I know I'm in "hours mode", say, I don't need `currentHour` in the C# code. I just need some `currentMoment` (or whatever), that would then allow me to hold the state, and it seems to me like that bookkeeping could be the same for all "modes".

Feels a bit awkward to have the main time management handled with Lua on what is effectively `int`s, when one could have nice `DateTime` functions on the C# side, etc. But I understand the desire to have it all in Lua. I guess my ideal option would be to have the implementation of the "standard" time modes on the C# side, but then set things up with enough hooks so that one could have a "Custom" mode that a modder could then fully control with Lua.

But maybe the branching functions are making things more explicit and readable, and that's worth holding on to.

stavrosfa: > I was thinking startYear/startMonth/startWeek/startDay/startHour --> startMoment , and so on.

Does that mean that `startMoment` contains all of the above, or is a representation of just one time unit (which would be the current time unit)?

> If in Lua code, I know I'm in "hours mode", say, I don't need currentHour in the C# code. I just need some currentMoment

But what happens when you want to have something like Jan, 1945 (like in the WWII scenario), where you have multiple "states" for each time unit? Also, `currentHour` and the like, are only for display purposes, I don't think they are being used in the actual calculations, that's why I also chose to not include it in the json data.

>  I guess my ideal option would be to have the implementation of the "standard" time modes on the C# side

What if a mod wants to have 14 months in a year because it's on Mars or something, or anything that is non-standard? I think by exposing it the way I have in lua, the only future problem, would be adding more time units (e.x. decade), because there is no interface for that, or support variable, say, month lenghts in a year. Well, that, and the label translations maybe.

As I said, currently I am not seeing any clear benefit to massively changing the API, it's not that the implementation is the greatest ever, I just think it does the job and it provides moddability to a reasonable degree.

I would be open if anyone wanted to implement something with their style, but obviously offering the same output.

Last thing that's in my mind, although it's not like I designed the implementation based on that, and i don't want to offend our modder friends, but how much easier is for a modder to do something where they have to use polymorphism for `startMoment` , instead of a straight-forward `currentYear`? We could provide a more comprehensive API ourselves, sure, but doesn't this complicate things in this particular case for no real benefit?

ajhalme: I had a go at an alternative implementation. I used `DateTime` and its nice built-in serialization for some of the state handling: [Repo](https://github.com/ajhalme/OpenCiv3/tree/time-display-alt) [Commit](https://github.com/C7-Game/OpenCiv3/commit/da4a0996149ee461079be7e31a28d69a3e65c253)

Also added a `Custom` option, which should allow arbitrary time formats via Lua.

The main thing is to just make it so there's a way to change the current (serialized) moment field from Lua, and then to have the current (serialized) moment be interpretable and renderable as a string in Lua code.

stavrosfa: Thanks for taking the time to put together what you had in mind, I spent some time looking through this and playing with it, so here are my thoughts.

My first impression is that this seems way too convulated, on the c# side of things, and stepping a bit in a modder's shoes, it's pretty much unclear how much control I have on the json/lua side.

There are too many hardcoded assumptions being made, like that the negative year is "BC" and the positive "AD". Same with a week having 7 days, or the format of the time being displayed.
I best read time as dd/mm/yyyy for example (which is the correct one 😛), but our american friends might want mm/dd/yyyy, and we don't provide any support for that. Furthermore, we show the full date and time, like `6/6/1944 00:00` which I might not want be showing like that in my mod. Again, I have no control.

It was also very unclear too me, how do I setup a scenario with a date. I tried to add one where the start is D-Day (as in one of your tests) and the only way I could figure out what number to put in the `startMoment` field in the json, was run the test, have a failed assertion, and get a number like `613286208000000000` which I still don't know if it's that many hours or seconds or something with a 1/1/1970 base, or something completely different.
I tried to use an epoch converter, I don't think it could even process it. But even if it could, it's at the very least unintuitive.

I am not trying to bash on your version just because, I think this has a place in another system where the criteria and the business logic are different and more "real", but in this case, I honestly don't think it fits.

ajhalme: Fair enough. For one, having multiple `startX` seems aligned with the save files / Civ3 way of representing things. Building on that seems very reasonable.

The only correct date format, of course, is ISO 8601, but I figured I'd go for something a bit more presentable. ( see also https://xkcd.com/1179/ )

Let's just hope no-one wants cyclical time or time running backwards! :)

Might be an idea to add some tests to try out how turns get displayed in the standard Civ3 time modes, like I added in my tests.

stavrosfa: > Let's just hope no-one wants cyclical time or time running backwards! :)

I mean, everytime the unit % currentUnit == 0 it could simply do setCurrentUnit(currentUnit - 1) or something.

Also, for backwards time, since we have current and start time unit, couldn't we invert the operation and instead of adding on the current, subtracting?

I haven't tried any of these, but they seem very possible.

> Might be an idea to add some tests to try out how turns get displayed in the standard Civ3 time modes, like I added in my tests.

I wanted to, but then if the lua changes, the tests would start failing as well, since anyone modding, would be modding that file directly, not another file that overrides ours

Also, something that I really hadn't thought untill I saw your code.
With my version, unless you are using all time units, each unit can represent anything you want, that's not necessarily included in the c# side.

So say you want centuries and years in your display time text, but we only offer years. You could use the hour related fields for example, that are currently unused, change the implementation for hour to count for centuries, add the text and you are done.

Again, haven't tried this, but seems very doable.

Issue #910, opened 2026-03-05
Quick Start should be based on last New Game


ajhalme:
According to the Civ3 manual, Quick Start is supposed to run a game with the same parameters as what was set by the user when a New Game was last created.

As such, the data needed by Quick Start should probably be read from a file. This file can then be initially set with some defaults, which may or may not match the New Game defaults.

This mechanism should prove useful for testing and development as well.

Mentioned here: https://github.com/C7-Game/OpenCiv3/pull/905#discussion_r2886782116

ajhalme:
Implemented by #923.

PR #923, opened 2026-04-05
Implement Quick Start parameters persistence

albertone:
This PR addresses #910

Implements Quick Start with the last settings.

This PR implements the persistence of game parameters as described in the Civ3 manual for the **Quick Start** functionality. The game now remembers and replicates the settings (civilization, difficulty, world traits, and opponents) from the last manually created game.

When PlayerSetup calls StartGame, the settings are saved to C7.ini

The settings are loaded on QuickStartSetup to reproduce last game preferences.

### DIFF
--- a/C7/UIElements/NewGame/PlayerSetup.cs
+++ b/C7/UIElements/NewGame/PlayerSetup.cs
@@ -235,8 +235,7 @@ public partial class PlayerSetup : Control {
 		thread.Start();
 	}
 
-	private void PersistGameSettings(GlobalSingleton global)
-	{
+	private void PersistGameSettings(GlobalSingleton global) {
 		try {
 			WorldCharacteristics world = global.WorldCharacteristics;
 			C7Settings.SetValue("lastGame", "worldSize", world.worldSize.name);
--- a/C7/UIElements/NewGame/PlayerSetup.cs
+++ b/C7/UIElements/NewGame/PlayerSetup.cs
@@ -224,6 +224,8 @@ public partial class PlayerSetup : Control {
 			opponents = CollectSelectedOpponents()
 		};
 
+		PersistGameSettings(gameSetup);
+
 		// World generation can take a bit of time if multiple attempts are
 		// needed, so we don't want to tie up the UI thread.
 		Thread thread = new(() => {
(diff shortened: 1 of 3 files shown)


ajhalme:
Thanks for the PR, I'll take a look.

We like to keep the solution whitespace-linted: the builds' automated checks monitor this.

Can you run `dotnet format ./C7/C7.sln whitespace` in the root to format the files you modified? This is a useful one to include in your IDE's build settings as an alternative build.

C7/UIElements/NewGame/PlayerSetup.cs
```diff
@@ -235,7 +235,33 @@ private void CreateGame() {
...
+			C7Settings.SetValue("lastGame", "opponents", opponentsValue);
+
+			C7Settings.SaveSettings();
+		} catch (Exception e) {
+			log.Error(e, "Failed to persist game settings to C7.ini");
+		}
+	}
+
 	private void StartGame() {
+		GlobalSingleton global = GetNode<GlobalSingleton>("/root/GlobalSingleton");
```
ajhalme: Instead of fetching the global singleton and rebuilding the game here, I think we can do this persistence inside `CreateGame()`, after `gameSetup` has been created and before world generation (L226).

The `gameSetup` object should have all the settings needed inside `PersistGameSettings(..)` - and they're public, too.

C7/UIElements/NewGame/QuickStartSetup.cs
```diff
@@ -17,25 +18,26 @@ public static void Init(GlobalSingleton globalState) {
...
+			climate = GetSetting("climate", WorldCharacteristics.Climate.Normal),
+			temperature = GetSetting("temperature", WorldCharacteristics.Temperature.Temperate),
+			worldSize = GetWorldSize(save.WorldSizes),
 			mapSeed = new Random().Next(),
 		};
 
 		globalState.SaveGame = save;
 
-		Civilization player =  save.Civilizations.FirstOrDefault(x => x.name == "Netherlands") ?? save.Civilizations[1];
+		string lastCivilization = C7Settings.GetSettingsValueOrDefault("lastGame", "civilization", "Netherlands");
```
ajhalme: This game hydration data fetching logic is currently a mixture of 1) the new `GetSetting(..)` alone, 2) regular C7Settings fetch + first match, 3) a special "getter" (for worldSize).

Would be nicer, if we followed 3) and had getter function for each thing we are retrieving from the settings file.

C7/UIElements/NewGame/QuickStartSetup.cs
```diff
@@ -46,4 +48,41 @@ public static void Init(GlobalSingleton globalState) {
 
 		gameSetup.Populate(save);
 	}
+
+	private static T GetSetting<T>(string key, T defaultValue) where T : struct, Enum {
```
ajhalme: This could fit nicely inside C7Settings as `GetTypedSetting(..)` or something.

C7/UIElements/NewGame/QuickStartSetup.cs
```diff
@@ -46,4 +48,41 @@ public static void Init(GlobalSingleton globalState) {
...
+
+	private static List<SelectedOpponent> GetOpponents(int expectedCount) {
+		string opsRaw = C7Settings.GetSettingsValueOrDefault("lastGame", "opponents", "");
+		List<SelectedOpponent> opponents = [];
+
+		if (!string.IsNullOrEmpty(opsRaw)) {
+			foreach (string name in opsRaw
+				         .Split(',').Select(n => n.Trim())
+				         .Where(n => !string.IsNullOrEmpty(n))
+			        ) {
```
ajhalme: For readability, this enumerator could be defined outside the loop:

```
var cleanNames = opsRaw.Split(',').Select(n => n.Trim()).Where(n => !string.IsNullOrEmpty(n);
foreach (string name in cleanNames)

```

C7/UIElements/NewGame/PlayerSetup.cs
stavrosfa: I agree with this. I also had in mind to comment on something similar.

There has been some discussion in the past, about essentially having a single format that we save/read settings, for the game, like the units for example. 
So that would mean .ini files at some point would be replaced by json.

We don't need to do this here and now, but what I want to suggest would also help with this.

I would prefer if the C7Settings were not strings, that are duplicated all over the place, and instead be some kind of class/struct. Then for .ini we could also do something like `nameof(settings.fieldName)` if we want to keep the names as they are and we don't want to modify the code drastically.

At the very least, I would create constants out of each value that is referenced more than once.

albertone:
Hi,
Thanks for the feedback and the suggestions! I just pushed a whitespace fix, but I didn't see these comments before. I'll review the PR in the next few days to refactor the code according to your requests. Thanks for the guidance!

albertone:
Hi, following your suggestions I have refactored the settings management to improve stability and readability.

I created the LastGame static class in C7Settings using nameof() for all keys. This eliminates 'magic strings' and facilitates future refactoring.

Now the settings are saved using the GameSetup object just before world generation.

I extracted the data loading logic into specific private methods, leaving the Init method much cleaner and easier to follow.

I refactored the opponent list hydration to validate civilizations against the configuration. I also changed the delimiter to | to avoid conflicts with names containing commas (for example, in a future mod).

C7/UIElements/NewGame/PlayerSetup.cs
ajhalme: The `PersistGameSettings(..)` call site looks good and the function looks reasonable. The `SetValue(..)` call rows do look a bit unwieldy now and the logic relating to opponent collection generation/parsing would ideally be part of the same section of code, but this is fine as is.

I think the changes address concerns by @stavrosfa as well - maybe follow-up with a comment if you think otherwise.

C7/UIElements/NewGame/QuickStartSetup.cs
ajhalme: This new organisation looks fine to me.

Opponent selection doesn't support "None" civs, for games with fewer civs than the map allows, but this isn't supported by the UI either, so this is all fine for now.

C7/UIElements/NewGame/QuickStartSetup.cs
ajhalme: ok

C7/UIElements/NewGame/QuickStartSetup.cs
ajhalme: ok

ajhalme:
LGTM and working fine, thanks

@stavrosfa, any thoughts?

stavrosfa:
Looks good to me too, I didn't test it but since @ajhalme did, I am good on that end.

I only want to comment on the `nameof()` code. When I suggested it, what I had in mind, was to have a (LastGame) object that could be serialized to a json. So the `WorldSize` field for example, would have a value of `gameSetup.worldCharacteristics.worldSize.name`, etc..

Now it's used more like constats for the field names, which is fine, but I wanted to make sure I got my point across.

As I said, looks good, thanks!

ajhalme:
> [..] what I had in mind, was to have a (LastGame) object that could be serialized to a json. So the `WorldSize` field for example, would have a value of `gameSetup.worldCharacteristics.worldSize.name`, etc..

Yeah, a more object-oriented implementation could work here. One could even render both INI settings and this JSON out of the same instance, as needed.

Storing the last game information in the INI file is true to the original, so there's that. The original relies on numerical values instead of strings, but I think one is going to have trouble with the values when modding either way.

In any case, I think this implementation is fine and should prove useful as is. We can always revisit if we want to go for the JSON stuff.

albertone:
Hi @stavrosfa and @ajhalme, thanks for the feedback!


I see your point regarding JSON serialization; it would definitely be a cleaner object-oriented approach for the future. For now, my goal was to maintain the current .ini structure and stay within the scope of the issue.


I'm glad the implementation works for you. I'm open to help on the transition to JSON in a future refactor if the team decides to move away from the `.ini` file for these settings.

Issue #452, opened 2024-10-15
Find a canonical way to block input to the whole screen and/or make blocking modals


WildWeazel:
PopupOverlay.setMouseFilter recurses through the scene to update the mouse filter on every Control node. Investigate recommendations for how to accomplish this, and make it a generic function that can be used anywhere.

Original comment:
              let's make a TODO to look for a pattern for this. there has to be a better way than recursing through the whole scene

_Originally posted by @WildWeazel in https://github.com/C7-Game/Prototype/pull/409#discussion_r1800425049_

Billy-Lasso:
I'd like to take this one. Opened #990: it extracts the recursive mouse-filter logic into a reusable `NodeExtensions.SetMouseFilterRecursive` extension on `Node`, and while testing it I found and fixed a related leak where `TileInfoPopup` subscribed to `PopupOverlay.Click` without unsubscribing (a stale handler could dismiss later modals on an outside click). Both are in the one PR.

Billytifft:
Update: the active PR is now #991 (the earlier #990 was opened from the wrong account and closed).

PR #924, opened 2026-04-06
Add TileInfo

ajhalme:
Adds basic TileInfo popup.

The implementation is based on a new terrain layer that exposes the targeted terrain tile by selectively removing some of the overlays around the target tile. The terrain layer implementation allows for the Tile Info popup to stay in place even if the map is moved, but it does complicated some things, as we want to ultimately draw in the UI layer.

Implementation follows GotoLayer implementation in that we set the target tile object when we right click somewhere in the open map, or select the target from the context menu. In the TileInfo layer this selection is then ignored by all tiles expect the selected one. Presentation and content logic is mostly confined to the new layer, but the "stripping back the layers" implementation means that some of the overlay layers also need to know about the TileInfo context, so they *don't* draw parts that would be exposed inside the TileInfo window.

Trickiest part here is hiding the *city labels* from tiles to the North, NE, or NW of the target tile, as these overlap the target tile. We resort to some gnarly hackery to hide the labels. We want to keep the city (as it shows outside the frame), but hide the label (as it shows up inside).

I'm fairly confident that this approach is aligned with original's, as the size of the TileInfo frame perfectly covers the tiles South, SW, and SE of the target, each of which needs to set to temporarily not draw any units, as these would show up in the Tile Info window.

As a bonus, I added the tile coordinates up top, as I often wanted to know these while debugging. 

**Note**: standalone mode needs the corresponding frame asset.

--
### From the Civ Guide
tile_information_in_guide

### Original game samples (justification issues probably due to my dodgy Steam install)
tile_information_samples

### OpenCiv3
fileinfo_openciv3

### DIFF
--- a/C7/MapView.cs
+++ b/C7/MapView.cs
@@ -515,7 +515,7 @@ public partial class LooseView : Node2D {
 				layer.onEndDraw(this, gD);
 			}
 
-			foreach (GridLayer layer in layers.Where(layer => layer.visible && layer is GridLayer).Cast<GridLayer>()) {
+			foreach (var layer in layers.Where(layer => layer.visible && layer is GridLayer)) {
 				layer.onBeginDraw(this, gD);
 				foreach (VisibleTile vT in visibleKnownTiles.Concat(visibleUnKnownTiles)) {
 					layer.drawObject(this, gD, vT.tile, vT.tileCenter);
@@ -578,6 +578,22 @@ public partial class LooseView : Node2D {
 		}
 		TileKnowledge knowledge = gameData.GetFirstHumanPlayer().tileKnowledge;
 		return tile != Tile.NONE && knowledge.isTileKnown(tile);
+	}
+
+	public bool IsTileCoveredByTileInfo(Tile tile) {
+		var target = mapView.game.tileInfo;
+		if (target == null) return false;
+
+		return target.IsCovered(tile);
+	}
+
+	public bool HasCityLabelToHideFromTileInfo(Tile tile) {
+		if (!tile.HasCity) return false;
+
+		var target = mapView.game.tileInfo;
+		if (target == null) return false;
+
+		return target.HasCityLabelToCover(tile);
 	}
 }
 
--- a/C7/UIElements/Popups/PopupOverlay.cs
+++ b/C7/UIElements/Popups/PopupOverlay.cs
@@ -13,6 +13,7 @@ public partial class PopupOverlay : HBoxContainer {
 	[Signal] public delegate void BuildCityEventHandler(string name);
 	[Signal] public delegate void DiplomacySelectionEventHandler(ParameterWrapper<ID> opponentPlayer);
 	[Signal] public delegate void HidePopupEventHandler();
+	[Signal] public delegate void ClickEventHandler();
 
 	Control currentChild = null;
 
@@ -22,21 +23,17 @@ public partial class PopupOverlay : HBoxContainer {
 	public enum PopupCategory {
 		Advisor,
 		Console,
-		Info    //Sounds similar to the above, but lower-pitched in the second half
+		Info,    //Sounds similar to the above, but lower-pitched in the second half
+		TileInfo
 	}
 
 	public void OnHidePopup() {
-		// 1. enable mouse interaction with non-UI nodes
-		MouseFilter = MouseFilterEnum.Pass;
-		RemoveChild(currentChild);
-		currentChild = null;
+		Reconnect();
+		if (currentChild != null) {
+			RemoveChild(currentChild);
+			currentChild = null;
+		}
 		Hide();
-
-		// 2. enable mouse interactions with other UI elements
-		setMouseFilter(control, MouseFilterEnum.Pass);
-
-		// 3. enable clicking other UI elements
-		control.ProcessMode = ProcessModeEnum.Inherit;
 	}
 
 	public bool ShowingPopup => currentChild is not null;
@@ -45,15 +42,6 @@ public partial class PopupOverlay : HBoxContainer {
 		AudioStreamPlayer player = GetNode<AudioStreamPlayer>("PopupSound");
 		player.Stream = wav;
 		player.Play();
-	}
-
-	private void setMouseFilter(Node n, MouseFilterEnum filter) {
-		foreach (Node child in n?.GetChildren()) {
-			setMouseFilter(child, filter);
-		}
-		if (n is Control control) {
-			control.MouseFilter = filter;
-		}
 	}
 
 	public void ShowPopup(Popup child, PopupCategory category) {
@@ -72,29 +60,69 @@ public partial class PopupOverlay : HBoxContainer {
 		AddChild(child);
 		currentChild = child;
 
-		string soundFile = category switch {
+		var soundFile = category switch {
 			PopupCategory.Advisor => "Sounds/PopupAdvisor.wav",
 			PopupCategory.Console => "Sounds/PopupConsole.wav",
 			PopupCategory.Info => "Sounds/PopupInfo.wav",
-			_ => "",
+			_ => null
 		};
-		if (soundFile == "") {
-			log.Error("Invalid popup category");
-		}
-		AudioStreamWav wav = Util.LoadCiv3WAVFromDisk(soundFile);
 
-		// 1. prevent mouse interaction with non-UI elements (ie. the map)
-		MouseFilter = MouseFilterEnum.Stop;
+		var wav = soundFile == null ? null : Util.LoadCiv3WAVFromDisk(soundFile);
 
-		// 2. prevent clicking other UI elements
-		control.ProcessMode = ProcessModeEnum.Disabled;
-
-		// 3. ignore all mouse input on other UI elements (ie. button color changes on hover)
-		setMouseFilter(control, MouseFilterEnum.Ignore);
+		Isolate();
 
 		Show();
+
 		if (wav != null) {
 			PlaySound(wav);
+		}
+	}
+
+	/// <summary>
+	/// Creates a modal context by preventing UI events outside the popup context.
+	/// Inverse of `Reconnect(..)`.
+	/// </summary>
+	private void Isolate() {
+		// 1. Overlay catches mouse events, prevents event propagation
+		MouseFilter = MouseFilterEnum.Stop;
+
+		// 2. Stop the world: switch off UI elements
+		control.ProcessMode = ProcessModeEnum.Disabled;
+
+		// 3. Ignore all mouse input on UI elements
+		SetMouseFilter(control, MouseFilterEnum.Ignore);
+	}
+
+	/// <summary>
+	/// Unwind a modal context by allowing UI events outside the popup context.
+	/// Inverse of `Isolate(..)`.
+	/// </summary>
+	private void Reconnect() {
+		// 1. Let events propagate past the overlay
+		control.MouseFilter = MouseFilterEnum.Pass;
+
+		// 2. Let UI elements catch mouse inputs again
+		SetMouseFilter(control, MouseFilterEnum.Pass);
+
+		// 3. Restart the world: let UI elements run normal
+		control.ProcessMode = ProcessModeEnum.Inherit;
+	}
+
+	/// Recursively set MouseFilter on node children and their children, etc.
+	private static void SetMouseFilter(Node n, MouseFilterEnum filter) {
+		foreach (var child in n?.GetChildren() ?? []) {
+			SetMouseFilter(child, filter);
+		}
+		if (n is Control control) {
+			control.MouseFilter = filter;
+		}
+	}
+
+	public override void _GuiInput(InputEvent @event) {
+		// Raise an event when popup overlay catches a click that the popup itself misses.
+		// This usually means a click outside the popup modal.
+		if (Visible && @event is InputEventMouseButton ev && ev.Pressed) {
+			EmitSignal(SignalName.Click);
 		}
 	}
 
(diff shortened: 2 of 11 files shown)


ajhalme:
Added a basic standalone asset for the tile info box: https://github.com/C7-Game/Assets/pull/5

C7/Map/TileInfoLayer.cs
```diff
@@ -0,0 +1,185 @@
...
+			terrainLabel.Text = "";
+			terrainOverlayLabel.Text = "No information available";
+			resourceLabel.Text = "";
+			foodLabel.Text = "";
+			shieldsLabel.Text = "";
+			goldLabel.Text = "";
+		}
+
+		// Extra label
+		var extraPosition = boxTextureRect.Position + new Vector2(boxTextureRect.Texture.GetWidth() / 2f - 30, 0);
```
stavrosfa: I don't think this should be in the main game, a casual player doesn't know/care what tile this is.

If the need for this to exist is to help with debugging, there is already a method that logs (debug) the tile coordinates on right click.
https://github.com/C7-Game/OpenCiv3/blob/d298f23652cfd76ac3d034ddb3b69d0304196fb9/C7/Game.cs#L557

We also have the observer mode toggle coordinates action for showing tile coordinates.

C7/Map/TileInfoLayer.cs
```diff
@@ -0,0 +1,185 @@
...
+
+		_labels.AddRange([extraLabel, terrainLabel, terrainOverlayLabel,
+			resourceLabel, foodLabel, shieldsLabel, goldLabel]);
+	}
+
+	private void SetupFont() {
+		font = ResourceLoader.Load<FontFile>("res://Fonts/NotoSans-Regular.ttf", null, ResourceLoader.CacheMode.Ignore);
+		font.FixedSize = 20;
+
+		whiteFontTheme.DefaultFont = font;
```
stavrosfa: Is this actually used anywhere? Just by looking at it at the tile info graphics, I don't see a white font being used.

C7/Map/TileInfoLayer.cs
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+
+	private void SetupBoxTexture() {
+		ImageTexture boxTexture = TextureLoader.Load("tileinfo.box");
+		boxTextureRect = new TextureRect();
+		boxTextureRect.Texture = boxTexture;
+	}
+
+	private void AddCloseButton() {
```
stavrosfa: The button doesn't change textures when I hover/click. It shows the same texture at all times.

C7/Map/TileInfoLayer.cs
```diff
@@ -0,0 +1,185 @@
...
+		boxTextureRect.MouseFilter = Control.MouseFilterEnum.Ignore;
+		closeButton.MouseFilter = Control.MouseFilterEnum.Pass;
+
+		_labels.AddRange([extraLabel, terrainLabel, terrainOverlayLabel,
+			resourceLabel, foodLabel, shieldsLabel, goldLabel]);
+	}
+
+	private void SetupFont() {
+		font = ResourceLoader.Load<FontFile>("res://Fonts/NotoSans-Regular.ttf", null, ResourceLoader.CacheMode.Ignore);
+		font.FixedSize = 20;
```
stavrosfa: Perhaps not the right place, but could we make the font just a wee bit smaller? It seems kind of off to me..

C7/MapView.cs
```diff
@@ -579,6 +579,28 @@ private static bool IsTileKnown(Tile tile, GameData gameData) {
 		TileKnowledge knowledge = gameData.GetFirstHumanPlayer().tileKnowledge;
 		return tile != Tile.NONE && knowledge.isTileKnown(tile);
 	}
+
+	public bool IsTileCoveredByTileInfo(Tile tile) {
```
stavrosfa, line 583: Not sure this is the best approach. We have to do this for every tile, for multiple layers, multiple times a second, even when we are not actually drawing anything.

stavrosfa (commented):
Thanks!

Except what I already commented on in the code, I have some extra comments.

We should be able to hide the tile info when left-clicking outside the texture. We should also be able to hide it when Escape is pressed. Right now if I press escape this happens

Image

I was also able to move a unit while having this open. I think we should either hide the graphic upon performing some action, or disable other actions altogether (except closing the actual window of course).

I like that now when showing the tile info on a city tile, we can finally see the actual terrain and not an empty city, which was not really helpful. We can add some graphic in the future showing that its a city tile, but it's fine as is for now.

But I just realized, we also can't see any improvements like mines or roads on tiles.

When we inspect a tile that is known to us, but currently not visible, the tile still shows with that half-fog of war overlay instead of being "lit" as if it's visible, like in the original game.

Notes aside, what I think the original is doing, is not hiding and showing "layers", units,  etc, but rather, draw on top of what is already being shown, the current tile info.
But we will never know 😅

C7/Map/TileInfoLayer.cs
ajhalme: Removed theme usage, wasn't needed.

C7/Map/TileInfoLayer.cs
ajhalme: Adjusted to 12, so should now match the lower right box font (at default zoom).

C7/Map/TileInfoLayer.cs
ajhalme: I think this is because the box is still under the overall UI Control layer - the hover event doesn't travel through. Or could also be the layer is redrawing so quickly that the hover is effectively ignored.

Need to investigate a bit more.

C7/Map/TileInfoLayer.cs
ajhalme: Removed the coordinates from the infobox. Having the coords available in observer mode is plenty for what I had in mind -- I didn't realise the coord grid was available there.

That said, I think it might be reasonable to show more things in the info box, like something combat or traversal related, or player-specific values.

C7/MapView.cs
ajhalme: I guess the question is: do we want the infobox to draw as a map element (sticks to tile) or as a UI element (floats when panning the map).

If we want the infobox in the terrain layers, I don't see selective rendering happening without a check like this.

I've simplified and performance boosted this a bit by precomputing the covered sets of tiles at TileInfo init, meaning this lookup reduces to little more than a peek at the coverage hash set.

ajhalme:
> We should be able to hide the tile info when left-clicking outside the texture. We should also be able to hide it when Escape is pressed. [..]
I was also able to move a unit while having this open. I think we should either hide the graphic upon performing some action, or disable other actions altogether (except closing the actual window of course).

I agree with these, though this is starting to sound like #452 ...

The main question probably is: **Should TileInfo be part of the terrain layers (and fixed to the map) or part of the UI (floating when panning the map)?** The latter doesn't feel right to me, but having a UI element in the terrain makes the whole render sequence more complicated.

Are you sure we want to show improvements? I thought they are one of the layers peeled away so the terrain can shine through. E.g., the horse here is obscured by the mine:

Screenshot at 2026-04-07 23-52-27

Half-fog shouldn't be too hard to remove. I'll look into it.

C7/MapView.cs
stavrosfa: > I guess the question is: do we want the infobox to draw as a map element (sticks to tile) or as a UI element (floats when panning the map).

Maybe if hide the tile info graphic when cliking away or try to move the map, this won't be an issue?

> I've simplified and performance boosted this a bit by precomputing the covered sets of tiles at TileInfo init, meaning this lookup reduces to little more than a peek at the coverage hash set.

My idea for this was to basically split it into two parts, one where we check if (pseudocode) `tileInfoGraphic.isShown()` and then if so, do the computation if necessary, but if not return early, while checking only 1 simple boolean, if that makes sense.
I see you went a different way, I am going to check this probably tomorrow more thoroughly to understand what's happening.

stavrosfa:
I wasn't aware of the issue linked, but sounds very reasonable.

> Are you sure we want to show improvements? I thought they are one of the layers peeled away so the terrain can shine through. 

Not 100% sure which I like more tbh, I was commenting more in line with what the original does.

> E.g., the horse here is obscured by the mine:

In the screenshot the opposite is happening, not sure what you mean here

> Half-fog shouldn't be too hard to remove. I'll look into it.

This falls under "famous last words" hahaha
I hope it treats you better than it has treated me in past my friend 😄

C7/Map/TileInfoLayer.cs
ajhalme: Switched to a popup implementation, the button works as expected now.

Could probably have reused the Popup class's CancelButton, but didn't want to mess with its wiring.

C7/MapView.cs
ajhalme: Switched to a popup implementation. Implementing zoom was a bit of a pain, but it works okay.

Made it so clicking away should close the popup, and of course with a popup we are not constantly rendering it as a layer, so the other issue goes away. Some layers are still impacted by tileInfo selection.

ajhalme:
Popup based implementation. Now with tile improvements shown, and the half-fog.

The popup mechanism comes with its own control and event screening, so issues relating to interactions while the popup is open should go away. Esc-key works as well.

It seems one can now click through the minimap to trigger TileInfo, so that needs another look.

The way the popup scheme currently sets up the isolation might be a bit off as well, though I might well be misunderstanding what it's actually doing. I've tried to extract the salient parts out of the main popup class.

C7/UIElements/Popups/TileInfoPopup.cs
```diff
@@ -0,0 +1,162 @@
...
+	public override void _GuiInput(InputEvent @event) {
+		// ignore clicks inside box (except button)
+		if (Visible && @event is InputEventMouseButton ev && ev.Pressed) {
+			GetViewport().SetInputAsHandled();
+		}
+	}
+
+	public override void _UnhandledInput(InputEvent @event) {
+		// close on clicks outside box
+		if (Visible && @event is InputEventMouseButton ev && ev.Pressed) {
```
stavrosfa: I don't think that's working for either click.

C7/UIElements/Popups/TileInfoPopup.cs
```diff
@@ -0,0 +1,162 @@
...
+	private void InitCloseButton() {
+		ImageTexture xTexture = TextureLoader.Load("ui.cancel.normal");
+		ImageTexture xHover = TextureLoader.Load("ui.cancel.hover");
+		ImageTexture xPressed = TextureLoader.Load("ui.cancel.pressed");
+
+		closeButton.TextureNormal = xTexture;
+		closeButton.TextureHover = xHover;
+		closeButton.TexturePressed = xPressed;
+
+		closeButton.ActionMode = BaseButton.ActionModeEnum.Press;
```
stavrosfa: (Nit) Shouldn't we be closing the popup on release instead?

C7/UIElements/Popups/PopupOverlay.cs
```diff
@@ -58,6 +59,7 @@ public void ShowPopup(Popup child, PopupCategory category) {
 		currentChild = child;
 
 		string soundFile = category switch {
+			PopupCategory.Quiet => "Sounds/Check.wav",
```
stavrosfa: Is this is the original?
Also, what would the `quiet` category contain? What's its purpose? Can you leave a comment explaing it a bit?

stavrosfa (commented):
Thanks, looks good for the most part!

C7/UIElements/Popups/TileInfoPopup.cs
ajhalme: Changed

C7/UIElements/Popups/TileInfoPopup.cs
ajhalme: Yep, removed as no longer needed

C7/UIElements/Popups/PopupOverlay.cs
ajhalme: Remove. I don't think there's any sound for TileInfo, just slapped something there as the TileInfo popup needed some category and each category was required to define an associated sound. (Apparently there's a MiniMap sound, which I also had mapped here...)

Decided to just adjust the code so a popup needn't have any sound.

ajhalme:
Current implementation does all I want it to do, but I've been having a bad time trying to make the MiniMap and bottom right info box ignore the right click. Either I'm misunderstanding how event propagation and handling is supposed to work or there's some Godot weirdness at play.

The original game has an elegant solution to the problem of trying to draw TileInfo for a tile at the bottom edge of the screen: the TileInfo modal is drawn _out of position_, without moving the map. This implementation I've cobbled together doesn't allow this, as it's based this "stripping back the layers" approach. A more robust approach would require changing tile rendering so that individual tiles could be rendered effectively in isolation. I actually tried to do this, but there was a bit too much to adapt in the layer scheme to make it worthwhile. 

I've been considering an approach where the screen centered on the target tile, if one tried to get TileInfo on a tile close to the bottom of the screen / close to the UI elements. This could be a nice workaround. But having lookead at this for two weeks, I'm kinda keen to move on to something else. 

Anyway, I've force rebased the branch to a single commit to hide my earlier crimes with layers, etc. Some cleanup and reorg. Good to go, I reckon, if we can live with the right click glitches on the UI elements. Fixes welcome, if somebody has a better handle on Godot mouse events.

ajhalme:
Right, I think I got it. I didn't realise that there's _two_ MouseEvents from a right click: the press and the release. I was blocking the first ones, but you need to catch both if you want to prevent events reaching the Game Node in the tree. (I was losing my mind thinking the second event that was making it through was the same one as the one I was catching in the code...)

Anyway, this should fix the right-click-on-minimap issue. The patch is a bit flimsy, relying on hovered Control detection, but I think it's good enough.

This is good to go now, the TileInfo now does everything I wanted it to do.

stavrosfa (approved):
This is looking great, much more smooth and intuitive experience, thanks!

PR #926, opened 2026-04-23
Amb Reader

stavrosfa:
This will allow us to read .amb files.

It's by no means complete or fully tested, and could change in the future according to our needs.
It has worked well with what I have tested it so far, including with some .amb files that are known to be corrupted, I just haven't tested it with every .amb available to be sure of all the corner cases.

The `Sfx` struct is meant to be what we will be using as a sound effect data structure from now on, so even regular .wav files would have this format. This of course is also subject to change according to our needs.

I have included some useful links in the code if you want more info on parsing .amb and midi files.

### DIFF
--- a/ConvertCiv3Media/ReadAmb.cs
+++ b/ConvertCiv3Media/ReadAmb.cs
@@ -1,0 +1,653 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Serilog;
+
+namespace ConvertCiv3Media;
+
+public struct Sfx {
+	public float delayStart; // seconds before start playing
+							 // public float duration; // do we even need this?
+	public string wavName;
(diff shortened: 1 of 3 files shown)


ajhalme:
How are you testing the reader? Can you include some of the wiring or a manual test case?

stavrosfa:
Yeah, I probably should have included something simple in the first place. I have modified the `playSound()` in the `AnimationManager` class to play all the wavs at the proper time, with pitch and volume variations, so that's how I was testing it.
I didn't want to include this modified method in this PR so it can stay as clean as possible.

Now, I see the tests are failing because we are on github and there is no civ3 installation, so I will use what we currently use in the other tests to block it here.

ajhalme (approved):
LGTM

Issue #927, opened 2026-04-23
Bombard Mechanic


ajhalme:
_I figured I'd start an issue on bombardment just to gather information in one place before I get too deep into implementing things._

_Some pickings from the official guide._

----

## Overview

When you give a unit the order to Bombard, your cursor changes to cross-hairs. Use the cross-hairs to select the target of the bombardment. (If the cursor becomes anything other than cross-hairs, you’ve moved it outside the effective range of the bombardment.) Only certain units have this ability; check the Civilopedia entry for a unit if you suspect it might be capable of bombardment.

Bombard ability: Warships, bombers, fighters, and artillery units have the ability to bombard a target that’s within their range. Bombardment counts as a unit’s attack, and might damage defensive fortifications, harm units, or otherwise damage a city in the target square.

### Bombardment 

Artillery units (Catapults, Cannons, and all Artillery units) and warships (Frigate, Man-o-War, Ironclad, Destroyer, Battleship, Aegis Cruiser) have the ability to bombard a target that’s within their range.

Bombardment is an attack that does not involve moving into the same square as the defender. It’s a “stand-off” or “ranged” attack. The attacker takes no risk of damage.

Bombardment affects everything in the target square, not just enemy units. The projectiles you launch might damage defensive fortifications like Fortresses and city Walls, harm military units, destroy a portion of a city’s population, or demolish city improvements.

Note that the bombing attacks of fighters and bombers work in much the same way as this type of bombardment.

### Nulear Attacks

You launch a nuclear attack in the same way you target an air unit bombing mission.

All units in the target square and adjacent squares have only a 50% chance of surviving, regardless of their cultural allegiance (in other words, both theirs and yours). In addition, a bombed city loses half its population. The defense against most nuclear attacks is the Small Wonder SDI Defense.

### City Governor

[City governors can be set to] "Artillery – strictly offensive bombardment units, like Catapults".

### Terrain Damage

Terrain bombardment can cause craters that can be cleaned by Workers. Pressing [Shift]-[D] will automate a unit to clear damage.

### Air Mission: Bombing mission

Bombing mission ([B]): Drop bombs on the selected terrain square or enemy city. Air bombardment affects city improvements and city populations.

### Air Mission: Precision bombing

Precision bombing: Drop bombs that target improvements. This mission is only available if your civilization has researched the Smart Weapons advance for Stealth Bombers and Stealth Fighters.

### Bombard ([B])
[...]

PR #928, opened 2026-04-29
Add basic bombard by land units

ajhalme:
Basic bombard mechanic based on #927.

## Changes
- `base-ruleset.json` extension with `bombardRange` and `rateofFire` for all units
- `unit_bombard` unit button enabled, including "B" shortcut key 
- Texture based crosshair & disallowed cursor (Godot native)
- Mouse-based interactive target selection
- `BombardLayer` to render bombard grid, cursor, target selection
- `BombardInfo`, in the spirit of GotoInfo, to carry information about user inputs to BombardLayer
- Fixed/re-enabled "effect" animations for the bombard explosion
- Extend `bombard(..)` MapUnit logic with rate of fire (and hit notifications)
- Bombard target tile collection enumerator in `Tile`, `GetTilesWithinBombardRange(..)` -- Rank-based enumerator ignores the "corners" that are needed for the bombard grid. (Think "Rank-2 City BFC" vs "complete 5x5 tile square".)

## Extras
- If a saved game is set up such that an opponent civ is about to attack, CreateGame will get stuck in an infinite loop, waiting for animations. This PR includes a workaround that disabled animations on CreateGame load.

## Demo
artillery_oc3


ajhalme:
Known issues (Update: Resolved!):
- [x] Bombard before declaring war
- [x] Bombing your own units
- [x] Lethal bombard

C7Engine/EntryPoints/MessageToEngine.cs
```diff
@@ -78,6 +78,27 @@ public override async void process() {
...
+
+		public MsgBombard(ID unitID, Tile tile) {
+			this.unitID = unitID;
+			this.tileX = tile.XCoordinate;
+			this.tileY = tile.YCoordinate;
+		}
+
+		public override async void process() {
+			MapUnit unit = EngineStorage.gameData.GetUnit(unitID);
+			Tile tile = EngineStorage.gameData.map.tileAt(tileX, tileY);
```
stavrosfa, line 95: What is the purpose of passing the Tile in the constructor, extracting the X,Y coords, and the get the Tile itself from the coords? Why not use the Tile/TileId directly?

C7Engine/C7GameData/Tile.cs
```diff
@@ -702,6 +702,17 @@ public List<Tile> GetTilesWithinRankDistance(int rank) {
 			return result;
 		}
 
+		// Same as GetTilesWithinRankDistance, but includes "corner tiles",
+		// i.e., returns perfect tile squares. 
+		public List<Tile> GetTilesWithinBombardRange(int range) {
```
stavrosfa: This method could be used for other things (e.x. settler square preview), so I would suggest a more generic name for it

C7/Map/BombardLayer.cs
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	public override void drawObject(LooseView looseView, GameData gameData, Tile tile, Vector2 tileCenter) {
+		var bombardInfo = looseView.mapView.game.bombardInfo;
+		if (bombardInfo == null || bombardInfo.bombardingUnit.location != tile)
+			return;
+
+		var unit = bombardInfo.bombardingUnit;
+		var range = unit.unitType.bombardRange;
+		var reachableTiles = tile.GetTilesWithinBombardRange(range);
```
stavrosfa: What do you think about adding some kind of caching here? I mean the tiles are not going to change, and once you have calculated once which tiles are are within the range of Tile at X,Y, you always end up with the same result, right?
The key could be the tileid/x-y coords + the range as a string or something

C7/Animations/AnimationManager.cs
```diff
@@ -264,6 +299,7 @@ public C7Animation(AnimationManager civ3AnimData, AnimatedEffect effect) {
 		this.folderPath = "Art/Animations/" + effectCategories[effect];
 		this.iniFileName = effectINIFileNames[effect];
 		this.action = MapUnit.AnimatedAction.DEATH;
+		this.effect = effect;
```
stavrosfa, line 302: Not a comment on this particular PR, but since you had some friction with it now, does this whole system look ok to you (AnimationManager, C7Animation, AnimationTracker, etc)?
Let me know if you have any comments, I am trying to refactor this on the side, make it more versatile and extendible

C7/Map/UnitLayer.cs
```diff
@@ -162,13 +162,19 @@ private Vector2 GetFramePosition(MapUnit.Appearance appearance, C7Animation unit
 	}
 
 	public void drawEffectAnimFrame(LooseView looseView, C7Animation anim, float progress, Vector2 tileCenter) {
```
stavrosfa, line 164: Another question rather than critique per se, do you think the effects animations in general should live in the Unit Layer? What if we had a dedicated EffectsLayer?

stavrosfa (approved):
Good stuff, thanks!

I didn't playtest this yet, but I will and let you know if I have any more thoughts

C7Engine/EntryPoints/CreateGame.cs
```diff
@@ -41,10 +41,14 @@ public static async Task<Player> createGame(SaveGame save, CreateGameParams opti
 			false => throw new Exception($"The provided save does not contain a human player"),
 		};
 
+		EngineStorage.animationsEnabled = false;
```
stavrosfa: Sorry, I forgot to adress this in the previous review.

This looks inoccent enough, although I am pretty sure it's gonna have side effects. I need to investigate further.
If you have more detailed clues as to why the game gets stuck in an infinite loop, let me know.

C7Engine/EntryPoints/MessageToEngine.cs
ajhalme: Mainly just trying to follow the existing message patterns here.

This is the UI--Engine boundary, so having the message be easy to serialise is probably a good idea. If the message can be fully described with primitive values and then hydrated to the actual game object when processed, that's probably a win for portability, multiplayer, etc. Tile Id could work here as well, I haven't really looked at the usage patterns, but coordinates at least should be fairly stable in any game.

C7Engine/C7GameData/Tile.cs
ajhalme: Sure. Changed to `GetTilesWithinTileSquare(..)`, which isn't the greatest, but maybe a bit more versatile.

C7/Map/BombardLayer.cs
ajhalme: Added a simple dictionary cache in the layer using `<TileID, range>` as key.

C7/Animations/AnimationManager.cs
ajhalme: Current effect animation feels very glued-on on top of the unit animation scheme. Rethinking/refactoring the animation system as a whole makes sense.

The bomber plane bombardment animation might be a good starting point as it's a complex sequence. If you can support that, then everything else is easy.

I kinda like the approach of having the animation scheduling and progress-dependent rendering be separate parts. It's a bit much initially, but I can see how it can be really flexible. Maybe we need a central scheduling scheme and then a few different renderers. Could have simpler tile-based ones for unit and effect, and then a more complex renderer that can handle moving the render target on screen.

C7/Map/UnitLayer.cs
ajhalme: A separate EffectLayer could work fine. The effects generally are unit-linked so they _can_ in my view live alongside the units, but a separate layer could make it easier to, well, layer. The z-index hackery doesn't scale.

C7Engine/EntryPoints/CreateGame.cs
ajhalme: I tracked the issue down to the `fight(..)` method in `MapUnit`. The comment "Do combat rounds" (line 355) is at the start of a loop that never resolves with the attached save file. It's not actually looping, I think the animation system just isn't ready to receive the animation render requests for the AI civ's fight animation at game load.

Indeed, my save files load differently in OC3 and Civ3, as the same save file in the former starts with AI civs making their moves, while latter starts with human user's (previous?) turn ended.

Attached zipped save file reproduces the issue, getting stuck on `await attacker.animateAsync(ATTACK1);` in `fight(..)` (L358 in my codebase).

[on_load_animation_bug.json.zip](https://github.com/user-attachments/files/27265150/on_load_animation_bug.json.zip)

ajhalme:
Added basic improvements bombardment and even city and population bombardment. Had a stab at introducing city walls logic as well.

Added some basic wiring for requiring war declaration, disallowing friendly fire, disallowing bombarding of in-range empty tiles, etc.

Merged in the recent changes, plus the not yet merged test file change.

Good to go - bombard has everything I wanted for now. Didn't explore how things break with naval vessels and air things, but that's for another PR, I reckon.

C7Engine/C7GameData/City.cs
```diff
@@ -629,6 +629,16 @@ private void RemoveCitizen() {
 			residents.RemoveAt(residents.Count - 1);
 		}
 
+		public void RemoveRandomCitizen() {
```
stavrosfa, line 632: I would seperate the picking of the random citizen and their removal, and since there already is a method that handles the citizen removal, I would unify them into something like this

```
private void RemoveCitizen(int index = -1) {
    if (index == -1) 
        index = residents.Count - 1;
	residents[index].tileWorked.personWorkingTile = null;
	residents.RemoveAt(index);
}

public void RemoveRandomCitizen() {
	if (residents.Count == 1)
		return; // TODO: Handle extreme case

	var idx = GameData.rng.Next(residents.Count);
    RemoveCitizen(idx);
}
```

because, if we need to extend what a citizen removal does (e.x. emit an event, trigger a calculation, give/take gold, etc), we would have to modify it in 2 places

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -418,36 +418,217 @@ public async Task<CombatResult> fight(MapUnit defender) {
...
+
+		public bool canBombardTile(Tile tile) {
+			if (unitType.bombard == 0)
+				return false;
+
+			if (tile.HasImprovements)
+				return true;
+
+			MapUnit target = tile.FindTopDefender(this);
+			if (target?.owner == owner)
```
stavrosfa: I have to ask, why the null check? This potentially hides a bug by silently failing, and from what I can see in the FindTopDefender() method, it can't return null anyway, it will either return a MapUnit.NONE or a valid one.

C7Engine/C7GameData/Tile.cs
```diff
@@ -840,6 +855,11 @@ public void Add(TerrainImprovement improvement) {
 			}
 		}
 
+		public void Remove(TerrainImprovement improvement) {
```
stavrosfa, line 866: I don't this is as simple as that. This has to mirror adding a terrain improvement, and for example invalidate the trade network if the improvement was a road.
I would extract the invalidation part to a seperate method, and call it from both Add and Remove, just to be able to reuse it, just because I don't know what else might end up in there.

C7Engine/EntryPoints/CreateGame.cs
stavrosfa: I still don't get this fully. Does the saved game begin with the AI attacking? If so, it shouldn't be possible. It shouldn't be possible to save the game while someone else is moving in the first place.

If the game was saved at the very end of the round, where you have to press the button/Enter to finish the round, then loading it up begins the round without waiting for the human to trigger it, it's faulty behaviour.

Regardless, I don't think disabling all animations is a valid workaround, there is clearly something else going on here, and we should address this instead.

Also, as far as I understand, this is not bombard related, right? Any attack triggers this

C7Engine/EntryPoints/MessageToEngine.cs
stavrosfa: Sorry, which pattern are you referring to here? I am not seeing anything similar in the other messages.

Messages that can be serialized, is something that I haven't though of directly for this game.

The pattern so far seems to be to use DTOs instead of directly serializing an object, so I would guess that's how it would be for messages too, as they handle City, TilePath or MapUnit objects at the moment directly.

Anyway, I don't have a problem with this, if it were me, I would simplify it as I mentioned, but I don't mind this.

C7/Map/BombardLayer.cs
```diff
@@ -0,0 +1,100 @@
...
+	private void drawBombardTile(LooseView looseView, Vector2 tileCenter) {
+		var cS = MapView.cellSize;
+		var left = tileCenter + new Vector2(-cS.X, 0);
+		var top = tileCenter + new Vector2(0, -cS.Y);
+		var right = tileCenter + new Vector2(cS.X, 0);
+		var bottom = tileCenter + new Vector2(0, cS.Y);
+		DrawSquare(looseView, left, top, right, bottom);
+	}
+
+	private void drawTargetBombardTile(LooseView looseView, Vector2 tileCenter) {
```
stavrosfa, line 84: I believe it's the same thin brown outline pattern for the target square as in here:

https://github.com/C7-Game/OpenCiv3/blob/b8a7c41d2cb6f2ce44665251be3a58a9dbc0288b/C7/Map/TileAssignmentLayer.cs#L122

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -418,36 +418,217 @@ public async Task<CombatResult> fight(MapUnit defender) {
...
-			if (target.hitPointsRemaining <= 0) {
-				RollToPromote(target);
-				await target.animateAsync(MapUnit.AnimatedAction.DEATH);
-				target.RemoveFromPlay();
+		private async Task bombardCity(Tile tile) {
+			// Anecdotal: If there are no units left to hit, then citizens or buildings are hit, apparently with same probability.
+			// Anecdotal: "buildings (if I remember correctly) have a defense value of 16"
+			// Anecdotal: It seems population is killed off more quickly than buildings.
+
+			const int buildingDefence = 16;
```
stavrosfa, line 576: It would be a nice opportunity to expose and make these consts configurable, either in Lua(that would be my preference as you can also add behaviour) for example, or as a Rule.

(EDIT) This could be a TODO, not necessarily done now

C7Engine/EntryPoints/MessageToEngine.cs
ajhalme: Many of the engine messages in this file (`MsgSetUnitPath`, `MsgMoveUnit`, `MsgSetFortification`, ...) feature a constructor that takes a UnitId from a MapUnit. This id is then hydrated into the MapUnit inside `process(..)`. Why do these messages bother with an id instead of the full MapUnit?

This is the pattern I'm building on:
```
UnitId --> MapUnit
TileId/Coords --> Tile
```

Any thoughts on engine message data fields vs full DTOs, @esbudylin ?

C7Engine/EntryPoints/CreateGame.cs
ajhalme: I've removed this workaround. The load issue is not bombard-specific.

But yes, I observed the game loading with AI civ units making moves, including combat, which triggers this bug. I've been able to replicate this with both normal saves and ones I modified by giving the AI civs more units. (I was actually trying to replicate the city crashing issue you mentioned in issue 907 by having all the civs at war with each other: this mayhem led me to this issue initially.)

Anyway, I've created a new save file that reliably results in an all-black map for me:
[on_load_animation_bug_2.json.zip](https://github.com/user-attachments/files/27417937/on_load_animation_bug_2.json.zip)

C7Engine/C7GameData/City.cs
ajhalme: Cleaned this up with `RemoveLastCitizen`, `RemoveRandomCitizen`, and `RemoveCitizenAt`.

`RemoveCitizenAt` can do all the heavy lifting, as needed.

C7Engine/C7GameData/MapUnit.cs
ajhalme: Added TODOs -- this PR has enough stuff going on as is.

C7Engine/C7GameData/MapUnit.cs
ajhalme: Removed null conditional.

C7/Map/BombardLayer.cs
ajhalme: Yes, the mechanism looks similar - but just different enough that making one work like the other is a bit involved. Something to look into for consolidation/refactoring.

I'm also inclined to break up the monolithic `MapView.cs` file, bring LooseLayer closer to the usage. Maybe then we can have a layer utilities library where we can put some of these reusable pieces.

C7Engine/C7GameData/Tile.cs
ajhalme: I've revised this so both Add and Remove call `ApplyTerrainImprovementChange`.

C7Engine/EntryPoints/MessageToEngine.cs
esbudylin: Thanks for the ping! I'll take a look on the weekend.

C7Engine/EntryPoints/CreateGame.cs
ajhalme: I'm now thinking the animation / black screen issue probably stems from `Player` property `hasPlayedThisTurn`/`hasPlayedCurrentTurn` and the handling of it in the save state.

Either the human player has this property `true` when it shouldn't be, or the AI civs have it `false` when it shouldn't be. I'm not certain which way around it is or should be. Maybe active human should always be `hasPlayedThisTurn=false` at save time, or AI civs always `true`.

stavrosfa: It's not that, I tested it with your save. I have a workaround, but I feel like the issue is even deeper. Perhaps I can open a draft PR and we can discuss there?

C7/Map/BombardLayer.cs
stavrosfa: Sure, if you want to suggest something in the form of code, open a draft PR and we can discuss it

C7/Animations/AnimationManager.cs
stavrosfa: I think we need to move away from tile-based animations altogether, and move to X/Y (screen?) coordinate animations. Tile based anything seems to be very restrictive as it is right now.
Stuff that are not doable with the current setup:
1. Move unit to a tile, without hiding the units already on said tile, untill the unit has actually finished moving.
2. Move half a tile to initiate a fight
3. Move half a tile when retreating (basiaclly the same as above, on the opposite direction)
4. Probably all Bomber-like animations

stavrosfa (commented):
I don't have any more comments on this I think, we covered this PR and some extra stuff.

The only thing that seems to be missing, is the graphics for the cursor for the standalone mode, that crashes the game in the beggining. I also need to merge the unit icons PR I have opened, but this one at least doesn't crash the game.

I have asked on Discord as well if anyone is interested in creating a few graphics for us

ajhalme:
Adding Cursor.png here:
https://github.com/C7-Game/Assets/pull/6

Also included the file in `C7.lua`.

stavrosfa:
Great, feel free to merge if you feel this is done

PR #933, opened 2026-05-06
Support broadside attack facing

### LINKED ISSUE #902: Galleys don't face off correctly during combat
During combat, units typically face each other head on, but Galleys and likely other sea units should engage broadside to broadside.

Image

richardspink2:
﻿## Summary
- Import the Civ3 `TurnToAttack` flag into unit prototype data.
- Use that flag to rotate attack animations counter-clockwise for broadside units while keeping combat bonuses, retreats, and defensive bombard logic based on the real attack direction.
- Add `rotateBeforeAttack` data to the base ruleset so new games load the same unit prototype behavior.
- Add focused tests for normal facing, broadside facing, save conversion, and base-ruleset entries.

Fixes #902.

## Testing
- Created a loadable save file with two Frigates on adjacent coast tiles, loaded it in OpenCiv3, and fought the ships to check the firing direction.
- `dotnet build C7/C7.sln`
- `dotnet format C7/C7.sln whitespace --verify-no-changes`
- `dotnet test C7/C7.sln` with `CIV3_HOME` set: 62 passed
- `git diff --check`
- `dotnet list C7/C7.sln package --vulnerable --include-transitive`: no vulnerable packages

### DIFF
--- a/C7/Lua/game_modes/base-ruleset.json
+++ b/C7/Lua/game_modes/base-ruleset.json
@@ -792,6 +792,7 @@
       "defense": 0,
       "bombard": 0,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Egypt",
@@ -868,6 +869,7 @@
       "defense": 0,
       "bombard": 0,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Egypt",
@@ -945,6 +947,7 @@
       "defense": 0,
       "bombard": 0,
       "movement": 2,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Russia",
         "America",
@@ -989,6 +992,7 @@
       "defense": 0,
       "bombard": 0,
       "movement": 2,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Egypt",
@@ -1055,6 +1059,7 @@
       "defense": 6,
       "bombard": 0,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Egypt",
@@ -1125,6 +1130,7 @@
       "defense": 11,
       "bombard": 0,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Egypt",
@@ -1195,6 +1201,7 @@
       "defense": 1,
       "bombard": 0,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "A Barbarian Chiefdom",
         "Rome",
@@ -1263,6 +1270,7 @@
       "defense": 1,
       "bombard": 1,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Egypt",
@@ -1329,6 +1337,7 @@
       "defense": 2,
       "bombard": 0,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Egypt",
@@ -1393,6 +1402,7 @@
       "defense": 2,
       "bombard": 0,
       "movement": 1,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Egypt",
         "Greece",
@@ -1461,6 +1471,7 @@
       "defense": 1,
       "bombard": 0,
       "movement": 2,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "Rome",
         "Greece",
@@ -1530,6 +1541,7 @@
       "defense": 1,
       "bombard": 0,
       "movement": 2,
+      "rotateBeforeAttack": false,
       "producibleBy": [
         "A Barbarian Chiefdom",
         "Rome",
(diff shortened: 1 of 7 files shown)


ajhalme:
Thanks, looks pretty good.

A few issues, I've submitted a patch on your branch here: https://github.com/richardspink2/OpenCiv3/pull/1

- The rotation should be CCW (did you test this at all?)
- For this new unit property `rotateBeforeAttack` to function correctly in a new game, the unit prototypes in the "base ruleset" need to be told about it.
- And finally, the merge issue here is due to a recent merge on Development

sea_fight

richardspink2:
﻿Thanks for catching that. I did test it with a loadable save containing two adjacent Frigates, but I was checking the broadside behavior rather than the handedness of the rotation. I’ve corrected it to CCW, added the base-ruleset entries, rebased on latest Development, and added automated coverage so this no longer depends on visual inspection.

C7/Lua/game_modes/base-ruleset.json
```diff
@@ -2504,6 +2504,7 @@
       "defense": 0,
       "bombard": 16,
       "movement": 2,
+      "rotateBeforeAttack": true,
```
stavrosfa: As @ajhalme correctly pointed out, this was missing in the first commit, but I want to add my 2 cents here as well. I would refrain from doing this manually, I would load a standard civ3 game, save it as json, take the whole unitPrototype part of the json and paste it over the current one.
This will make sure that we don't make any falty assumptions on our end and everything has the same format.

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -313,15 +313,25 @@ public double RetreatChance(MapUnit opponent, bool isAttacking) {
...
+			// Set combat animation facing. We'll restore the defender's original facing direction at the end of the battle.
+			TileDirection attackDirection = attacker.location.directionTo(defender.location);
 			var defenderOriginalDirection = defender.facingDirection;
-			defender.facingDirection = attacker.facingDirection.reversed();
+			attacker.facingDirection = attacker.GetAttackAnimationDirection(attackDirection);
+			defender.facingDirection = defender.GetDefenseAnimationDirection(attackDirection);
 
-			IEnumerable<StrengthBonus> attackBonuses  = attacker.ListStrengthBonusesVersus(defender, CombatRole.Attack , attacker.facingDirection),
-								   defenseBonuses = defender.ListStrengthBonusesVersus(attacker, CombatRole.Defense, attacker.facingDirection);
+			IEnumerable<StrengthBonus> attackBonuses  = attacker.ListStrengthBonusesVersus(defender, CombatRole.Attack , attackDirection),
```
stavrosfa: I would keep it as `attacker.attackDirection`. Even if it seems redundant, this was probably done in the first place to make the code is more readable.
Some of these methods (fight, move, etc) are pretty long, and it might be hard to follow the logic and keep track of which unit `this `is.

stavrosfa (commented):
Thanks, looks good overall!

C7/Lua/game_modes/base-ruleset.json
ajhalme: I did this, but there was quite a bit of noise from unit upgrades and things. Makes sense to bring in the rotateBeforeAttack=false cases as well, so they don't become noise for the next person updating the rules.

I think we need a better/automated mechanism for updating the base ruleset.

stavrosfa: Hmm, what extra fields did a `unitPrototypes` entry include? If there are extra stuff, either your repo is not up to date, or there is something we have already missed.

>I think we need a better/automated mechanism for updating the base ruleset.

I agree, do you have something specific in mind?

C7Engine/C7GameData/MapUnit.cs
richardspink2: Updated the direction variables to attackerAttackDirection and defenderDefenseDirection so the fight method stays explicit about which unit and direction are being used while keeping combat logic separate from animation-facing direction.

C7/Lua/game_modes/base-ruleset.json
richardspink2: Updated the base ruleset so every unitPrototype now declares rotateBeforeAttack. Non-rotating entries are explicitly false, and the TurnToAttack units remain true. The test now fails if any unitPrototype omits the field or if the true set changes unexpectedly. Kept the ruleset automation idea out of this PR so this stays scoped to #902.

stavrosfa:
Don't hate me for this @richardspink2, usually I am not that slow, but I totally missed this up until now.

This is the first, (if I am not completely mistaken) of what the editor calls Unit Abilities, to be implemented.

image


I think the best way to go about this, is to have these as flags, similar to what buildings do 

e.x.

image

For 1 such ability, having boolean fields is not such a big deal, but there are 31 abilities * 124 unit prototypes = 3844 fields, with mostly grabage information.

I don't know if that makes sense to you both.

richardspink2:
﻿That makes sense. I updated this to use a unit prototype flag instead of a saved boolean field, matching the existing building/tech flag pattern. The runtime still has a rotateBeforeAttack convenience property backed by the flag, and the base ruleset now only marks the TurnToAttack units with flags: ["rotateBeforeAttack"] instead of carrying false values for every other unit.

ajhalme:
LGTM, working fine.

ajhalme (approved):

stavrosfa (approved):
Thanks, looking great overall!

PR #935, opened 2026-05-15
Some improvement to the game loop

stavrosfa:
This PR addresses the issue, where upon launching the game, if the player had finished their turn, the AI would immediatly take control, and the game would either get stuck in a loop, or have a black screen.

The main issue was occuring because after loading the game (game data & map), a call to advance the turn was made.

If the human player had finished their turn, either by choise, or by (game) design, that meant the loop was moving on to the next player in the list, effectively skipping the human player. 
Now that was working well when in regular play mode, but when loading in directly where the human player has finished their turn, the issue becomes clear.

So, now, no matter what, when loading a new/saved/scenario game, the human player will go first. Even if they have no units left, the have to end the turn manually with the usual methods.

### Some other things:

- Some of the `Game` class code was re-ordered, to try and make it more readable, and logically grouped where I thought it would be useful.

- Removed the PreGame entry from the `GameState` enum as it wasn't really used anywhere, and also interfered with fixing the main bug of this PR

- I added the `skipFirstTurn` player flag when loading up a .biq scenario (eample usage: `9 WWII in the Pacific.biq`)

- I disabled being able to click anything while the AI are playing their turns.

- Some class clean-up, renaming and removing unused imports, etc

- Hide all the buttons when they are created to avoid showing them all at once, they already are getting shown/hidden properly when a unit is selected.

- Added `isSettler` flag to the `UnitPrototype` class, so that we don't have to check the string name of the prototype

I hope I didn't miss anything, let me know if you have questions or comments

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -9,7 +9,72 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 
+public class GotoInfo {
+	public Tile destinationTile = null;
+	public int moveCost = -1;
+	public TilePath path = null;
+	public HashSet<System.Numerics.Vector2> pathCoords;
+	public bool attackingMove = false;
+	public Player requiresWarDeclarationOnPlayer = null;
+};
+
+public class TileInfo {
+	public Tile targetTile;
+	public HashSet<Tile> coveredTiles = [];
+	public HashSet<Tile> cityLabelsToHide = [];
+
+	public TileInfo(Tile tile) {
+		targetTile = tile;
+
+		List<TileDirection> coverage = [TileDirection.SOUTHWEST, TileDirection.SOUTH, TileDirection.SOUTHEAST];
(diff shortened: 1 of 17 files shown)


ajhalme:
LGTM. Appears to fix the attack-on-load crash game save I shared recently. :+1:

ajhalme (approved):

stavrosfa:
Adding the save for the crash game you mentioned for future reference. 

Another good example that this PR is working is the `9 WWII in the Pacific.biq` scenario when picking anyone except Japan, and since they skip their first turn, all units are disabled (which should trigger the original error in some way, and Japan should start moving, and also incrementing the turn to Jan 1942), but the turn starts with the human player and the game waits for the button or Spacebar/Enter to be pressed.

[on_load_animation_bug_2.json.zip](https://github.com/user-attachments/files/28032134/on_load_animation_bug_2.json.zip)

PR #936, opened 2026-05-18
Add basic transport load/unload

ajhalme:
Add basic support for transport mechanic: loading and unloading units from a vehicle with transport capacity.

- Add UI element to show transport contents
- Extend C7 and unit actions with load/unload
- Add engine message for load/unload
- Load unit load status from Civ3 saves
- Adjust MapUnit `move(..)` to allow load/unload
- Adjust MapUnit `move(..)` so transports can move other units
- Adjust tile unit display to prioritise transport over strongest defender, when out to sea
- Adjust tile context menu to highlight transport-loaded units
- Add load and unload buttons, with keyboard shortcuts (`L`/`U`)
- Extend base ruleset with capacity (+patch missing/misplaced things)

## Manual

**Loading and Unloading**
You can have a ship wait until it is loaded to capacity with units by clicking the Load order or pressing [L]. Boarding a
ship uses up all a unit’s movement points for the turn. If you attempt to move a naval unit into a land square that does not contain a port city, any passengers who have not already moved this turn are offered the option to disembark and make landfall. You can also order a ship to unload all its passengers by clicking the Unload order or pressing [L].

## Demo

transport_basic

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -360,6 +360,10 @@ public partial class Game : Node {
 			case MsgUnitMoved mUUAAB:
 				EmitSignal(SignalName.UnitMoved, new ParameterWrapper<MapUnit>(mUUAAB.Unit));
 				break;
+			case MsgTransportUnloaded mTU:
+				// UnitMoved is enough to refresh UI
+				EmitSignal(SignalName.UnitMoved, new ParameterWrapper<MapUnit>(mTU.Unit));
+				break;
 		}
 	}
 
@@ -612,8 +616,6 @@ public partial class Game : Node {
 		bool canMove = unitSelector.SetSelectedUnit(unit);
 		if (!canMove) {
 			TemporaryPopup.Show(this, "This unit has already moved.", screenPosition);
-		} else {
-			unit.ResetFacingDirection();
 		}
 	}
 
@@ -998,8 +1000,9 @@ public partial class Game : Node {
 				new MsgLoadToTransport(CurrentlySelectedUnit.id).send();
 		}
 		if (currentAction == C7Action.UnitUnload) {
-			if (CurrentlySelectedUnit != MapUnit.NONE && CurrentlySelectedUnit != null)
-				new MsgUnloadFromTransport(CurrentlySelectedUnit.id).send();
+			if (CurrentlySelectedUnit != MapUnit.NONE && CurrentlySelectedUnit != null) {
+				new MsgUnloadTransport(CurrentlySelectedUnit.id).send();
+			}
 		}
 
 		Terraform terraform = C7Action.ToTerraform(currentAction);
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -959,6 +959,17 @@ public partial class Game : Node {
 			}
 		}
 
+
+		if (currentAction == C7Action.UnitLoad) {
+			// TODO: Which transport?
+			if (CurrentlySelectedUnit != MapUnit.NONE && CurrentlySelectedUnit != null)
+				new MsgLoadToTransport(CurrentlySelectedUnit.id).send();
+		}
+		if (currentAction == C7Action.UnitUnload) {
+			if (CurrentlySelectedUnit != MapUnit.NONE && CurrentlySelectedUnit != null)
+				new MsgUnloadFromTransport(CurrentlySelectedUnit.id).send();
+		}
+
 		Terraform terraform = C7Action.ToTerraform(currentAction);
 
 		if (CurrentlySelectedUnit == MapUnit.NONE || CurrentlySelectedUnit == null
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -559,7 +559,7 @@ public partial class Game : Node {
 			}
 		} else {
 			// Select unit on tile at mouse location
-			HandleUnitSelection(eventMouseButton);
+			HandleUnitSelectionTileClick(eventMouseButton);
 		}
 	}
 
@@ -572,22 +572,34 @@ public partial class Game : Node {
 		}
 	}
 
-	private void HandleUnitSelection(InputEventMouseButton eventMouseButton) {
+	private void HandleUnitSelectionTileClick(InputEventMouseButton eventMouseButton) {
+
 		Tile tile = PositionToTile(eventMouseButton.Position);
 		if (tile == null) {
 			return;
 		}
 
 		// TODO: This should really be the top unit.
-		MapUnit to_select = tile.unitsOnTile.FirstOrDefault();
-		if (to_select == null || to_select.owner != controller) {
+		MapUnit unit = tile.unitsOnTile.FirstOrDefault();
+		if (unit == null || unit.owner != controller) {
 			return;
 		}
 
-		bool canMove = unitSelector.SetSelectedUnit(to_select);
+		SelectUnit(unit, eventMouseButton.Position);
+	}
+
+	private void SelectUnit(MapUnit unit, Vector2 screenPosition) {
+		bool canMove = unitSelector.SetSelectedUnit(unit);
 		if (!canMove) {
-			TemporaryPopup.Show(this, "This unit has already moved.", eventMouseButton.Position);
+			TemporaryPopup.Show(this, "This unit has already moved.", screenPosition);
+		} else {
+			unit.ResetFacingDirection();
 		}
+	}
+
+	public void SelectUnit(MapUnit unit) {
+		var screenPos = mapView.screenLocationOfTile(unit.location);
+		SelectUnit(unit, screenPos);
 	}
 
 	private void HandleRightMouseButton(InputEventMouseButton eventMouseButton) {
--- a/C7/Lua/texture_configs/civ3.lua
+++ b/C7/Lua/texture_configs/civ3.lua
@@ -420,6 +420,13 @@ textures.lower_left_infobox = {
     },
 }
 
+textures.transport_infobox = {
+    box = {
+      path = INTERFACE .. "box trans color.pcx",
+      alpha = INTERFACE .. "box trans alpha.pcx",
+    },
+}
+
 textures.tileinfo = {
       box = {
         path = TILEINFO
--- a/C7/UIElements/GameStatus/StatusUtils.cs
+++ b/C7/UIElements/GameStatus/StatusUtils.cs
@@ -2,26 +2,24 @@ using C7.Textures;
 using C7GameData;
 using Godot;
 
-public static class StatusUtils
-{
-    public static (Sprite2D unitSprite, Sprite2D unitTintSprite) GetUnitSprites(Game game, MapUnit unit)
-    {
-        var am = game.animationController.civ3AnimData;
-        
-        var (baseFrame, tintFrame) = am.GetAnimationFrameAndTintTextures(unit);
-        
-        var color = unit.owner.GetPlayerColor();
-        ShaderMaterial material = PlayerTextureUtil.GetShaderMaterialForUnit(color);
+public static class StatusUtils {
+	public static (Sprite2D unitSprite, Sprite2D unitTintSprite) GetUnitSprites(Game game, MapUnit unit) {
+		var am = game.animationController.civ3AnimData;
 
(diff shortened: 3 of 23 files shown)


C7/UIElements/GameStatus/StatusUtils.cs
```diff
@@ -0,0 +1,25 @@
+using C7.Textures;
+using C7GameData;
+using Godot;
+
+public static class StatusUtils {
```
stavrosfa: StatusUtils doesn't seem like a very suitable name for this class. Perhaps something like SpriteUtils?

C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -1186,6 +1204,7 @@ private void ImportUnitPrototypes() {
 				prototype.attack = prto.Attack;
 				prototype.defense = prto.Defense;
 				prototype.movement = prto.Movement;
+				prototype.capacity = prto.Capacity;
```
stavrosfa, line 1210: We are missing a couple of Unit special actions. 

Looking at the editor, each unit (PRTO) than can be loaded into something (not sure if this means an Army as well, but I am pretty sure it means a transport unit), has a "Special Action" called `Load`

Check the Infantry for example, it has the `Load` action

Image

In contrast, a Galley that can load, and subsequently unload units, has the `Unload` action

Image

The Galley doesn't have the `Load` because, it can't be loaded into something else.

These two actions are located here 
https://github.com/C7-Game/OpenCiv3/blob/8119dd4fb8a9238a5d0ffaabb74df0c826518d38/QueryCiv3/BiqSections/Prto.cs#L164

So that means, that we can directly check if a unit can be loaded into something that can load units, right now we can load anything, to a unit with enough capacity.

C7/Lua/texture_configs/civ3.lua
```diff
@@ -420,6 +420,13 @@ textures.lower_left_infobox = {
     },
 }
 
+textures.transport_infobox = {
+    box = {
+      path = INTERFACE .. "box trans color.pcx",
```
stavrosfa, line 425: These two need to be imported somewhere around here
https://github.com/C7-Game/OpenCiv3/blob/8119dd4fb8a9238a5d0ffaabb74df0c826518d38/C7/Lua/texture_configs/c7.lua#L67
and we also need to add the equivalent texture for the standalone mode

C7/UIElements/GameStatus/TransportInfoBox.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Linq;
+using C7.Textures;
+using Godot;
+using C7GameData;
+using Serilog;
+using C7Engine;
+
+[GlobalClass]
+[Tool]
+public partial class TransportInfoBox : Civ3TextureRect {
```
stavrosfa, line 12: Feel free to ignore this, but my ocd doesn't let me ignore that the box is missalighned compared to the LowerInfoBox. I know it's the same in the original, feel free to leave this as is 😄 

A few actual comments on the box:

1. We can click throught, although we shouldn't
2. The textures in the original are about half the texture size of the InfoBox. Perhaps to accomodate for larger capacities? Should we stick to that?
3. In the original, there is a box outline around (only) the transport unit? Again, should we be replicating that?
4. We should be able to select a unit inside a transport to activate/wake, etc.. That also probably means, that these are not simple texture, rather texture buttons.

stavrosfa (commented):
Good stuff, I can imagine this being a slight pain to work out 😅 

I have encountered some bugs and such, but i mostly think these can be attributed to the fact that anything can be loaded into anything with enough capacity. Should be ok once this is implemented.

C7/UIElements/GameStatus/StatusUtils.cs
ajhalme: Renamed to "SpriteUtils", pushed up a level

C7/UIElements/GameStatus/TransportInfoBox.cs
ajhalme: - Adjusted box position a bit -- sadly the visually heavy next turn blinker makes the simple box looks out of place still
- Removed the click-through, but need to revisit the texture button even capture
- Resized units sprites, but each unit still goes into a 50x50 unit button -- unit sprites don't seem to share a common baseline, or I'm missing something
- Added transport unit highlight box with Line2D

C7Engine/C7GameData/ImportCiv3.cs
ajhalme: Lines 1151-1152 in `ImportCiv3.ImportBicUnits()` solve this, I think. I added them in, but then commented out as I figured I'd get the `capacity` and some cleanup in first in the base ruleset. I can take another look.

For the base ruleset, I think the unit promotion stuff isn't quite lined up perfectly, as when I save, I seem to get unitPrototype upgrade paths featuring unique units instead of the regular ones.

stavrosfa: Not sure what lines 1151-1152 do, I checked yout PR and the dev branch, I don't see anything relevant to this. Can you link the lines from the current dev branch?

Regarding the unit upgrades, can you share a save and explain which ones don't work as expected? Also, do you mean that they broke during this feature development, or were they already broken?

C7/UIElements/GameStatus/TransportInfoBox.cs
stavrosfa: >unit sprites don't seem to share a common baseline, or I'm missing something

I think I know what you probably mean. Since these sprites are practically a single frame from a .flic file, you should know that each .flic animation, has a X and Y offset `from the top left corner` of the image in order to able to place them all at the same "baseline".

You can see some more details here

https://github.com/C7-Game/OpenCiv3/blob/c2e56fa7ac95cbff17a85e1e91abe5ffffbdbc3d/C7/Map/UnitLayer.cs#L136

here
https://github.com/C7-Game/OpenCiv3/blob/c2e56fa7ac95cbff17a85e1e91abe5ffffbdbc3d/C7/Animations/AnimationManager.cs#L331

and here
https://github.com/C7-Game/OpenCiv3/blob/c2e56fa7ac95cbff17a85e1e91abe5ffffbdbc3d/ConvertCiv3Media/ReadFlic.cs#L24

stavrosfa: Having said that, I had another look at the original, it doesn't seem to use the flic thumbnail, rather the `units_32.pcx` image that is also used in the city production pop-up/screen.

These are already the correct size and are alligned, the current implementation is missing the civ color tint for the small thumbnails, which I have implemented and sits on my hard drive.

Maybe for now, just use the `units_32.pcx` instead, even with the wrong colours, and this will be updated at some point along with the others.

C7/Lua/texture_configs/civ3.lua
ajhalme: Added, basic art ready over at https://github.com/C7-Game/Assets/pull/7

C7/UIElements/GameStatus/TransportInfoBox.cs
ajhalme: Added transport click buttons for selection.

Scaling unit sprites seems to work fine for the transported, inclined to leave as is.

C7Engine/C7GameData/ImportCiv3.cs
ajhalme: I've documented the drift between current base ruleset and what I'm seeing when applying the `unitPrototype` field from a saved game to the base ruleset here: https://github.com/C7-Game/OpenCiv3/pull/937

C7/Game.cs
```diff
@@ -572,24 +572,36 @@ private void OnDoubleLeftMouseButtonClick(InputEventMouseButton eventMouseButton
...
+		SelectUnit(unit, eventMouseButton.Position);
+	}
+
+	private void SelectUnit(MapUnit unit, Vector2 screenPosition) {
+		bool canMove = unitSelector.SetSelectedUnit(unit);
 		if (!canMove) {
-			TemporaryPopup.Show(this, "This unit has already moved.", eventMouseButton.Position);
+			TemporaryPopup.Show(this, "This unit has already moved.", screenPosition);
+		} else {
+			unit.ResetFacingDirection();
```
stavrosfa: Why is this done? It looks weird too I think

stavrosfa (commented):
Not a whole lot of comments. I think this looks mostly done,
I would still like the Load/Unload special actions to be a part of this, but if you don't intend to add them here, please add a TODO, and you could also link this PR and my comments so it's quite clear what needs to be done in the future.

C7/Game.cs
ajhalme: Reusing some of the existing unit selection code resulted in units selected from inside a transport pointing `NORTH` every time. For transport units, this default pose needs applying somewhere, but maybe this is not the place to do it.

ajhalme:
> I would still like the Load/Unload special actions to be a part of this

I agree, would be nice to have these. Happy to look into it, but I think the base ruleset needs cleaning up first. Otherwise it's a bit of a hassle to cherry-pick the right `unitPrototype` rows. (See the issue outlined in #937)

stavrosfa:
A few comments from my play testing.

1. When a unit is loaded and there is no other defending unit, we should be showing the transport instead of the loaded unit (unless explicitly selected).

What is currently happenning:
Screenshot_88

These few updated lines in the UnitLayer seem to solve this, although it might need some more tweaking, not sure

```
MapUnit bestDefender = units.First(u => u.loadedOnUnitId == null), selected = null, transporter = null, doingInterestingAnimation = null;
var currentlySelectedUnit = looseView.mapView.game.CurrentlySelectedUnit;

foreach (var u in units) {
          if(u.loadedOnUnitId != null && u != currentlySelectedUnit) continue;
```

2. The Unload action button should be shown for the transport units (except Army I think), not for the loaded units themselves.
What is currently happenning (Nipponese Infantry is selected) :
Screenshot_90

So the Unload action, should be on the trasport and Unload the loaded units (the original game gives a few options to what you want to unload, perhaps a future TODO?)


3. Out-of-scope comment/thought
When a coastal city tile has a transport with loaded units and there no other defending units, if a land unit attacks, should the transport be destroyed/captured, since it cannot fight back, and the loaded units destroyed in any case? I don't think there is a case where the trasnport fights back..
I tried to create a scenario where I have only loaded units on a trasport on a city tile, but there is no option to load them, so I couldn't test it.

4. Out-of-scope comment/thought no 2
If I want to send a land unit on a transport that is on a sea tile but right next to a land tile. I can't use the go-to (this is faithful to the original). I can however use the numpad keys to move it and load it in (again this is faithful to the original) 

Should we keep it like this, or refine in the future?

5. I see you implemented the CanCarryAircraft actions etc, but the actuall mechanic is not implemented, correct? I can see for example that the Bombers/Fighters are not loaded on the Carriers.

image

That's fine if that's so, and we can leave it for the future, I am jsut making sure I understand correctly.

ajhalme:
Thanks for taking a look. I'll look into (1) and (2).

(3) might one be one for the Civ experts on Discord to weigh in on. I wonder if a test could be created via editor by giving player character a city with transports and units, then setting up the scene in-game, and then back in editor handing over the city and units to opponent.

(4) seems like a useful quality of life upgrade. (5) is not implemented, but as you say, I loaded up the flags: feels like a nice first issue for a new contributor. Indeed (4) and (5) together could be nice issue, maybe with (3) as well, if we get agreement on the behaviour.

C7/Game.cs
ajhalme: Moved the ResetFacingDirection to unit boarding, making boarding even more like `fortify()` without animation.

ajhalme:
(1) and (2) implemented

stavrosfa:
That looks fine as far as I am concerned! Thanks!

stavrosfa:
BTW, if you come across things that you feel they could be a good fit as a first time issue for newcomers, consider adding them as issues and tagging them as such (especially if we don't have many).

I 'll try to do the same, so we can have a variety of first tasks, around 5-10 maybe, and we can update them as we go.

PR #937, opened 2026-05-24
Cleaning up base ruleset drift

ajhalme:
**EDIT:** From draft to a fix

- Made a script for applying changes in saved game `unitPrototype` to the matching base ruleset field.
- Revised the unitPrototype Civ3 load logic
- Misc fixes in base ruleset 

~~This draft PR is not intended to be merged as is, just a way to document "drift" between what the save files are given and what the base ruleset provides. The diff/drift here is based on latest Development branch as of this writing.~~

## Issues

Notable issues when reading a Civ3 save in OC3, saving the game in standard OC3 json save file format, and then applying `unitPrototype` on top of the OC3 base ruleset:
- Unique units replacing `upgradeTo` paths
- flag placement (`rotateBeforeAttack`): the field order in code doesn't match current ruleset order, resulting in this add/remove noise
- blank `producibleBy` getting removed
- Unicode encoding of city names, etc.
- Whitespace

### DIFF
--- a/C7/Lua/game_modes/base-ruleset.json
+++ b/C7/Lua/game_modes/base-ruleset.json
@@ -960,7 +960,9 @@
         "Hittites",
         "Portugal"
       ],
-      "upgradeTo": "Explorer",
+      "upgradesTo": [
+        "Explorer"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1242,7 +1244,12 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Swordsman",
+      "upgradesTo": [
+        "Gallic Swordsman",
+        "Immortals",
+        "Legionary",
+        "Swordsman"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1310,7 +1317,10 @@
         "Byzantines",
         "Inca"
       ],
-      "upgradeTo": "Longbowman",
+      "upgradesTo": [
+        "Berserk",
+        "Longbowman"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1376,7 +1386,10 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Pikeman",
+      "upgradesTo": [
+        "Musketman",
+        "Pikeman"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1443,7 +1456,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Medieval Infantry",
+      "upgradesTo": [
+        "Medieval Infantry"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1514,7 +1529,10 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Horseman",
+      "upgradesTo": [
+        "Horseman",
+        "Mounted Warrior"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1587,7 +1605,15 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Knight",
+      "upgradesTo": [
+        "Ansar Warrior",
+        "Keshik",
+        "Knight",
+        "Rider",
+        "Samurai",
+        "Three-Man Chariot",
+        "War Elephant"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1657,7 +1683,10 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Musketman",
+      "upgradesTo": [
+        "Musketeer",
+        "Musketman"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1729,7 +1758,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Guerilla",
+      "upgradesTo": [
+        "Guerilla"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1798,7 +1829,10 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Rifleman",
+      "upgradesTo": [
+        "Rifleman",
+        "Swiss Mercenary"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1866,7 +1900,11 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Cavalry",
+      "upgradesTo": [
+        "Cavalry",
+        "Cossack",
+        "Sipahi"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -1940,7 +1978,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Infantry",
+      "upgradesTo": [
+        "Infantry"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -2081,7 +2121,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Mech Infantry",
+      "upgradesTo": [
+        "Mech Infantry"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -2153,7 +2195,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Modern Armor",
+      "upgradesTo": [
+        "Modern Armor"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -2373,7 +2417,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Trebuchet",
+      "upgradesTo": [
+        "Trebuchet"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -2444,7 +2490,10 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Artillery",
+      "upgradesTo": [
+        "Artillery",
+        "Hwach'a"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -2519,7 +2568,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Radar Artillery",
+      "upgradesTo": [
+        "Radar Artillery"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -2884,7 +2935,10 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Caravel",
+      "upgradesTo": [
+        "Caravel",
+        "Carrack"
+      ],
       "unproducible": false,
       "flags": [
         "rotateBeforeAttack"
@@ -2956,7 +3010,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Galleon",
+      "upgradesTo": [
+        "Galleon"
+      ],
       "unproducible": false,
       "flags": [
         "rotateBeforeAttack"
@@ -3105,7 +3161,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Transport",
+      "upgradesTo": [
+        "Transport"
+      ],
       "unproducible": false,
       "flags": [
         "rotateBeforeAttack"
@@ -3178,7 +3236,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Destroyer",
+      "upgradesTo": [
+        "Destroyer"
+      ],
       "unproducible": false,
       "categories": [
         "Sea"
@@ -3764,7 +3824,10 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Jet Fighter",
+      "upgradesTo": [
+        "F-15",
+        "Jet Fighter"
+      ],
       "unproducible": false,
       "categories": [
         "Air"
@@ -4318,7 +4381,9 @@
       "producibleBy": [
         "Aztecs"
       ],
-      "upgradeTo": "Swordsman",
+      "upgradesTo": [
+        "Swordsman"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -4358,9 +4423,10 @@
       "producibleBy": [
         "Babylon"
       ],
-      "upgradeTo": "Longbowman",
+      "upgradesTo": [
+        "Longbowman"
+      ],
       "unproducible": false,
-      "variantOf": "Archer",
       "categories": [
         "Land"
       ],
@@ -4399,9 +4465,10 @@
       "producibleBy": [
         "Greece"
       ],
-      "upgradeTo": "Musketman",
+      "upgradesTo": [
+        "Musketman"
+      ],
       "unproducible": false,
-      "variantOf": "Pikeman",
       "categories": [
         "Land"
       ],
@@ -4440,9 +4507,10 @@
       "producibleBy": [
         "Zululand"
       ],
-      "upgradeTo": "Musketman",
+      "upgradesTo": [
+        "Musketman"
+      ],
       "unproducible": false,
-      "variantOf": "Spearman",
       "categories": [
         "Land"
       ],
@@ -4481,9 +4549,10 @@
       "producibleBy": [
         "Rome"
       ],
-      "upgradeTo": "Medieval Infantry",
+      "upgradesTo": [
+        "Medieval Infantry"
+      ],
       "unproducible": false,
-      "variantOf": "Swordsman",
       "categories": [
         "Land"
       ],
@@ -4525,9 +4594,10 @@
       "producibleBy": [
         "Persia"
       ],
-      "upgradeTo": "Guerilla",
+      "upgradesTo": [
+        "Guerilla"
+      ],
       "unproducible": false,
-      "variantOf": "Medieval Infantry",
       "categories": [
         "Land"
       ],
@@ -4569,9 +4639,10 @@
       "producibleBy": [
         "Egypt"
       ],
-      "upgradeTo": "Knight",
+      "upgradesTo": [
+        "Knight"
+      ],
       "unproducible": false,
-      "variantOf": "Chariot",
       "categories": [
         "Land"
       ],
@@ -4613,9 +4684,10 @@
       "producibleBy": [
         "China"
       ],
-      "upgradeTo": "Cavalry",
+      "upgradesTo": [
+        "Cavalry"
+      ],
       "unproducible": false,
-      "variantOf": "Knight",
       "categories": [
         "Land"
       ],
@@ -4658,9 +4730,10 @@
       "producibleBy": [
         "Iroquois"
       ],
-      "upgradeTo": "Knight",
+      "upgradesTo": [
+        "Knight"
+      ],
       "unproducible": false,
-      "variantOf": "Horseman",
       "categories": [
         "Land"
       ],
@@ -4702,9 +4775,10 @@
       "producibleBy": [
         "France"
       ],
-      "upgradeTo": "Rifleman",
+      "upgradesTo": [
+        "Rifleman"
+      ],
       "unproducible": false,
-      "variantOf": "Musketman",
       "categories": [
         "Land"
       ],
@@ -4746,9 +4820,10 @@
       "producibleBy": [
         "Japan"
       ],
-      "upgradeTo": "Cavalry",
+      "upgradesTo": [
+        "Cavalry"
+      ],
       "unproducible": false,
-      "variantOf": "Knight",
       "categories": [
         "Land"
       ],
@@ -4790,9 +4865,10 @@
       "producibleBy": [
         "India"
       ],
-      "upgradeTo": "Cavalry",
+      "upgradesTo": [
+        "Cavalry"
+      ],
       "unproducible": false,
-      "variantOf": "Knight",
       "categories": [
         "Land"
       ],
@@ -4832,7 +4908,6 @@
         "Russia"
       ],
       "unproducible": false,
-      "variantOf": "Cavalry",
       "categories": [
         "Land"
       ],
@@ -4875,9 +4950,10 @@
       "producibleBy": [
         "Germany"
       ],
-      "upgradeTo": "Modern Armor",
+      "upgradesTo": [
+        "Modern Armor"
+      ],
       "unproducible": false,
-      "variantOf": "Tank",
       "categories": [
         "Land"
       ],
@@ -4968,7 +5044,6 @@
         "America"
       ],
       "unproducible": false,
-      "variantOf": "Jet Fighter",
       "categories": [
         "Air"
       ],
@@ -5085,9 +5160,10 @@
       "producibleBy": [
         "Mongols"
       ],
-      "upgradeTo": "Cavalry",
+      "upgradesTo": [
+        "Cavalry"
+      ],
       "unproducible": false,
-      "variantOf": "Knight",
       "categories": [
         "Land"
       ],
@@ -5130,7 +5206,6 @@
         "Spain"
       ],
       "unproducible": false,
-      "variantOf": "Explorer",
       "categories": [
         "Land"
       ],
@@ -5172,9 +5247,10 @@
       "producibleBy": [
         "Scandinavia"
       ],
-      "upgradeTo": "Guerilla",
+      "upgradesTo": [
+        "Guerilla"
+      ],
       "unproducible": false,
-      "variantOf": "Longbowman",
       "categories": [
         "Land"
       ],
@@ -5214,7 +5290,6 @@
         "Ottomans"
       ],
       "unproducible": false,
-      "variantOf": "Cavalry",
       "categories": [
         "Land"
       ],
@@ -5257,9 +5332,10 @@
       "producibleBy": [
         "Celts"
       ],
-      "upgradeTo": "Medieval Infantry",
+      "upgradesTo": [
+        "Medieval Infantry"
+      ],
       "unproducible": false,
-      "variantOf": "Swordsman",
       "categories": [
         "Land"
       ],
@@ -5301,9 +5377,10 @@
       "producibleBy": [
         "Arabia"
       ],
-      "upgradeTo": "Cavalry",
+      "upgradesTo": [
+        "Cavalry"
+      ],
       "unproducible": false,
-      "variantOf": "Knight",
       "categories": [
         "Land"
       ],
@@ -5346,9 +5423,10 @@
       "producibleBy": [
         "Carthage"
       ],
-      "upgradeTo": "Musketman",
+      "upgradesTo": [
+        "Musketman"
+      ],
       "unproducible": false,
-      "variantOf": "Pikeman",
       "categories": [
         "Land"
       ],
@@ -5387,7 +5465,9 @@
       "producibleBy": [
         "Korea"
       ],
-      "upgradeTo": "Artillery",
+      "upgradesTo": [
+        "Artillery"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -5460,7 +5540,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "Guerilla",
+      "upgradesTo": [
+        "Guerilla"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -5533,7 +5615,9 @@
         "Inca",
         "Maya"
       ],
-      "upgradeTo": "TOW Infantry",
+      "upgradesTo": [
+        "TOW Infantry"
+      ],
       "unproducible": false,
       "categories": [
         "Land"
@@ -6420,9 +6504,10 @@
       "producibleBy": [
         "Sumeria"
       ],
-      "upgradeTo": "Pikeman",
+      "upgradesTo": [
+        "Pikeman"
+      ],
       "unproducible": false,
-      "variantOf": "Spearman",
       "categories": [
         "Land"
       ],
@@ -6461,9 +6546,10 @@
       "producibleBy": [
         "Hittites"
       ],
-      "upgradeTo": "Knight",
+      "upgradesTo": [
+        "Knight"
+      ],
       "unproducible": false,
-      "variantOf": "Chariot",
       "categories": [
         "Land"
       ],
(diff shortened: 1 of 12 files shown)


ajhalme:
The ruleset-save patch/compare script below. (note `-n` flag for a dry-run)

I run it like so:
```
./merge.sh <OpenCiv>/C7/Lua/game_modes/base-ruleset.json <Civ3>/Conquests/Saves/curragh5.SAV.json

```

```
#!/usr/bin/bash

# Usage: ./merge.sh [-n] <A.json> <B.json>
# -n  Dry run: show diff without modifying A.json

DRY_RUN=false

if [ "$1" = "-n" ]; then
    DRY_RUN=true
    shift
fi

if [ $# -ne 2 ]; then
    echo "Usage: $0 [-n] <target.json> <source.json>"
    exit 1
fi

A="$1"
B="$2"
UPDATED="${A%.json}_updated.json"

if [ ! -f "$A" ]; then
    echo "Error: '$A' not found"
    exit 1
fi

if [ ! -f "$B" ]; then
    echo "Error: '$B' not found"
    exit 1
fi

jq --slurpfile b "$B" '.unitPrototypes = $b[0].unitPrototypes' "$A" > "$UPDATED"

if [ "$DRY_RUN" = true ]; then
    echo "Dry run — diff of '$A' vs proposed changes:"
    diff "$A" "$UPDATED"
    rm "$UPDATED"
else
    mv "$UPDATED" "$A"
    echo "Done"
fi
```

stavrosfa:
I didn't realize that this was blocking progress (e.x. Load/Unload abilities in PR #936 ), but it makes sense.

If the bug is only when saving (which is my understanding), maybe we can do the simplest thing possible, and stop overwritting the parts of the json file that should be static, UnitPrototypes being one of them. These are not expected to change mid-game, as well as the map size for example, etc. These are mostly the "rules" of the game, not the gameplay itself.

So we could start/load a game, cache in this case the UnitPrototypes portion, and when the times comes to save a game, we can simply add this back.
It doesn't feel like a hack to me, and it serves exactly its purpose. Also, for the modding part, even if the json remains the same, someone could ovewrite it anyway using lua during initialization, so it should affect this either.

Also, after writing this, I realized that the original game pretty much does the same thing. The .sav file seems to have two (I mean virtual here, but I think they are also pretty distinctly separated in the binary file) sections (as opposed to one in the .biq file), one is for the BIQ stuff (rules) and another for the SAV stuff (gameplay). I can't know how the original implementation is, but I believe it's close to what I am describing here.

Let me know what you think about this approach, or if you have come up with something else entirely.

ajhalme:
I figured out an alternative implementation of `ImportUnitUpgrades()` in the Civ3 import code. It's not perfect, but it works quite well. Had to add hacks Berserker, which was undershooting the upgrade target, and Swiss Mercenary, which was overshooting.

With these code changes plus the base ruleset cleanup shown in this PR, we end up with save files where unitPrototype definitions match the base.

Freezing the base ruleset UnitPrototype definitions and applying them to saves as is, is a reasonable, but I am bit concerned it might hide some bugs we'd rather surface. And freezing doesn't solve the practical problem of how to apply changes to the base ruleset cleanly and correctly when new features have been created.

stavrosfa:
I run the current test suite locally for this PR, and it fails in at least 2 places. If I comment out the failed Assert, the next one fails, and so on.

Although I normally wouldn't have an issue with hardcoding some stuff like you did for Berserk and Swiss Mercenary, keep in mind it's not just the conquests.biq we are parsing here. There are multiple scenarios with different unit upgrade chains, and even unknown / future scenarios that we haven't / can't test.

That is why I proposed something like saving the state when importing, because then we only need need to worry about only 1 place in the pipeline. If we get the importing stuff right, then, the saving of the same stuff, should be 0 effort.

Regarding the Berserk and Swiss Mercenary upgrades, is the issue present in the current (development branch) implementation? 

Also, if we find weird cases, I believe it's worth writing a corresponding test about this, so we can narrow it down as much as possible.

ajhalme:
Ok, I think I've got something.

First of all apologies for wasting review time on the previous: my Mac wasn't fully set up, and given how we organise the tests, I was getting green across the board even for stuff wasn't running at all. I've addressed this by changing the test annotations with imported library `Xunit.SkippableFact`, which allows for explicit Skip when a Boolean property is true. Skipped test show up in test runners in a different way. No more silent fails!

Turns out I SEVERELY underestimated how subtle the unit upgrade scheme is. I thought I figured out how it works, but I was miles off and didn't get back on track until I spotted this link in some change-set:
https://forums.civfanatics.com/threads/how-to-upgrade-regular-units-to-uus.108396/

The reason I want to change the way it currently is set up -- the way it is in Civ3 save files -- is because it's a completely mad scheme at first glance. If we can hide this complexity from modders, we should do that. Nobody (else) needs to figure out why Samurai seems to upgrade to War Elephant.

A much more elegant solution (IMO) is to precompute the upgrade "closure" at Civ3 load, and to normalise on a representation where each unit's _every_ upgrade option is explicitly listed out in the base ruleset. For example, instead of "Samurai -> War Elephant" and runtime scanning to figure out what the _real_ upgrade is, we can just have "Samurai -> [Cavalry]" and "Horseman -> ["Ansar Warrior", "Keshik", "Knight", "Rider", "Samurai", "Three-Man Chariot", "War Elephant"], etc. Civ-parameterised upgrade functions will always filter out invalid options, so the choice is always unique (or null). We still need to do some runtime scanning and resolving, but this representation has the added bonus that upgrade "families" or "unit variants" are automatically grouped together, as they should be. And if you squint, you can just about see how all the civs perfectly map to the upgrade targets of each unit.

I'm intentionally changing the field from `upgradeTo` to `upgradesTo` to ensure that old saves will break and aren't misread for extra confusion. (We should maybe consider a save file version number at some point...)

The PR now contains:
- base ruleset cleanup / stable formatting
- revised Civ3 unitproto import
- revised runtime upgrade resolution funcs
- `upgradesTo` re-wiring
- skippable test annotation + new test extension library dependency

--

If happy with this approach, the .JSON-test UnitAvailability_JSON needs updating to accommodate these change - I don't know what that involves. Both the regular game test and the Conquests scenario upgrade test now run green for me.

If not happy with this approach, I can (reluctantly) strip this back to just the base ruleset formatting changes - I see now why the upgradeTo values are so weird. (But maybe this could be better documented.)

stavrosfa:
>First of all apologies for wasting review time on the previous

No time was wasted, it was a solution that I didn't like, but it was also one that you eliminated as a possible one (and one I didn't have to think/implement/dismiss), so it's a win for me.

>Turns out I SEVERELY underestimated how subtle the unit upgrade scheme is

Yeah, it's a lot more convoluted than what it seems, you can only see it when you get your hands dirty..

Now to the current solution.
I like this one, and as my previous attempt at a very simple scheme didn't work quite as well as I had hoped, I think this is a step in the right direction.
It still has some issues that the current test suite doesn't check, but I think we are getting there.

When we play as the Netherlands, the Spearman should upgrade to Swiss, and then to Rifleman. But even with the right techs/resources for a Rifleman, the upgrade to a Spearman is still the Swiss. The name `GetProducibleUpgrade` is misleading at this point, and also, think about what would happen if we actually had the button to upgrade the unit.

If this gets ironed out I think we will have something good on our hands

Here is an updated test to showcase a problematic behaviour (#region Netherlands)

<details>
<summary>UnitAvailability_SAV</summary>

	public async void UnitAvailability_SAV() {
		// This tests a Conquests game with Conquests rules from a .SAV file

		Skip.If(Civ3TestData.ShouldSkipCiv3DependentTests(), "No Civ3 install found.");

		#region setup

		string saveName = "Conquests 16 Players.SAV";
		string uri = "https://www.dropbox.com/scl/fi/gmxbx1mtrammzfc6vly1g/Conquests-16-Players.SAV?rlkey=2z1es5aetqva4ymv59qduq1at&st=d0udmb3w&dl=1";

		(SaveGame game, Exception ex, string savePath) = await LoadGameAndData(saveName, SAVES_FOLDER, uri);

		Assert.Null(ex);
		Assert.NotNull(game);
		Assert.True(File.Exists(savePath));

		C7GameData.GameData gd = game.ToGameData(PathUtils.luaRulesDir);
		EngineStorage.InitializeGameDataForTests(gd);

		List<UnitPrototype> protos = gd.unitPrototypes;

		// setup civilizations and players
		var netherlands = gd.civilizations.FirstOrDefault(c => c.name == "Netherlands");
		var nether = new Player() { civilization = netherlands };
		var americans = gd.civilizations.FirstOrDefault(c => c.name == "America");
		var amer = new Player() { civilization = americans };
		var ottomans = gd.civilizations.FirstOrDefault(c => c.name == "Ottomans");
		var otto = new Player() { civilization = ottomans };
		var russians = gd.civilizations.FirstOrDefault(c => c.name == "Russia");
		var rus = new Player() { civilization = russians };
		var koreans = gd.civilizations.FirstOrDefault(c => c.name == "Korea");
		var kor = new Player() { civilization = koreans };
		var hittites = gd.civilizations.FirstOrDefault(c => c.name == "Hittites");
		var hit = new Player() { civilization = hittites };

		// setup resources
		var noResources = new HashSet<Resource>(){};
		var horses = gd.Resources.FirstOrDefault(r => r.Key == "Horses");
		var iron = gd.Resources.FirstOrDefault(r => r.Key == "Iron");
		var saltpeter = gd.Resources.FirstOrDefault(r => r.Key == "Saltpeter");
		var rubber = gd.Resources.FirstOrDefault(r => r.Key == "Rubber");
		var oil = gd.Resources.FirstOrDefault(r => r.Key == "Oil");
		var aluminum = gd.Resources.FirstOrDefault(r => r.Key == "Aluminum");

		// setup techs
		var wheel = gd.techs.FirstOrDefault(t => t.Name == "The Wheel");
		var bronze = gd.techs.FirstOrDefault(t => t.Name == "Bronze Working");
		var horsebackRiding = gd.techs.FirstOrDefault(t => t.Name == "Horseback Riding");
		var mathematics = gd.techs.FirstOrDefault(t => t.Name == "Mathematics");
		var feudalism = gd.techs.FirstOrDefault(t => t.Name == "Feudalism");
		var engineering = gd.techs.FirstOrDefault(t => t.Name == "Engineering");
		var chivalry = gd.techs.FirstOrDefault(t => t.Name == "Chivalry");
		var metallurgy = gd.techs.FirstOrDefault(t => t.Name == "Metallurgy");
		var milTradition = gd.techs.FirstOrDefault(t => t.Name == "Military Tradition");
		var nationalism = gd.techs.FirstOrDefault(t => t.Name == "Nationalism");
		var flight = gd.techs.FirstOrDefault(t => t.Name == "Flight");
		var replaceableParts = gd.techs.FirstOrDefault(t => t.Name == "Replaceable Parts");
		var rocketry = gd.techs.FirstOrDefault(t => t.Name == "Rocketry");

		// setup some unit prototypes
		var spearman = protos.FirstOrDefault(p => p.name == "Spearman");
		var swiss = protos.FirstOrDefault(p => p.name == "Swiss Mercenary");
		var horseman = protos.FirstOrDefault(p => p.name == "Horseman");
		var chariot = protos.FirstOrDefault(p => p.name == "Chariot");
		var threeManChariot = protos.FirstOrDefault(p => p.name == "Three-Man Chariot");
		var knight = protos.FirstOrDefault(p => p.name == "Knight");
		var cavalry = protos.FirstOrDefault(p => p.name == "Cavalry");
		var sipahi = protos.FirstOrDefault(p => p.name == "Sipahi");
		var cossack = protos.FirstOrDefault(p => p.name == "Cossack");
		var catapult = protos.FirstOrDefault(p => p.name == "Catapult");
		var trebuchet = protos.FirstOrDefault(p => p.name == "Trebuchet");
		var cannon = protos.FirstOrDefault(p => p.name == "Cannon");
		var rifleman = protos.FirstOrDefault(p => p.name == "Rifleman");
		var hwacha = protos.FirstOrDefault(p => p.name == "Hwach'a");
		var artillery = protos.FirstOrDefault(p => p.name == "Artillery");
		var fighter = protos.FirstOrDefault(p => p.name == "Fighter");
		var jet = protos.FirstOrDefault(p => p.name == "Jet Fighter");
		var f15 = protos.FirstOrDefault(p => p.name == "F-15");

		#endregion

		#region Ottomans
		Tile ankaraTile = new Tile(ID.None("tile"));
		City ankara = new City(ankaraTile, otto, "Ankara", gd.ids.CreateID("Ankara"));

		Assert.True(horseman.producibleBy.Contains(ottomans));
		Assert.True(knight.producibleBy.Contains(ottomans));
		Assert.True(sipahi.producibleBy.Contains(ottomans));
		Assert.False(cavalry.producibleBy.Contains(ottomans));

		otto.knownTechs.Add(horsebackRiding.id);

		Assert.True(horseman.CanProduce(ankara, new HashSet<Resource>() { horses, iron }));
		Assert.False(knight.CanProduce(ankara, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.False(sipahi.CanProduce(ankara, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.False(cavalry.CanProduce(ankara, new HashSet<Resource>() { horses, iron, saltpeter }));

		otto.knownTechs.Add(chivalry.id);
		// The Horseman is now obsolete
		Assert.False(horseman.CanProduce(ankara, new HashSet<Resource>() { horses, iron }));
		Assert.True(horseman.GetProducibleUpgrade(ankara, new HashSet<Resource>() { horses, iron }) == knight);

		// even though we have saltpeter, we don't know Military Tradition, so the Knight is not yet obsolete
		Assert.True(knight.CanProduce(ankara, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.False(cavalry.CanProduce(ankara, new HashSet<Resource>() { horses, iron, saltpeter }));

		otto.knownTechs.Add(milTradition.id);
		// the Knight is now obsolete
		Assert.False(knight.CanProduce(ankara, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.True(knight.GetProducibleUpgrade(ankara, new HashSet<Resource>() { horses, iron, saltpeter }) == sipahi);
		Assert.False(cavalry.CanProduce(ankara, new HashSet<Resource>() { horses, iron, saltpeter }));

		Assert.True(sipahi.GetProducibleUpgrade(ankara, new HashSet<Resource>() { horses, iron, saltpeter }) == null);
		#endregion

		#region Netherlands
		Tile amsterdamTile = new Tile(ID.None("tile"));
		City amsterdam = new City(amsterdamTile, nether, "Amsterdam", gd.ids.CreateID("Amsterdam"));
        
		Assert.True(spearman.producibleBy.Contains(netherlands));

		nether.knownTechs.Add(bronze.id);
		
		Assert.True(spearman.CanProduce(amsterdam, noResources));

        nether.knownTechs.Add(feudalism.id);
		Assert.True(spearman.CanProduce(amsterdam, noResources));
		Assert.True(swiss.CanProduce(amsterdam, new HashSet<Resource>() { iron }));
        // if the city has iron then we can't produce Spearman anymore, as we can produce their upgrade, the Swiss Mercenary
		Assert.False(spearman.CanProduce(amsterdam, new HashSet<Resource>() { iron }));
        
        nether.knownTechs.Add(nationalism.id);
		Assert.True(rifleman.CanProduce(amsterdam, noResources));
        
        // these 2 currently fail, but they should be passing
		Assert.True(spearman.GetProducibleUpgrade(amsterdam, new HashSet<Resource>() { iron, saltpeter }) == rifleman);
		Assert.False(spearman.GetProducibleUpgrade(amsterdam, new HashSet<Resource>() { iron, saltpeter }) == swiss);
        
		Assert.False(spearman.CanProduce(amsterdam, new HashSet<Resource>() { iron, saltpeter }));
		Assert.False(swiss.CanProduce(amsterdam, new HashSet<Resource>() { iron, saltpeter }));
		Assert.True(swiss.GetProducibleUpgrade(amsterdam, new HashSet<Resource>() { iron, saltpeter }) == rifleman);
		Assert.True(rifleman.CanProduce(amsterdam, new HashSet<Resource>() { iron, saltpeter }));
		#endregion

		#region Hittites
		Tile hattusasTile = new Tile(ID.None("tile"));
		City hattusas = new City(hattusasTile, hit, "Hattusas", gd.ids.CreateID("Hattusas"));

		Assert.True(horseman.producibleBy.Contains(hittites));
		Assert.False(chariot.producibleBy.Contains(hittites));
		Assert.True(threeManChariot.producibleBy.Contains(hittites));


		hit.knownTechs.Add(wheel.id);

		Assert.False(horseman.CanProduce(hattusas, new HashSet<Resource>() { horses }));
		Assert.False(chariot.CanProduce(hattusas, new HashSet<Resource>() { horses }));
		Assert.True(threeManChariot.CanProduce(hattusas, new HashSet<Resource>() { horses }));

		#endregion

		#region Russians
		Tile moscowTile = new Tile(ID.None("tile"));
		City moscow = new City(moscowTile, rus, "Moscow", gd.ids.CreateID("Moscow"));

		Assert.True(horseman.producibleBy.Contains(russians));
		Assert.True(knight.producibleBy.Contains(russians));
		Assert.True(cossack.producibleBy.Contains(russians));
		Assert.False(cavalry.producibleBy.Contains(russians));

		rus.knownTechs.Add(horsebackRiding.id);

		Assert.True(horseman.CanProduce(moscow, new HashSet<Resource>() { horses, iron }));
		Assert.False(knight.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.False(sipahi.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.False(cavalry.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));

		rus.knownTechs.Add(chivalry.id);
		// The Horseman is now obsolete
		Assert.False(horseman.CanProduce(moscow, new HashSet<Resource>() { horses, iron }));
		Assert.True(horseman.GetProducibleUpgrade(moscow, new HashSet<Resource>() { horses, iron }) == knight);

		// even though we have saltpeter, we don't know Military Tradition, so the Knight is not yet obsolete
		Assert.True(knight.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.False(cavalry.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.False(cossack.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));

		rus.knownTechs.Add(milTradition.id);
		// the Knight is now obsolete
		Assert.False(knight.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));
		Assert.True(knight.GetProducibleUpgrade(moscow, new HashSet<Resource>() { horses, iron, saltpeter }) == cossack);
		Assert.False(cavalry.CanProduce(moscow, new HashSet<Resource>() { horses, iron, saltpeter }));

		Assert.True(cossack.GetProducibleUpgrade(moscow, new HashSet<Resource>() { horses, iron, saltpeter }) == null);
		#endregion

		#region Korean
		Tile seoulTile = new Tile(ID.None("tile"));
		City seoul = new City(seoulTile, kor, "Seoul", gd.ids.CreateID("Seoul"));

		// In conquests they both the Cannon and the Hwacha are permitted in the rules for the Koreans,
		// maybe an oversight in the original rules, but it's there.
		// We only care that the Trebuchet can upgrade to a Hwacha and not to a Cannon
		Assert.True(catapult.producibleBy.Contains(koreans));
		Assert.True(trebuchet.producibleBy.Contains(koreans));
		Assert.True(cannon.producibleBy.Contains(koreans));
		Assert.True(hwacha.producibleBy.Contains(koreans));

		Assert.False(catapult.CanProduce(seoul, noResources));
		Assert.False(trebuchet.CanProduce(seoul, noResources));
		Assert.False(cannon.CanProduce(seoul, noResources));
		Assert.False(hwacha.CanProduce(seoul, noResources));

		kor.knownTechs.Add(mathematics.id);

		Assert.True(catapult.CanProduce(seoul, noResources));
		Assert.False(trebuchet.CanProduce(seoul, noResources));
		Assert.False(cannon.CanProduce(seoul, noResources));
		Assert.False(hwacha.CanProduce(seoul, noResources));

		kor.knownTechs.Add(engineering.id);

		Assert.False(catapult.CanProduce(seoul, noResources));
		Assert.True(trebuchet.CanProduce(seoul, noResources));
		Assert.False(cannon.CanProduce(seoul, noResources));
		Assert.False(hwacha.CanProduce(seoul, noResources));

		Assert.True(catapult.GetProducibleUpgrade(seoul, noResources) == trebuchet);

		kor.knownTechs.Add(metallurgy.id);

		Assert.False(catapult.CanProduce(seoul, new HashSet<Resource>() { saltpeter }));
		Assert.False(trebuchet.CanProduce(seoul, new HashSet<Resource>() { saltpeter }));
		Assert.False(cannon.CanProduce(seoul, new HashSet<Resource>() { saltpeter }));
		Assert.True(hwacha.CanProduce(seoul, new HashSet<Resource>() { saltpeter }));

		Assert.True(trebuchet.GetProducibleUpgrade(seoul, new HashSet<Resource>() { saltpeter }) == hwacha);
		Assert.False(trebuchet.GetProducibleUpgrade(seoul, new HashSet<Resource>() { saltpeter }) == cannon);

		kor.knownTechs.Add(replaceableParts.id);
		Assert.False(hwacha.CanProduce(seoul, new HashSet<Resource>() { saltpeter }));
		Assert.True(artillery.CanProduce(seoul, new HashSet<Resource>() { }));
		Assert.True(hwacha.GetProducibleUpgrade(seoul, new HashSet<Resource>() { }) == artillery);

		#endregion

		#region America
		Tile seattleTile = new Tile(ID.None("tile"));
		City seattle = new City(seattleTile, amer, "Seattle", gd.ids.CreateID("Seattle"));

		Assert.True(fighter.producibleBy.Contains(americans));
		Assert.False(jet.producibleBy.Contains(americans));
		Assert.True(f15.producibleBy.Contains(americans));

		Assert.False(fighter.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));
		Assert.False(jet.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));
		Assert.False(f15.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));

		amer.knownTechs.Add(flight.id);
		Assert.True(fighter.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));
		Assert.False(jet.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));
		Assert.False(f15.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));

		Assert.True(fighter.GetProducibleUpgrade(seattle, new HashSet<Resource>() { oil, aluminum }) == null);

		amer.knownTechs.Add(rocketry.id);
		Assert.True(fighter.CanProduce(seattle, new HashSet<Resource>() { oil }));
		Assert.False(fighter.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));
		Assert.False(jet.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));
		Assert.True(f15.CanProduce(seattle, new HashSet<Resource>() { oil, aluminum }));

		Assert.True(fighter.GetProducibleUpgrade(seattle, new HashSet<Resource>() { oil, aluminum }) == f15);
		#endregion
	}
</details>

Regarding the .JSON-test UnitAvailability_JSON update, once we have settled on a stable solution, I will update the json on Dropbox, and give you the new link to update the unit test, if that's ok.

>I'm intentionally changing the field from upgradeTo to upgradesTo

This is fine as far as I am concerned, although old saves would still fail since one is a string and the other a list

>We should maybe consider a save file version number at some point

Perhaps it's too early since the scheme is constantly changing and we don't really care about backward compatibility at this point.
It would only be helpful as a point of reference, we could start bumping the minor version every time we update the json.
If you feel like it, do the honors of adding the first 0.0.0001 (or something) version

ajhalme:
Ok, looking pretty good now. I've added the Netherlands test case, and the upgrade tests are all green.

After careful consideration, I've concluded that "Musketman -> Swiss Mercenary" has to be interpreted as a "reverse upgrade rule". Using a _shield value heuristic_, I've implemented some logic that detects cases where the relation should be applied in reverse . There's only 7 such cases (6 unique) among the 141 upgrade rules in a regular Civ3 game save, and only one of them has an impact on the upgrade chain. And that's the Swiss Merc causing trouble.

The shield value heuristic is simply that if you are declaring an upgrade `A --> B`, but `A.shieldCost > B.shieldCost`, then we apply `B --> A` instead. It might be possible to replace this heuristic with one that considers tech tree dependencies and resources and things, but the shield scheme feels very reasonable to me. In-game, I'd expect the upgrade cost to be a function of the shield cost difference anyway.

I reorganised the unit upgrade code with more of the civ/city/resources-parameterised stuff in the unitPrototype class. Also annotated and slightly renamed some of the upgrade functions.

And then applied the new logic to the save file and the base ruleset: the Spearman is now correctly annotated to upgrade to Pikeman and Swiss Merc.

stavrosfa:
Hmm.. I didn't realize that you were going to implement something right away, so I was expecting some discussion to happen first.

The swiss merc upgrading to a musketman is wrong. Check out this thread as well:

https://forums.civfanatics.com/threads/swiss-mercenary-and-upgrades.70049/post-1391927

I don't think shields are a good indicator, because, at least in my mind, it takes away any advantage a player has with a "unique" unit. By forcing the Swiss merc to upgrade (or disallow production after gunpowder) to a unit like muskteman which has the same defense, but at double the shield cost, takes away the whole unique unit advantage, and only a couple techs after being able to build it. If the player has built the Great Library, it could go away even faster.

Also in the rules there is no such chain.

It goes like
Pikeman -> Musketman -> Swiss Merc -> Musketeer -> Rifleman

So it's even defined in there that Swiss Merc > Musketman in a sense, and of course only for the Dutch. French get a similar advantage with the Musketeer, and all end up having Riflemen which is better than all of the others obviously.

Note that the Dutch can still build a Musketman if they don't have iron!

I think that we need to somehow keep track of these chains per civ, besides the practical reasons of knowing what upgrades to what, but also, to be able to calculate upgrade costs. 
For example, if we as Dutch want to pay to upgrade a Swiss Merc to a Rifleman, I think it would cost 30 gold, but if the Swiss Merc upgrades first to a Musketman, then it would be 60 gold because we'd have to pay for 2 upgade-steps worth of gold.
I don't have any hard evidence for this, I should check this out too, but I think you 'll get the idea I am trying to convey.

I think these should be passing as well

// add somewhere

```
var gunpowder = gd.techs.FirstOrDefault(t => t.Name == "Gunpowder");
```

and 

```
var musketman = protos.FirstOrDefault(p => p.name == "Musketman");
```

// **after** feudalism checks but **before** nationalism checks

```
nether.knownTechs.Add(gunpowder.id);
Assert.True(swiss.CanProduce(amsterdam, new HashSet<Resource>() { iron }));
Assert.False(swiss.CanProduce(amsterdam, new HashSet<Resource>() { saltpeter }));
Assert.True(swiss.CanProduce(amsterdam, new HashSet<Resource>() { iron, saltpeter }));
Assert.True(musketman.CanProduce(amsterdam, new HashSet<Resource>() { saltpeter }));
Assert.False(musketman.CanProduce(amsterdam, new HashSet<Resource>() { iron, saltpeter }));
Assert.False(rifleman.CanProduce(amsterdam, new HashSet<Resource>() { iron, saltpeter }));
      
Assert.Null(swiss.GetProducibleUpgrade(amsterdam, new HashSet<Resource>() { iron, saltpeter }));
Assert.False(swiss.GetProducibleUpgrade(amsterdam, new HashSet<Resource>() { iron, saltpeter }) == musketman);
```

Do let me know if you want to discuss this more, I am keen on doing this right. We can continue here, or send a message on discord.

ajhalme:
Ha, didn't even cross my mind that `Musketman -> Swiss` would be _correct_.

I was somewhat checking my work against this site (top search hit), which has the Swiss upgrading to Musketman: https://civilization.fandom.com/wiki/Swiss_Mercenary_(Civ3)

But your explanation makes sense. Here's another reference:
https://forums.civfanatics.com/threads/swiss-mercenary-problem.68848/

More than happy to rip out the heuristic and leave the pure and much more elegant parsing logic in place!

Added the new Swiss test cases, which ran green without any modifications to the upgrade logic.

I think the upgrade cost calculation is probably best left to another PR.

--

The reverse upgrade idea, for posterity:

```
		private static bool UnusualUpgradeRelation(UnitNode a, UnitNode b) {
			// Shield value heuristic: higher cost -> lower cost indicates reverse relation
			return a.shieldCost > (b?.shieldCost ?? int.MaxValue);
		}

		/// <summary>
		/// Manipulate the unit upgrade linked lists to apply an upgrade relation in reverse.
		/// </summary>
		private static void AddReverseUpgrade(UnitNode a, UnitNode b) {
			// From:	X -> a -> b -> Y
			// To:		X -> b -> a -> Y

			b.next?.prev.Remove(b);
			a.next = b.next;
			b.next?.prev.Add(a);

			b.next = a;
			foreach (UnitNode aPrev in a.prev) {
				aPrev.next = b;
				b.prev.Add(aPrev);
			}
			a.prev.Clear();
			a.prev.Add(b);
		}
```

stavrosfa:
This seems to work just fine now, thanks!
The json save file also seems to agree with our base json file now, except for a few cases where a `'` is present

>I think the upgrade cost calculation is probably best left to another PR.

Yeah, I was only mentioning it, because it will come up at some point

I have updated the unit test to include some more tests, and also updated the json file link.

Feel free to add more tests cases if you feel like it's appropriate.

(I tried to copy paste the contents here, but it was a mess, so I just attached the file)

[UnitPrototypeTest.cs](https://github.com/user-attachments/files/28970230/UnitPrototypeTest.cs)

ajhalme:
Patched the test file - thanks for putting together all the test cases! I think we have more than plenty for now.

> a few cases where a `'` is present

Not quite sure what this is about. Some encoding issue? Is it a blocker here?

stavrosfa:
>Not quite sure what this is about. Some encoding issue? Is it a blocker here?

Sorry, I thought it was something you had mentioned earlier, but I was wrong.

It's not a blocker in any case.

I feel this is an improvement on the current system, I haven't seen anything out of palce so far, so feel free to merge if there is nothing else from your side.

ajhalme:
Oh, that was just the city names and things: "Leonardo's Workshop" had the `'` as a unicode point. It doesn't seem to need to be that way, and the `jq` tool I used to make the "stable" just uses the regular character, so this will change those across the board.

I'll merge this in as is.

PR #938, opened 2026-05-25
Unify Lua modding under GameMode (Redux)

ajhalme:
Attempt to merge @esbudylin's Lua consolidation effort on top of latest `Development` branch – a continuation of #906.

My plan was to take all the new abstractions, but leave the Lua code where it is for now, but I'm not sure that's viable. If there's appetite, we could just go for the whole Lua migration and cleanup afterwards.

All the PediaIcons code looks really shaky. I couldn't immediately see how to fix Standalone mode `Load Game`, but `New Game` seems to work at least.


stavrosfa:
To be totally honest, I don't feel really comfortable with this.

I think we should first ask @esbudylin if they intend to finish this, (or it's done and just needs some documentation or something), and then we can see our options.

My biggest fear is that this is a huge PR, I haven't really looked into it, because I was expecting the extra documentation Yegor mentioned, and even if I do, I am sure we can expect some disruptions that are natural with this type of changes, which will take a lot of time to debug, especially without the support of the original author.

If the codebase was in a better and more stable state than it currently is, I would be more inclined to plow through this.

But, even with my newer commits, as well as with the "older" codebase, there are major bugs (old and introduced) that are in constant need to be solved and refined, and I feel like we should be putting our efforts there (and this is what I am trying to do), or at least, not modify a huge chunk of the current codebase.

@WildWeazel I would like your input on this too please

WildWeazel:
Popping in without context yet- has Yegor talked about where he was going with this, or was there discussion about how to wrap it up? And is this currently blocking other progress? The mod interface is one of those things I want to make sure we "get right" since it will be such a pivotal part for other people to build on, so in general I'd favor taking time to get a consensus, even from non-dev modders at times. I'll try to catch up on this over the weekend.

esbudylin:
Hi everyone!

When I opened the original PR, I intended to return to it for testing and minor polishing (adding comments and maybe renaming a few methods here and there). Unfortunately I ran out of time and energy for the project very shortly after submitting these changes. I don't think I'll be able to return to them in the upcoming months.

It's obviously a big PR, but most of the changes are caused by reorganizing the Lua files with minimal changes to their content. I think in terms of ease of maintenance the project is better off with these changes, since their goal is to reduce the number of systems responsible for Lua integration. However, I likely won't be able to help if anything goes wrong, so the decision is obviously up to active contributors.

ajhalme:
Thanks Yegor - for the original PR and otherwise. Hope to see you contributing again sometime in the future!

I like the design here, it's definitely an improvement. I'm confident we can refine it into something we are all happy to proceed with. I agree that this isn't as big as it might look at first glance: reorganisation always leads to some extra noise in PRs. The load pattern is also sufficiently fundamental that I feel this is one of those cases where it either works perfectly or it doesn't work at all -- if it doesn't break immediately, it's probably fine.

As for blocking progress, maybe not immediately, but it would be nice to have a more stable base and path forward for mods. I can see some more substantial changes like the animation system, tile map stuff, Mac builds, etc., also benefitting from this indirectly.

Plus, if we can get OC3 to support custom modding, that might give the project a nice boost. I've been toying with the idea of a simple mod I'd like to try to make as an OC3 demo. (A chess theme!)

Also, I'll also note that the standalone Load Game issue I mentioned earlier seems to already be present in Development. This PR doesn't introduce it.

WildWeazel:
Good to hear from you @esbudylin and thank you again for all of your hard work! We wouldn't have gotten Carthage out the door without you. I'm very familiar with running out of time and energy to keep up. Know that your contributions are recognized and appreciated, and we look forward to you rejoining us as life allows.

stavrosfa:
Sorry to see you take a break (hopefully it's only that) @esbudylin.

You were a great help since I joined the project (and here I must mention @TomWerner as well), and your contributions to the project is being actively appreciated. 

I really do hope to see you make a come back when you feel is right!

WildWeazel:
Narrator: he did not, in fact, catch up that weekend.

stavrosfa:
For some reason I read that with Morgan Freeman's voice in my head 😅

WildWeazel:
I appreciate the gravitas, but this is clearly a Ron Howard setting

ajhalme:
The essence of this PR really is this shift below and the resulting re-wiring. A move from an explicit file-oriented model to one where the primary unit of (mod) customisation is the directory. Clears up some of the related file path lookups in favour of a more centralised approach. 

```csharp
GameMode.Config basic = new("civ3");
GameMode.Config standalone = new("civ3", ["standalone"]);
// GameModeConfig basic = new("base-ruleset.json");
// GameModeConfig standalone = new("base-ruleset.json", ["standalone.lua"]);
```

WildWeazel:
That much makes sense to me at least. It's more analogous to the scenario search path, and better supports "modular mods" where you can choose a whole directory/archive to load. And I have no issue with continuing to make people put expected/required files in a specific location, if that's the implication.

ajhalme:
I gather @WildWeazel is somewhat supportive of this PR, is that right?

@stavrosfa - any additional thoughts on this?

If I type up some documentation on what we have here and what the direction might be, would you be open to this getting merged? I'm happy to wait for major in-flight branches to land before we push forward with this, if that's the concern.

That said, I'd rather we make a firm Yes/No decision on this sooner rather than later, so we don't have to drag these branches around.

stavrosfa:
I am not against it, and I agree that sooner is better than later.

Now that we have finished merging some stuff, we can merge this one too.

I don't expect to open any new PR's at the moment, but thanks for asking.This will also give me some time to play around with it a bit once it's ready.

Some documentation would be nice for sure, not just for us, but for the modders as well.

ajhalme:
Claude did a pretty good job with a game mode README. I cleaned it up and added it in as `/C7/Lua/README.GameModes.md`

Here's an offcut comparing old scheme and new:

## Old vs. new

| | Old (file-based) | New (directory-based) |
|---|---|---|
| Unit of customization | an individual file | a directory |
| Config shape | `GameModeConfig(string baseModePath, List<string> addonPaths)` — paths to specific files | `GameMode.Config(string baseModeDir, List<string> addonPaths)` — directory names |
| Ruleset addons | ✅ supported | ✅ supported |
| Texture addons | ❌ not supported — textures always loaded from one hardcoded file (`texture_configs/civ3.lua` or `c7.lua`) | ✅ supported |
| Behavior/rule addons | ❌ not supported — rules always loaded from one hardcoded file (`rules/civ3.lua`) | ✅ supported |
| Loaders | three independent entry points: `GameModeLoader.Load` (ruleset), `TextureLoader.SetConfig(dir, file)` (textures), `RulesEngine.Initialize(dir, file)` (behaviors) | one entry point: `GameMode.Load(gameModesDir, config)` |
| Save-game awareness | ❌ — reloading a save always re-derived the ruleset from a hardcoded path (long-standing `TODO` in the code) | ✅ — a save stores the `GameMode.Config` it was created with, so reloading reconstructs the exact same mode/addon stack |

ajhalme:
I had a go at rebasing my changes and the merges I did in order to have a clean commit on top of the original draft, but it got really tricky really quickly replaying all the changes that have happened on main. For sizeable, long-lived feature branches like this one it's useful to have a clean revert target, so I've squash-merged the whole thing on top of main.

This sadly discards the original attribution, but let at least this record show that **Yegor did all the heavy lifting here**. Thanks again @esbudylin.

Issue #868, opened 2026-02-07
Feature Request - HiDPI Support


oneguynick:
**Problem:** The text and UI elements do not support scaling. It makes it incredibly difficult to play.

**Version:** OpenCiv3 0.3 Dutch

**Troubleshooting:**

- CLI Run: ./OpenCiv3.x86_64 -f --display-driver wayland --delta-smothing enable
- Assets: Tried both bulit-in and Civ3 assets

**Env:**

- Linux 6.18
- Wayland
- KDE Plasma 6.5
- Archlinux

<img width="1575" height="1473" alt="Image" src="https://github.com/user-attachments/assets/ace4236a-22db-4bf7-92ec-08e0b3ccb41b" />
<img width="958" height="2396" alt="Image" src="https://github.com/user-attachments/assets/07f87c81-5a16-4f51-9de0-bf8ad95d9e68" />

PR #939, opened 2026-05-30
Set HiDPI=false in Godot project settings

### LINKED ISSUE #876: When in full-screen mode, each component does not fit properly[Mac M1]
## Issue
When in full-screen mode, each component doesn't fit properly, making it very difficult to see and operate.
## Evidence
Image

## Enviroment
---------------------------------
OS: macOS 15.3.2 24D81 arm64
Host: MacBookPro17,1
Kernel: 24.3.0
Uptime: 11 days, 12 hours, 13 mins
Packages: 366 (brew)
Shell: zsh 5.9
Resolution: 1440x900
DE: Aqua
WM: Quartz Compositor
WM Theme: Blue (Dark)
Terminal: ghostty
CPU: Apple M1
GPU: Apple M1
Memory: 2639MiB / 16384MiB

ajhalme:
I picked up a second hand 2022 MacBook Air the other day. The laptop boasts a high-DPI "Liquid Retina" display: _"13.6-inch (diagonal) LED-backlit display with IPS technology; 2560x1664 native resolution at 224 pixels per inch"_

OpenCiv3's Godot project settings for displays default to **high-DPI being enabled**, which when debugging in Godot (via Rider) results in a tiny game window, as the pixels render in true size. Setting high-DPI off gives a more useful game window, as Godot will scale things up in a reasonable way. (There are additional scaling settings to tweak, but just this one change is a big help.)

I suspect this fixes #876 and possibly #868 as well, though confirming this would require a release build or more testing with different hardware. My guess is that this flag will determine the `NSHighResolutionCapable` flag on Mac builds -- see workaround https://github.com/C7-Game/OpenCiv3/issues/876#issuecomment-3902350086 -- but to confirm I'll first have to figure out how Mac builds work.

### Before
_Compare Rider's menu text and OpenCiv3 status box text sizes._
Screenshot 2026-05-30 at 01 13 30

### After  
Screenshot 2026-05-30 at 00 52 54

### DIFF
--- a/C7/project.godot
+++ b/C7/project.godot
@@ -26,6 +26,7 @@ LogManager="*res://LogManager.cs"
 
 window/size/window_width_override=1152
 window/size/window_height_override=768
+window/dpi/allow_hidpi=false
 window/per_pixel_transparency/allowed=true
 
 [dotnet]


stavrosfa:
I can't verify this, if you feel this is an overall improvement and doesn't break anything else, feel free to merge this.

It's a small enough change to track anyway

ajhalme:
Looking at the build output binary of a local OpenCiv3 Mac build with an without `window/dpi/allow_hidpi=false` set, my understanding is that Godot does NOT set the `NSHighResolutionCapable`  in the app's `Info.plist` to `false` on `allow_hidpi=false`.

As such, with this patch, I believe Godot itself does the scaling at runtime.

While this fix doesn't work in the expected way, it does work - the `allow_hidpi=false` setting makes the built app render in a more useful way, so I'm merging this in.

Godot has additional scaling settings we can explore another time. These include the scaling mode and integer scaling vs fractional scaling - the configuration after this patch is the best one of the ones I tried. Also worth noting that more recent versions of Godot change the scaling defaults, so we may be forced to revisit this in the future as we upgrade Godot versions.

I'll be adding references to this PR and the reported issue on this new wiki page: https://github.com/C7-Game/OpenCiv3/wiki/MacOS-Support

PR #940, opened 2026-06-04
Go-to caching

stavrosfa:
There was a bug at some point I was trying to figure out, and I realized that we keep calculating the same path over and over again, which made my debugging efforts difficult.

This PR basically says that, if the path is the same as the current one, don't recalculate it.

### DIFF
--- a/C7Engine/AI/Pathing/AStarAlgorithm.cs
+++ b/C7Engine/AI/Pathing/AStarAlgorithm.cs
@@ -38,9 +38,9 @@ namespace C7Engine.Pathing {
 			this.isPassable = isPassable;
 		}
 
-        private Tile lastStartTile = null;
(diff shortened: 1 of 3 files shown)


C7Engine/AI/Pathing/AStarAlgorithm.cs
```diff
@@ -38,6 +38,10 @@ public AStarAlgorithm(EdgeWalker<Tile> edgeWalker, Func<Tile, Tile, double> cost
 			this.isPassable = isPassable;
 		}
 
+		private Tile lastStartTile = null;
+		private Tile lastDestTile = null;
+		private Tile lastNeighTile = null;
```
ajhalme, line 43: The end is n(e)igh!

ajhalme (approved):
Seems very reasonable

PR #942, opened 2026-06-18
Some enhancements on unit movement on railroads

stavrosfa:
Added the "0" text and disabled animations when unit moves on railroad
Added some helper methods to support the above

### DIFF
--- a/C7/Map/GotoLayer.cs
+++ b/C7/Map/GotoLayer.cs
@@ -54,8 +54,7 @@ public partial class GotoLayer : LooseLayer {
 		staticCursorRect.Position = position - new Vector2(staticCursor.GetWidth(), staticCursor.GetHeight()) / 2;
 
 		gotoLabel.Theme = attackingMove ? redFontTheme : whiteFontTheme;
-		gotoLabel.Text = moves > 0 ? moves.ToString() : " ";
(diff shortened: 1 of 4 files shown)


C7/Map/GotoLayer.cs
```diff
@@ -55,6 +55,7 @@ public void DrawStaticGoToCursor(LooseView looseView, Vector2 position, int move
 
 		gotoLabel.Theme = attackingMove ? redFontTheme : whiteFontTheme;
 		gotoLabel.Text = moves > 0 ? moves.ToString() : " ";
+		gotoLabel.Text = moves == 0 && !attackingMove ? moves.ToString() : gotoLabel.Text;
```
ajhalme: This L57-L58 combo looks a bit curious - maybe needs a function to resolve the text?

ajhalme (approved):
Looks reasonable

C7/Map/GotoLayer.cs
stavrosfa: Yeah, it was done kind of out of frustration, I have now simplified it.

stavrosfa:
I am merging this now that I have the time, since it's a very small change.
If you have further comments, let me know, and if changes are needed I will address them accordingly.

PR #944, opened 2026-06-19
AI tweaks to civ economy management

ajhalme:
A few tweaks to AI handling of the economy:
- Adjust production choices so Wealth is picked when the economy is in a bad shape
- Let AI run a deficit on some turns, if the economy can tolerate it
- Fix bug in finance update, let sliders fall all the way to try and keep the civ in the game

With these fixes I regularly get tiny pangea games to 2000AD and beyond without anyone crashing their economy.


stavrosfa (approved):
I ddon't really have any comments on this. 
We can do 100 different things to improve the AI, every step that makes the experience better is a win in my book.

PR #945, opened 2026-06-20
Enhancement/worker actions

stavrosfa:
1. Added a confirmation popup whenever a worker that is working a job to verify we want to stop them from working anymore.
2. "This unit has already moved" message is now unified and using the MsgShowTemporaryPopup message.
3. Workers now contribute their efforts at the end of the turn instead of right away, except when the terraforming action can be completed on the same turn. 
For example, if 2 workers are assigned on turn 10 to work on a road (0 terraforming progress), and a third one is assigned to road the tile as well, the road will be completed. Same with a mine and 6 workers, etc..
4. Unified the HandleUnitSelection method code that was duplicated in Game and RightClickMenu classes.
5. isWorker in UnitPrototype was refined

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -603,8 +603,7 @@ public partial class Game : Node {
 			});
 		}
 	}
-
-	// TODO: unify same parts with RightClickTileMenu->SelectUnit()
+    
 	private void HandleUnitSelection(InputEventMouseButton eventMouseButton) {
 		Tile tile = PositionToTile(eventMouseButton.Position);
 		if (tile == null) {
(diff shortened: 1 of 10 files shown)


ajhalme (approved):
Looks reasonable

stavrosfa:
Hey @ajhalme, I had to resolve some conflicts in `Game.cs`. Could you please have a quick look at HandleUnitSelectionTileClick() and the 2 SelectUnit() methods to make sure I didn't break anything?

ajhalme (approved):
Seems fine to me.

The `SetSelectedUnit` signature in `UnitSelector` is a bit curious. Instead of "whether the selected unit has remaining moves", I'd have the call return "whether target unit was successfully selected". And then one could make a separate call for available moves as needed.

As it stands, making the extra `unit.WorkerJob != null` check (to effectively determine whether the worker was selected) right after the call in `Game.cs` is awkward, but seems functional enough.

stavrosfa:
Yeah I hear you, it's a bit clanky, but it should be revisited separately imo.

`unit.WorkerJob != null` could also mean a Crusader that is building a fortress (not implemented yet), and effectively any moddable unit that terraforms anything and we don't want to stop it without confirmation.

PR #946, opened 2026-06-20
RightClickMenu enhancements

stavrosfa:
Added a menu separator as is in the original game, cleaned up the text and the order of items.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -638,13 +638,15 @@ public partial class Game : Node {
 	private void HandleRightClickOnTile(Tile tile, InputEventMouseButton eventMouseButton) {
 		bool shiftDown = Input.IsKeyPressed(Godot.Key.Shift);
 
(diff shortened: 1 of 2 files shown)


ajhalme:
Can't unsee the wavy justification relative to the text (OC3 vs Civ3):

menus

stavrosfa:
Yeah, I know, I tried to fix this, but as far as I could tell it needs a more holistic refactor to be able to allign the text buttons.
I tried to add simple spaces before and after the text to simulate a margin, but it felt like a very cheap solution.

ajhalme (approved):

stavrosfa:
I somehow stumbled on the solution by mistake, looking for something else.
Turns out the buttons already use a Style Box, and we can specify the margin there.

stavrosfa:
Unless I find a bug, I am done with this for now, let me know if it looks ok so it can be merged, thanks!

ajhalme (approved):
LGTM!

PR #947, opened 2026-06-21
Locked Alliances implementation

stavrosfa:
1. Implemented the locked alliance/war mechanic.
2. Players in locked alliance/war now are aware of each other, and are unable to make peace or war with their enemies/allies respectively.
3. There is no option in the editor to say "USA has a relationship/knows about the existance of China" for example. The alliances produce a "Mutual Protection Pact" agreement by default as far as I could tell, but I think embassies are how players know of each other.
4. Replaced "Declare war" entry in diplomacy screen when talking to an ally to "Trade World maps". This is not replaced when players are not at war, only when in locked peace.
5. There some random minor code cleanup as per usual
6. In test I added "EngineStorage.animationsEnabled = false;" because there was a bug and the test looped infinitely (we should keep this in mind for future tests too, or somehow disable animation entirely for every unit test)

### DIFF
--- a/C7Engine/C7GameData/ImportCiv3.cs
+++ b/C7Engine/C7GameData/ImportCiv3.cs
@@ -213,6 +213,8 @@ namespace C7GameData {
 			ImportBicLeaders();
 			ImportBicUnits();
 			ImportBicCities();
+			ImportEmbassies();
+			ImportAlliances();
 
 			SetMapDimensions(biq, save);
 			SetWorldWrap(biq, save);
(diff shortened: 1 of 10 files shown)


C7Engine/C7GameData/Save/SaveGame.cs
```diff
@@ -409,6 +431,8 @@ private void ConvertCultureGroups(GameData data) {
 		public string DefaultExperienceLevel; // key
 		public List<Civilization> Civilizations = new List<Civilization>();
 		public HashSet<CultureGroup> CultureGroups = new HashSet<CultureGroup>();
+		public HashSet<Alliance> Alliances;
```
ajhalme: Would it be a less of a hassle overall, if `Alliances` was always initialised (even for Barbs) as an empty collection?

C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -886,6 +888,106 @@ private SavePlayer MakeSavePlayerFromCiv(Civilization civ, bool isHuman, string
...
+					};
+					playerWithEmbassy.playerRelationships.Add(other.id.ToString(), pr);
+
+					// simple peace deal, that could be removed if in a locked war (check ImportAlliances())
+					pr.multiTurnDeals.Add(MultiTurnDeal.DEFAULT_PEACE);
+				}
+			}
+		}
+
+		private void ImportAlliances() {
```
ajhalme, line 927: This method is doing many things and relies on quite subtle indexing to get stuff out of the Biq and into game data. I feel like the data structures here could do a bit more to help with the heavy lifting.

I can even see how an `AllianceManager` class could work and be a nice way to contain all of the relevant logic in one place. Starting to get a bit concerned with how busy some of these key classes are getting.

But it is what it is.

ajhalme:
> In test I added "EngineStorage.animationsEnabled = false;" because there was a bug and the test looped infinitely (we should keep this in mind for future tests too, or somehow disable animation entirely for every unit test)

Well this sounds familiar! :)

Disabling animations for unit tests is reasonable I suppose.

C7Engine/C7GameData/Save/SaveGame.cs
stavrosfa: I guess, to be honest I can't remember why, but I did have a reason for doing this. I changed it as suggested, doesn't seem to have any negative effect. 
I will update it.

C7Engine/C7GameData/ImportCiv3.cs
stavrosfa: I was thinking the same thing ealier, that the class is getting bigger and bigger, but at the same time, the biq/sav data we are parsing here are 25 years old, they aren't really going to change.

I could break up the method to a bunch of more contained methods, but I really don't see the benefit, it's about 60 lines of code, and broken up methods would make the back and forth a bit tedious..

Unless you mean something different from what I understood, let me know.

stavrosfa:
>Well this sounds familiar! :)

Has this happened before? I wouldn't be surprised 😅

ajhalme:
> > Well this sounds familiar! :)
> 
> Has this happened before? I wouldn't be surprised 😅

https://github.com/C7-Game/OpenCiv3/pull/935

I came across a load issue in my bombard mechanic testing that I worked around with `animationsEnabled=false`, but you reorganised the unit load sequence such that this issue did not manifest at game load.

stavrosfa:
Oh ok, makes sense. 
I think though that this was just a side effect rather than the actual culprit. This also affected runtime, this one only affects the unit tests because they are being run headless.

C7Engine/C7GameData/ImportCiv3.cs
ajhalme: It's fine as is. I just had to read through the function multiple times to figure out what it was doing, only to find out that it's doing something quite simple.

ajhalme (approved):

C7Engine/C7GameData/ImportCiv3.cs
stavrosfa: I tried to leave some section comments to be easy to follow the high level logic.

The biq/sav sections are pretty weird sometimes, and I have spent more time than I would want to admit trying to figure them out, that's why I skip many of the details as to why it looks the way it does. But at the same time, I am trying to be as straight forward and verbose as possible and make it a step by step process, for myself as well as for others.

ajhalme (approved):

ajhalme:
So, it turns out that GOG doesn't include "2 MP Rise of Rome.biq" with Civ 3 Complete.

https://forums.civfanatics.com/threads/rise-of-rome-scenario-missing-an-explanation.678919/

Meaning that `CheckScenariosInCiv3Subfolder` in `LoadAllConquestScenarios` in `SaveTests` started throwing for me: expected 8, found 7.

I have the Steam version as well, so was able to source the missing `.biq`, but, idk, maybe we need to exclude Rise of Rome from the test.

stavrosfa:
This is fine by me.
We could comment it out, leave a comment explaining what's going on, and we should be good.

PR #948, opened 2026-06-21
Unit movement point fix and HP enhancements

stavrosfa:
1. Fixes the problem where in the beginning of a new game a scout for example only had 1 move left instead of 2. This was the fault of how I was initializing the data in `Game.cs`
2. Added bonus HP for all units. So now, a veteran Ancient Cavalry will have 5 hp instead of 4, an elite AC 6, etc..
3. Now barbarians spawn with the appropriate HP (which is configurable in the json)
4. Changed the default spawning direction to South
5. The HP bar for conscript units will go from green to red, instead of from green to orange.


ajhalme (approved):
This all looks reasonable

PR #949, opened 2026-06-26
Patch SaveTest to account for missing "2 Rise of Rome.biq"

ajhalme:
Scenario file "2 Rise of Rome.biq" is missing from the Conquests folder in some recent releases of the game.

See: https://forums.civfanatics.com/threads/rise-of-rome-scenario-missing-an-explanation.678919/

With this patch we check the wayward scenario file separately, so that its absence only skips one test, instead of failing (or requiring us to silently tolerate missing files in) LoadAllConquestScenarios as a whole

As discussed in https://github.com/C7-Game/OpenCiv3/pull/947#issuecomment-4805349439


stavrosfa (approved):
Works as expected, thanks!

PR #950, opened 2026-06-29
Civ & Unit loading fixes

stavrosfa:
Fixed not loading civs in custom scenarios that are not part of the game.

Fixed barbarian units not loading in .biq files, because their tribe index was out of range. In .sav files the problem doesn't seem to exist.

### DIFF
--- a/C7Engine/C7GameData/ImportCiv3.cs
+++ b/C7Engine/C7GameData/ImportCiv3.cs
@@ -940,7 +940,8 @@ namespace C7GameData {
 
 			// import player alliances
 			for (int i = 0; i < theBiq.GameAlliance[0].Length; i++) {
-				save.Players.Where(p => p.isIncludedInGame || p.isBarbarian).ToList()[i + 1].alliance = alliances.FirstOrDefault(a => a.index == theBiq.GameAlliance[0][i])?.name;
+				var filteredPlayers = save.Players.Where(p => p.isIncludedInGame || p.isBarbarian).ToList();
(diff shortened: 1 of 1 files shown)


ajhalme (approved):
`.ToList()[i + 1]` is a bit grim. Maybe init the filtered civ collection outside the loop?

Which `.biq` is choking on barb tribes?

stavrosfa:
It's a custom .biq file I created to test a few things.

Apparentrly I can't attach it here, I am sending this over on discord (https://discord.com/channels/909680526498082837/909857733388406844/1521306795241504918)

Without the fix, the .biq, although valid, doesn't load at all. You will also notice the barbarians in the middle, without the fix, we skip adding them, because their tribe index is larger than the number of players.

Btw, tribes are not implemented at all.

ajhalme:
LGTM -- Nice tester, patch fixes the .biq load issue for me.

--

The compiler will probably optimise it away anyway, but this is what I had in mind:
```
	var filteredPlayers = save.Players.Where(p => p.isIncludedInGame || p.isBarbarian).ToList();
	for (int i = 0; i < theBiq.GameAlliance[0].Length; i++) {			
		filteredPlayers[i + 1].alliance = alliances.FirstOrDefault(a => a.index == theBiq.GameAlliance[0][i])?.name;
	}
```

ajhalme (approved):

stavrosfa:
Ah, yeah, makes sense, sorry, it's been a long day 😅 

Keep that .biq handy, I 'll be pushing some pacthes in the next few days for various things, I built this as I went, wanting to test every possible case I could think of, that's why it's set up like that.

ajhalme (approved):

PR #951, opened 2026-07-02
Intent based movement

stavrosfa:
Introducing an intent behind movement to be able to determine movement better.
Various small fixes.

The `CanEnter()` method was getting progressively messier and it was hard to read, so I took a shot at refactoring the whole thing.

Another reason behind this commit, was to synchronise mouse based go-to movement and the keyboard key strokes movement. They behaved very differently, as the keyboard based movement was lacking half the functionality.

I didn't mean to remove the TileProbe class, but as the refactoring was progressing, it came as a natural thing to do.

The AI (in observer mode) seems to behave ok with this too, the can move, declare war, fight and take cities, at the same rate as before, so I don't think I broke someting.

I restored the correct tecch box pop up that was altered some time ago, and made the font smaller and not so "offensive". I don't know how well this will play in other resolutions, I can't test it to be honest.

Let me know what you think!

### DIFF
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -825,11 +825,11 @@ namespace C7GameData {
 			if (this.IsLandUnit() && !tile.IsLand())
 				return Intent.Disabled;
 
-			var hasOtherOwnersCity = HasOtherOwnersCity(tile, unitOwner);
+			var hasForeignCity = HasForeignCity(tile, unitOwner);
 			var isHumanOwner = this.owner.isHuman;
 			var isActiveTile = this.owner.tileKnowledge.isActiveTile(tile);
 
-			if (isHumanOwner && !isActiveTile && hasOtherOwnersCity) {
+			if (isHumanOwner && !isActiveTile && hasForeignCity) {
 				return Intent.Disabled;
 			}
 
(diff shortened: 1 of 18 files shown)


C7Engine/C7GameData/MapUnit.cs
```diff
@@ -775,96 +775,144 @@ public void OnEnterTile(Tile tile) {
...
+			intent = this.ResolveIntent(tile);
+			return intent == Intent.MoveFreely || intent == Intent.Load || intent == Intent.Unload;
+		}
+
+		public bool CanEnterForcefully(Tile tile) {
+			return CanEnterForcefully(tile, out _);
+		}
+		public bool CanEnterForcefully(Tile tile, out Intent intent) {
+			intent = this.ResolveIntent(tile);
+			return intent == Intent.MoveFreely || intent == Intent.Fight || intent == Intent.Load || intent == Intent.Unload || intent == Intent.WarDeclaration;
```
ajhalme: Maybe a bit too much repetition here: easy to miss updating the other cases, if we add a new intent.

Thinking:
```
CanEnter = CanEnterPeacefully + fight
CanEnterForcefully = CanEnter + declare war
```

C7Engine/C7GameData/MapUnit.cs
```diff
@@ -948,10 +996,27 @@ private static bool HasHostileUnits(Tile tile, Player player) {
...
+		}
+
+		private static Player OtherUnitsOwner(Tile tile) {
+			return tile.unitsOnTile.FirstOrDefault()?.owner;
+		}
 
 		private static bool HasHostileCity(Tile tile, Player player) {
 			return tile.HasCity && AtWar(player, tile.cityAtTile.owner);
 		}
+		private static bool HasOtherOwnersCity(Tile tile, Player player) {
```
ajhalme: Slightly awkward name here - maybe `HasForeignCity(..)`?

ajhalme:
Good stuff, the Intent scheme looks like an useful abstraction.

The `bool` return value of `CanEnterTile(Tile tile, TileProbe probe)` was making things quite complicated at different call sites, so having a more flexible internal representation and simpler signature for `CanEnterTile` should prove useful.

I'll do a bit of playtesting, but looks very reasonable.

stavrosfa:
Yeah, the boolean return value was akward, as then we had to also check the intent after, checking the same things again and again.

You can use the .biq I shared earlier this week, it even has units that are not seen by the (human)player. If you come up with a case I missed, let me know.

ajhalme:
LGTM. I'm sure there are corner cases to revisit in the future, but this all seems fine.

On your test `.biq`, I happened to notice two things:

 1. cavs can hit multiple times per turn -- same on Development, so no bug introduced here 
 
 2. if you route a cavalry unit over neutral railroad and through fog of war where the two barbarians wait in ambush, then this version, the current Development, and Civ3, all do slightly different things. On current Development, the cav attacks the barb and then stops; on this branch, the cav attacks barb and then routes around the second barb; and in Civ3, the cav doesn't attack at all, and just dynamically reroutes around the ambush.

Something to look at another time.

ajhalme (approved):

stavrosfa:
Thanks for having a look!

Yeah, units can attack as longs as they have movement points left, I think it was always there.

As for units rerouting or attacking, besides the technical part, it's also a matter of taste. We can try to emulate 100% what's happening in the original, or we could diverge. I can see good and bad in both, so I don't really prefer one over the other.

With these 2 lines in the `move` method in `MapUnit`, we can avoid the ambush.

```
var canEnter = this.path?.destination == newLoc ? CanEnter(newLoc) : this.CanEnterPeacefully(newLoc);
var canMove = (newLoc != Tile.NONE) && canEnter && (movementPoints.canMove);
```

I don't know if I somehow messed up the pathing algorithm, the rerouting doesn't seem to take the shortest path when avoiding the ambush you mention. Also, I am not sure how the rerouting is triggered. I tried to find it but couldn't.

I don't know if we are happy for know as is and we can merge, and check this out again in the future further, or we should keep on looking at this.
To be honest I would merge since it adds a bit of nice functionality and doesn't necessarily take away anything.

Let me know!

ajhalme:
Fine as is, let's get this in

PR #952, opened 2026-07-04
Connect city to (rail)road network

stavrosfa:
1. Connect city tiles to (rail)road network if an adjacenet neighbour has a (rail)road.
2. Enable railroad worker action with hotkey when a road is alreay present. 
3. Break up initialization of game data, to convert ruleset data first and then some gameplay data, because for example a gampley city/tile needs to know the what terraforms are available. 
4. Small refactors/fixes

These fixes enable us to add a (rail)road terrain improvement on the actual tile, instead of faking it with `HasCity` etc..

The only thing I don't really like is how I initialize the `EngineStorage.gamedata`, I may come up with something better.

Again, I will share a .biq file on Discord (https://discord.com/channels/909680526498082837/909857733388406844/1523044397170692288) to test this out.

### DIFF
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -1289,10 +1289,7 @@ namespace C7GameData {
 		}
 
 		public bool CanPerformTerraformAction(Terraform terraform) {
-			var containsTerraform = unitType.terraformActions.Contains(terraform);
-			var meetsRequirements = terraform.MeetsRequirements(owner, location);
-			var hasCity = location.HasCity();
-			return containsTerraform && meetsRequirements && !hasCity;
+			return CanPerformTerraformAction(terraform, location);
 		}
 
 		public bool CanPerformTerraformAction(Terraform terraform, Tile tile) {
--- a/C7Engine/C7GameData/MapUnit.cs
+++ b/C7Engine/C7GameData/MapUnit.cs
@@ -1,4 +1,5 @@
 using static C7GameData.PlayerRelationship;
+using static C7GameData.Tile;
 
 namespace C7GameData {
 	using Serilog;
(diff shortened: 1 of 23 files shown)


C7Engine/C7GameData/MapUnit.cs
```diff
@@ -1283,17 +1288,23 @@ public bool canBuildCity() {
 			return city;
 		}
 
-		public bool canPerformTerraformAction(Terraform terraform) {
-			return unitType.terraformActions.Contains(terraform) && terraform.MeetsRequirements(owner, location);
+		public bool CanPerformTerraformAction(Terraform terraform) {
```
ajhalme, line 1291: This can probably reuse the more general one:
`CanPerformTerraformAction(terraform, location)`

C7Engine/C7GameData/Tile.cs
```diff
@@ -86,17 +88,59 @@ private Yield ApplyTerrainImprovementModifiers(Tile tile) {
...
+			var hasRailroad = this.HasRailroad();
+
+			// remove stuff like a forest, jungle etc (whatever can be cleared by a worker action), but not hills/mountains etc
+			if (this.overlayTerrainType.allowedFoliageAction != TerrainType.Civ3FoliageAction.None)
+				ClearTerrainOverlay();
+
+			overlays.Clear();
+
+			// Auto connect cities to adjacent road/railroad network
+			if ((!hasRoad && !hasRailroad && this.neighbors.Any(p => p.Value.HasRoad()) || hasRoad)) {
```
ajhalme: These predicates are a bit hard to parse: `!hasRoad` and `hasRoad` in the same statement...

If I'm reading this right, might be cleaner to just return early if tile already has road/railroad. Or maybe we need some idempotent functions like `EnsureHasRailroad(..)`, which only adds a railroad if it's missing, so you can always call it.

ajhalme:
Looks reasonable

ajhalme (approved):

stavrosfa:
I hope this is a bit clearer now, I removed the redundant `hasRoad` check.

ajhalme (approved):
LGTM

PR #953, opened 2026-07-04
Harmonise advisors

ajhalme:
- Add barebones trade, cultural, and foreign advisors
- Harmonise new and existing advisors, so they all function broadly the same from a UI perspective
- Fix popup logic, so one can now quickly switch between advisors
- Replace function key references with dedicated actions, and wire these to Godot InputMap
- Bring in a bunch of custom assets for advisor backgrounds, etc.

There's probably a way to reduce repetition in advisors even further, but one step at a time.

New assets here: https://github.com/C7-Game/Assets/pull/8

### DIFF
--- a/C7/UIElements/Advisors/CulturalAdvisor.cs
+++ b/C7/UIElements/Advisors/CulturalAdvisor.cs
@@ -5,7 +5,7 @@ using Godot;
 
 [GlobalClass]
 [Tool]
-public partial class TradeAdvisor : Control  {
+public partial class CulturalAdvisor : Control  {
 
 	[Export] public TextureRect background;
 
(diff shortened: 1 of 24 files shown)


C7/UIElements/Advisors/CulturalAdvisor.cs
```diff
@@ -0,0 +1,46 @@
+using System.Numerics;
+using C7Engine;
+using C7GameData;
+using Godot;
+
+[GlobalClass]
+[Tool]
+public partial class CulturalAdvisor : Control {
```
stavrosfa, line 8: This comments is for every XAdvisor class. Wouldn't it be better if we had a parent Advisor class that would contain the functionality these classes share? 
Unless this was a problem with the Godot UI and instantiation or something.

Or perhaps your comment
> There's probably a way to reduce repetition in advisors even further, but one step at a time.

was referring to this

C7/UIElements/Advisors/ScienceAdvisor.cs
```diff
@@ -6,112 +6,61 @@
 using static C7Engine.MsgChooseResearch;
 using static C7GameData.EraUtils;
 
-public partial class ScienceAdvisor : TextureRect {
+[GlobalClass]
```
stavrosfa, line 9: I am not sure what changed, but now when opening the Sience Advisor, it seems to be lagging when switching between Eras, and I get this error multiple times

```
ERROR: Condition "p_child->data.parent != this" is true.
   at: remove_child (scene/main/node.cpp:1687)
```

stavrosfa (commented):
The functionality seems to be working as intended, thanks!

C7/UIElements/Advisors/CulturalAdvisor.cs
ajhalme: Yeah, I tried with an abstract class and interfaces and things and Godot Editor got a bit weird about them. Maybe later versions of Godot have better support.

Plus, I suspect a bit of flexibility in how the advisers are set up might prove useful. Common toolkit of util UI methods to call strikes a nice balance.

C7/UIElements/Advisors/ScienceAdvisor.cs
ajhalme: Fixed. I re-parented the UI components to the background, as the main control changed from TextureRect to general Control, but missed the `RemoveChild(..)` in `ChangeEraAndDrawTree(..)`.

ajhalme:
Pulled in the new assets. Also added the tabs to Foreign Advisor.

stavrosfa:
Great, the science advisor seems to be ok now! The tabs in foreign advisor also seem to be changing properly.

Feel free to merge

PR #954, opened 2026-07-06
Unit bombardment improvements

stavrosfa:
1. Improved bombardment behaviour
2. Added lethal bombardment behaviour
3. Split MapUnit class to multiple files
4. Minor fixes/refactors

I took a stab at improving the bombardment mechanic. Now it should behave more like what we expect.

Also, the MapUnit class was getting too hard to navigate, I took the initiative and split it in 3 parts (could be more), one that contains the `is`, `has`, `can` stuff (mostly), one for moving, fighting, etc, and one for the bombard mechanic itself, as it was close to 300 lines of autonomous code.

Let me know what you think about this approach, we could certainly do something similar to `Tile` and other large classes.

### DIFF
--- a/C7/Game.cs
+++ b/C7/Game.cs
@@ -52,25 +52,36 @@ public class TileInfo {
 public class BombardInfo {
 	public MapUnit bombardingUnit;
 	public Tile mouseTile;
+	public MapUnit.BombardTarget bombardTarget = MapUnit.BombardTarget.None;
 
 	public BombardInfo(MapUnit bombardingUnit) {
 		this.bombardingUnit = bombardingUnit;
 	}
 
-	public bool requiresWarDeclaration(Tile tile, out Player player) {
+	public bool RequiresWarDeclaration(Tile tile, out Player player) {
 		player = null;
 
 		var bombarder = bombardingUnit.owner;
-		var foreignUnits = tile.unitsOnTile.Where(x => x.owner != bombarder).ToList();
-		if (!foreignUnits.Any())
+
+		if (bombardTarget == MapUnit.BombardTarget.None)
 			return false;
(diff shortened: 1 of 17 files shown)


ajhalme:
Looks reasonable. Making more use of `partial` is certainly an interesting idea. I definitely support bold, principled refactoring. 

I was thinking the way to break up large classes would be to push functionality out into specialized, probably static classes. Easier for testing as well.

So instead of a `Fight(..)` method in `MapUnit`, you'd have MapUnit instances call something like `FightBehaviour.Fight(mapUnit, ...)` and then you could hide all the fighting-specific functionality in `FightBehaviour.cs` (or just `Fight.cs`, maybe `MapUnitBehaviours/Fight.cs`).

If a more elaborate setup is needed, we could even compose from stateful classes like with a `FightController` class instantiated inside each `MapUnit`.

But happy for us to do more with `partial`, if you are keen. I haven't used `partial` much in other .NET projects, so can't say if it leads to a tidy code base in the long term or not.

stavrosfa:
Partial seemed like a reasonable choice, in the sense that it's non-destructive. It's the same class, split across multiple files; come compile time it's the same as before.

We definitely need to do something, because there are many things to come, and having 5K line classes will be a nightmare, although I am unsure on the approach.

Static classes perhaps is one way, although I am a bit apprehensive of them. I always keep in the back of my mind how it might affect some future multiplayer implementation too. Or how it could affect the moddability of the game.. I don't really know, and unless we get our hands dirty with this, we won't really find out.

If you ever play around with something regarding this, feel free to share as a draft PR, so we can see where it takes us.

ajhalme (approved):

PR #955, opened 2026-07-08
Unit visibility - Radar

stavrosfa:
1. Added the radar ability to units, made configurable in rules.
2. Improved overall unit visibility calculations.
3. Removed forgotten (and false) TODO in MapUnit_Bombard.cs

That's pretty much it, let me know if you have any questions or comments!

### DIFF
--- a/C7/Lua/civ3/ruleset.json
+++ b/C7/Lua/civ3/ruleset.json
@@ -2720,7 +2720,8 @@
       ],
       "unproducible": false,
       "flags": [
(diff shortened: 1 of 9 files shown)


ajhalme (approved):
LGTM

(Can't believe that `return angle switch` bit is valid syntax)

stavrosfa:
Rider did this, I just ALT+Entered on the first if to see if there is something that can be done, because that if/else if block was ugly 😄

PR #956, opened 2026-07-09
Seed persistance

stavrosfa:
Read and use seeds from .biq and .sav files.
Persist seed numbers across save files.
Added some unit tests.

I wanted to push this by itself just so we have a clean point of reference if it goes wrong.

I chose not to add any variation to the .biq and .sav file seeds, like adding a magic number at the end (just so it's not the same) because then I would have to check for overflows/underflows etc.
At the end of the day these numbers from the original game don't mean much to us anyway.


ajhalme (approved):
LGTM

PR #957, opened 2026-07-12
Ruins-Craters-Pollution Overlays

stavrosfa:
1. Added ruins, craters and pollution textures.
2. Added minor functionality for improvements.
3. Split Tile into multiple files.
4. Added some extra flags in the Tile classes for .biq and .sav files. Since I was doing the tedious work of figuring out/confirming which is which, I just added them in.
5. Renamed methods.

The bulk of the changes are due to renaming methods and splitting the Tile class into multiple files.

I kinda had a hard time with this one, mainly because I wasn't sure where to add these overlays. I ended up adding them as "improvements" along side roads, mines etc, for multiple reasons.
One is that they affect food (craters -> -1) and pollution pretty much affects all aspects of production/growth. We also need to enable future modes, where for example, ruins cost +1 movement, but give +1 production, etc..
Another reason was the rendering.
Ruins for example, go over roads, but under mines (which live one the same layer). We would either have to split each improvement into their own layer, which I am not really against or for, it jsut is what it is.

Regarding functionality, I didn't add the workers' abilities to clear rubble/pollution etc.
I take these new improvements into account when calculating for food, or when destroying a city, I spawn ruins. This will need a second pass for sure, as this was already getting heavy in terms of changes.

Another things regarding ruin, pollution and crater textures, besides the ones that connect (not ruins), are the single tile ones.
There are variations for all of them, and I am pretty sure it's decided randomly, based on the game seed.
My previous PR #956 , addresses the seed persistance for this reason.

Right now, if someone checks out this PR, the single tile textures will spawn randomly every time, but will not be consistent across games. For this to work as intended, the previous PR is needed, as mentioned.

I have also created a new PR for the standalone assets [here](https://github.com/C7-Game/Assets/pull/9)

Let me know if you have any feedback, thanks!


C7/Map/TileOverlayLayer.cs
```diff
@@ -51,14 +81,14 @@ public override void drawObject(LooseView looseView, GameData gameData, Tile til
 			}
 		}
 
-		private void drawIrrigaton(LooseView looseView, Tile tile, Rect2 screenTarget) {
+		private void DrawIrrigaton(LooseView looseView, Tile tile, Rect2 screenTarget) {
```
ajhalme: And didn't fix the typo!

ajhalme (approved):
Looks very reasonable.

I agree that it's a bit tricky to integrate the negative terrain "improvements". The argument that we need them all in one place so the effects can be applied correctly makes sense to me.

--

From an OOP design PoV, it feels like we might need a stronger abstraction than `TerrainImprovement`. If we had a `TerrainFeature` or similar, we could inherit both `TerrainImprovement` and `TerrainEffect` from it, and then make use of either the general case (`TerrainFeature`) when want both kinds, or the specialized functions when interested in one of the subclasses. Many of the existing improvement-oriented functions could probably stay as they are, focused on the "positive" improvements.

This would then allow a refactoring like:
- `TerrainImprovement` -> `TerrainFeature` : `TerrainImprovement` + `TerrainEffect`
- `GetManMadeImprovements` -> `GetTerrainImprovements`
- `GetEffectImprovements` -> `GetTerrainEffects`

This might then extend to a third subclass for airfields, outposts, etc.

**That said,** it would be quite a hassle to implement, as "terrain improvement" is widely used as the primary tile feature abstraction. I'm happy with this slight awkwardness of having the effects be managed as improvements like the other things.

C7/Map/TileOverlayLayer.cs
stavrosfa: Done!

stavrosfa:
Since you don't have any major issues with it now, and the foundation is set, I believe we can merge as is.

I 'll be happy to take another stab at it at some point, maybe after integrating airfields and outposts and stuff, so I can have a better understanding of the whole thing.

In all this madness, barbarian camps also need to be included somewhere, and stop having a single boolean driving the whole thing.
In the original, all these are simply "overlays". I don't know how the handle them iternally in the code, but perhaps we are overthinking this..

Btw, I did try to create seperate classes and have a base class like TerrainModification (or something), I can't remember exactly what didn't quite work, but I scrapped it and went with this approach.

I also tried to have simple booleans like hasRuins, hasPollution, etc, but that wan clearly not enough.

I even tried to rename the TerrainImprovement class, to better represent what it does, but that bit me as well.

PR #958, opened 2026-07-12
Add GameViews: Demographics, Victory Status, etc.

ajhalme:
Add base presentation functionality for what I'm calling "Game Views", the screens presented when pressing F7-F11 in the original game:
- "Wonders of the world"
- Victory Status / Histogram
- Palace (existing minigame; rewired & harmonised with other views)
- Space Race
- Top 5 Cities / Demographics

Implementation is basically identical to Advisors (https://github.com/C7-Game/OpenCiv3/pull/953), with Godot keymapping and dedicated controls. Note that I cross-wired the signals, so switching to a "game view" hides any advisor, and vice versa.

Palace was fully fleshed out, but I've made it behave a bit more like the other screens for consistency. I also removed the all-black framing around it: we can add it back in ([NinePatchRect](https://docs.godotengine.org/en/stable/classes/class_ninepatchrect.html) might be the thing to try), but in my view(!) this "windowed" mode makes it work a bit nicer with the others. If we want to wire up a custom variant of the Palace minigame in OC3, we can get into that as well.


stavrosfa:
FYI, the project needs to be re-built from inside godot to take effect. I was getting all sorts of errors, but I remembered that if nodes and stuff are changed, building from the editor fixes it.

I don't mind Windowed/black screen around the frames, but the background should be blocked from being able to click stuff outsiede the window. Now I can move around the map, select units, etc..

We could have a transparent node blocking the whole BG, like we do (in part) with the LowerRightInfoBox.

In some frames, I can't click through from inside the frame, but in others, this brakes major functionality. For example, the Science advisor is unsuable now, I can't select anything.

This looks to be the case for F1-F6 at least.

image

ajhalme:
Godot gave me lots of trouble for these overall changes. My Godot for some reason wants to embed the textures into the scenes, fails to load file changes from outside the editor, and throws all kinds of weird errors here and there when using custom classes. Maybe I'm holding it wrong, but I'm not impressed - hopefully later versions of Godot fix some of this stuff.

Only thing I didn't include in the original push was some of these `.uid` files, but I've included those now. There's a few of those from earlier PRs as well. Hopefully that resolves any build issues. (Another thing a Godot upgrade might fix, as I believe the _uids_ travel in the scene files nowadays.)

I've fixed the prevent-click-around-the-window functionality by restoring the event gates for advisor/gameView. Also fixed a bug in the advisor/gameViews switching code that was resulting in the main control not getting hidden when switching, which I think led to the non-responsive Science Advisor issue.

stavrosfa (approved):
Looking good now, the functionality seems to have been restored, and F1-11 frames are working as intented. F1 & F6 are working (these are the only ones with UI functionality), as well as the palace minigame.
Thanks!

stavrosfa:
Godot for me is almost unusable when trying to do something from within the editor. I am really not impressed either, and have whinned about it multiple times in the past, I won't repeat this here 😅 

Untill C# becomes a first class citizen, which as far as I understand won't happen any time soon, we will be running in to all sorts of problems.

stavrosfa:
> we can add it back in ([NinePatchRect](https://docs.godotengine.org/en/stable/classes/class_ninepatchrect.html) might be the thing to try),

Ah, I wanted to commnet on this.

I believe that is what is done (manually?! though), for the pop ups we already have in the game (which also need some touch up)

https://github.com/C7-Game/OpenCiv3/blob/b9d46f55a63c05ee3a147267798d5df89f157029/C7/UIElements/Popups/Popup.cs#L104

PR #959, opened 2026-07-13
Abandon City - Civ Destruction

stavrosfa:
Add the option to abandon a city.
Update some checks for civ destruction.

### DIFF
--- a/C7Engine/EntryPoints/CityInteractions.cs
+++ b/C7Engine/EntryPoints/CityInteractions.cs
@@ -46,7 +46,7 @@ namespace C7Engine {
 		}
 
 		public static void DestroyCity(City city) {
-			DestroyCity(city.location.XCoordinate, city.location.YCoordinate);
+			DestroyCity(city.location);
 		}
 		public static void DestroyCity(Tile tile) {
 			DestroyCity(tile.XCoordinate, tile.YCoordinate);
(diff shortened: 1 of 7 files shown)


C7Engine/EntryPoints/CityInteractions.cs
```diff
@@ -45,29 +45,43 @@ public static City BuildCity(Tile tileWithNewCity, Player owner, string name) {
 			return newCity;
 		}
 
+		public static void DestroyCity(City city) {
+			DestroyCity(city.location.XCoordinate, city.location.YCoordinate);
```
ajhalme: I would have reached out for `DestroyCity(city.location)` here, but fine as is of course.

ajhalme (approved):
LGTM.

I don't think Civ3 allows you to destroy your last city. You *can* take out your capital in Civ3, but you can't finish off your civ like this. That said, this self-destroy seems to be handled OK here.

stavrosfa:
> I don't think Civ3 allows you to destroy your last city. You can take out your capital in Civ3, but you can't finish off your civ like this

I didn't want to add any potentially moddable functionality, or any rules as to how we want to handle that, I only wanted a quick way to be able to abandon my cities (mainly so I could work on the ruin spawning part of it)

C7Engine/EntryPoints/CityInteractions.cs
stavrosfa: Makes sense, thanks

ajhalme (approved):

PR #979, opened 2026-07-24
Domestic Science-Luxury Sliders made draggable

stavrosfa:
Made the icons for science and luxury sliders draggable using the godot native HSlider.
Update the plus/minus textures as they were using some other file, and removed the shadow which is causing some texture artefacts. 

I am working on some bigger UI stuff, and I wanted to see how the draggable sliders work, so instead of creating a prototype just for this, I figured I'd use the opportunity to update the domistic advisor sliders.

I think I am going to update the scrollbar next, as this is mainly what I wanted to figure out in the first place.

EDIT:
Found this after pushing this, you can update the sliders by clicking on the (invisible) slider bar. I guess that's ok. 
My only concern is that it extends a bit "outside" what looks like the bar in the UI so it might be confusing as to why it moves when you click a bit outside the bar?


C7/UIElements/Advisors/domestic_advisor.tscn
```diff
@@ -15,12 +15,41 @@ font_color = Color(0, 0, 0, 1)
 
 [sub_resource type="StyleBoxEmpty" id="StyleBoxEmpty_5hkgh"]
 
-[node name="DomesticAdvisor" type="CenterContainer" node_paths=PackedStringArray("background", "close", "changeGovernment", "governmentLabel", "scienceStatus", "treasury", "incomeDetails", "expenseDetails", "incomeSummary", "expenseSummary", "sumSummary", "growth", "cityListContainer", "eatenFood", "fullFood", "wastedShield", "goodShield", "wastedGold", "goodGold", "happyFace", "contentFace", "beaker", "treasuryIcon")]
+[sub_resource type="Image" id="Image_fbojv"]
+data = {
```
ajhalme: I don't think we want to embed Image (sub-)resources into the scene as data. These are small images, so it's less bad than a full blown background image, but even if we wanted to embed images as resources, we'd probably want to do it with dedicated `.tres`.

I asked Claude what we can do here, and section "Fixing scenes that already have it embedded" has instructions on how to fix existing embeddings via `.tres`.
~https://claude.ai/chat/a9387651-a620-40ef-ac43-d7178d5e87ba~
https://claude.ai/share/dfc6bd00-b472-4a57-a619-8d18bd0eac1e

Bad news that there probably isn't a way to do this cleanly, as long as we are loading texture dynamically with our custom loader. There's a mention of `ResourceSaver` which I guess we could look at as a native replacement to the current in-memory texture-cache (?), but I don't know. 

Note that a `.tres` still embeds the image in the project, which we certainly don't want to do with Civ3 assets.

C7/UIElements/Advisors/DomesticAdvisor.cs
```diff
@@ -145,31 +160,72 @@ private void CreateUI() {
...
+
+		if (value > luxuryRate) {
+			MoreLuxury();
+		} else {
+			LessLuxury();
+		}
+	}
+
+	private void MoreScience() {
+		new MsgChangeSliders(true, false, false, false).send();
```
ajhalme: I feel like there's a enum in this `MsgChangeSliders(..)` signature, trying to break free.

ajhalme (commented):
Sliders are sliding great. In-scene texture embedding worth another look.

C7/UIElements/Advisors/domestic_advisor.tscn
esbudylin: I can't open the claude link, but I remember facing exactly the same problem with Godot embedding images from [Tool] scripts into the scene files. I solved it with [this util method](https://github.com/C7-Game/OpenCiv3/blob/9dde9b6818d1e5c16341398b36117846b0f321f8/C7/Util.cs#L376).
 
Basically, it prohibits Godot from saving the state of certain properties into a scene. I've already created classes Civ3TextureRect and Civ3TextureButton that make use of this util. I think the problem with this PR can be solved the same way by creating a custom Civ3HSlider script and assigning it to the slider nodes.

ajhalme: Fixed the link to be shareable.

I've seen the `_ValidateProperty(..)` method override mentioned in some workarounds. Creating these custom UI classes is definitely an option, but of course it's a bit of a hassle to make one for each textured UI widget and building block. From what I've seen, the custom node support is also a bit flaky, at least in this version of Godot.

But yeah, happy with more custom textured components.

stavrosfa: As a rule of thumb I am not against creating custom UI widgets.

> I think the problem with this PR can be solved the same way by creating a custom Civ3HSlider script and assigning it to the slider nodes.

I tried to do that, following the principle from the Civ3TextureRect/Civ3TextureButton, but the issue here is that, I am not assigning the HSlider icons directly, rather, I am overriding the theme icons, so there are no properties that I can access to which I can apply the ApplyNoSaveFlag. 
I tried figuring out if I can create a custom Civ3Theme for example, and apply it there, but no luck. There is also not a direct way to apply these icons as far as I understand.
I had the same luck Civ3Texture2D and Civ3ImageTexture, I can't find any relevant properties.

So I am not sure how to move forward with this. Should I look into the `.tres` thing that was mentioned? I am afraid this would be more of a hassle anytime we want to make a change in the UI, no?

Let me know if you have any input.

stavrosfa: @ajhalme You mentioned something in #958 [here](https://github.com/C7-Game/OpenCiv3/pull/958#issuecomment-4953066086) , how did you manage to not embed any textures in there?

C7/UIElements/Advisors/DomesticAdvisor.cs
stavrosfa: I updated it. I chose to leave the bools inside the class since it's easier to do 
`if (moreScience)`
 rather than 
`if (slider == DomesticSlider.MoreScience)`
 there. 

The API however is improved by doing
 `new MsgChangeSliders(DomesticSlider.MoreScience).send();` 
instead of 
`new MsgChangeSliders(true, false, false, false).send();`

C7/UIElements/Advisors/domestic_advisor.tscn
ajhalme: I didn't resolve this in my PRs, I just removed the embedded textures by hand.

--

This kind of thing seems to work:

```csharp
[GlobalClass]
[Tool]
public partial class Civ3HSlider : HSlider {
	[Export] string textureConfigKey;

	public ImageTexture Texture { get; set; }

	public override void _Ready() {
		if (textureConfigKey != null) {
			Texture = TextureLoader.Load(textureConfigKey);
			
			AddThemeIconOverride("grabber", Texture);
			AddThemeIconOverride("grabber_highlight", Texture);
			AddThemeIconOverride("grabber_disabled", Texture);
			AddThemeStyleboxOverride("slider", new StyleBoxEmpty());
		}
	}

	public override void _ValidateProperty(Godot.Collections.Dictionary property) {
		Util.ApplyNoSaveFlag(property, [PropertyName.Texture]);
	}
}
```
And then (etc.):
```
[ext_resource type="Script" uid="uid://lgy72whmxcwt" path="res://UIElements/Civ3HSlider.cs" id="4_gltop"]
...
testSlider = NodePath("Background/TestSlider")
...
[node name="TestSlider" type="HSlider" parent="Background"]
layout_mode = 0
offset_left = 566.0
offset_top = 350.0
offset_right = 764.0
offset_bottom = 380.0
value = 5.0
script = ExtResource("4_gltop")
textureConfigKey = "icons.science"
metadata/_custom_type_script = "uid://lgy72whmxcwt"
```

C7/UIElements/Advisors/DomesticAdvisor.cs
ajhalme: `Enum -> handler function` was how I was thinking about it. With dedicated case handlers, we can push the do-nothing gates into the handlers as well. (And the handlers are good candidates for methods on a dedicated policy-manager type class, if we want to move the engine side apply out of this class one day.)

So, something like:

```csharp
	public class MsgChangeSliders : MessageToEngine {
		public enum PolicyChoice {
			MoreScience,
			LessScience,
			MoreLuxury,
			LessLuxury,
		}

		private PolicyChoice policy;

		public MsgChangeSliders(PolicyChoice policy) {
			this.policy = policy;
		}

		public override void process() {
			Player player = EngineStorage.gameData.GetFirstHumanPlayer();

			switch (policy) {
				case PolicyChoice.MoreScience:
					MoreScience(player);
					break;
				case PolicyChoice.LessScience:
					LessScience(player);
					break;
				case PolicyChoice.MoreLuxury:
					MoreLuxury(player);
					break;
				case PolicyChoice.LessLuxury:
					LessLuxury(player);
					break;
				default:
					throw new ArgumentOutOfRangeException();
			}

			// Update citizen moods in all cities, as changing the sliders can
			// change moods.
			foreach (City city in player.cities) {
				city.RecalculateCitizenMoods(EngineStorage.gameData);
			}

			// Update the ui to reflect our changes.
			new MsgUpdateUiAfterDomesticChange().send();
		}

		private static void MoreLuxury(Player player)
		{
			if (player.luxuryRate == 10)
				return;

			player.luxuryRate++;
			if (player.taxRate > 0) {
				player.taxRate--;
			} else {
				player.scienceRate--;
			}
		}

		private static void LessLuxury(Player player)
		{
			if (player.luxuryRate == 0)
				return;

			player.luxuryRate--;
			player.taxRate++;
		}

		private void MoreScience(Player player)
		{
			if (player.scienceRate == 10) {
				return;
			}
			
			player.scienceRate++;
			if (player.taxRate > 0) {
				player.taxRate--;
			} else {
				player.luxuryRate--;
			}

		}

		private static void LessScience(Player player)
		{
			if (player.scienceRate == 0)
				return;

			player.scienceRate--;
			player.taxRate++;
		}

}
```

stavrosfa: Yeah, I see what you mean.
I can see the MoreLuxury() methods moving away from this class, but I think the initial intention was to be always using the Engine Messages, and not allow outside interference, so it's more difficult to mess up.

C7/UIElements/Advisors/domestic_advisor.tscn
stavrosfa: Sorry for the long answer

I did (simplified)

```
[GlobalClass]
[Tool]
public partial class Civ3HSlider : HSlider {
    public ImageTexture grabber { get; set; }
    
    public void Initialize() {
        if (grabber != null) { AddThemeIconOverride("grabber", grabber); }
    }
    
    public override void _ValidateProperty(Godot.Collections.Dictionary property) {
        Util.ApplyNoSaveFlag(property, [PropertyName.grabber]
    }
}

```

but after removing the `sub_resource` stuff manually and building the project in godot, I still get the images byte array.

Then I did 

`[Export] public ImageTexture grabber { get; set; }`

and removed the packed arrays again by hand.
Now, I don't get the exported property in the godot editor (which I don't really care for, this is just a side-effect) and I don't get the packed arrays.

It's easier to show the diff than try to explain what's going on

image

What I am unsure of, is that I still get the `sub_resource` and the icon references.

So I removed those by hand as well, rebuilt the project, and now it looks like

image

The icons still exist in the editor, so I guess all in all this is ok?

stavrosfa:
This looks like it works in the way we want (on my machine 😛)

Please have a look and let me know.

Thanks for the guidance and feedback to the both of you!

stavrosfa:
I have another scene, with the exact same setup, and it seems that it doesn't work. 

The moment you save the .tscn for any reason, the textures are regenarated. I will try this in a while with this one too, because I think I only tested it with building and running the game, not altering and saving a scene.

Sometimes, I 'll do the simplest thing, like turning the visibility of a node off, saving the scene, turn it on again, save again, then I will start getting the textures from Civ3TextureRect and Civ3TextureButton items, and it will get corrupted, throwing null pointers at literally stuff like, 

```
TextureRect advisorHead = new(); // class field

advisorHead.Texture = AdvisorHead.GetPopupImage(AdvisorHead.Advisor.Domestic, AdvisorHead.Mood.Happy, eraIndex: 0); // inside a method called in _Ready
```

The struggle is real with this one, I feel like I keep fighting the engine...

C7/UIElements/Advisors/domestic_advisor.tscn
ajhalme: I think the key to this pattern is to never have an `ImageTexture` directly in the scene. The script can have an image texture, but it's initialized via key reference passed into our custom loader. This way the script can fully own the texture load, and can block the native texture data caching at the scene level via `_ValidateProperty(..)`.

Godot like to manage textures itself, which is reasonable for most games, but as we are all about dynamic loading (see: mods), we have to do these things indirectly.

I'll take a closer look at these changes, but the diffs look directionally correct to me.

stavrosfa: Check the latest commit, things have changed quite a bit since this post

ajhalme:
> I have another scene, with the exact same setup, and it seems that it doesn't work. [..] The moment you save the .tscn for any reason, the textures are regenarated. I will try this in a while with this one too, because I think I only tested it with building and running the game, not altering and saving a scene.

Existing or new? If existing, and these are the cases I manually "fixed", then this is kind of expected: texture loading isn't custom everywhere yet.

stavrosfa:
> Existing or new? If existing, and these are the cases I manually "fixed", then this is kind of expected: texture loading isn't custom everywhere yet.

It was actually the same scene, just on a different branch. It wasn't the textures you are referring to, because I took the packed array rgba data and displayed it, so i know that each and every one was one of mine.

No matter, I have come up with a different approach, so it's mostly irrelevant now.

ajhalme (approved):
LGTM.

The Godot editor event detection hack in `Civ3HSlider` is not to my liking, but at least it's neatly tucked away in that class. I suppose every engine has it's little quirks.

C7/UIElements/Civ3HSlider.cs
```diff
@@ -0,0 +1,66 @@
...
+[Tool]
+public partial class Civ3HSlider : HSlider {
+	public ImageTexture grabber { get; private set; }
+	public ImageTexture grabberHighlight { get; private set; }
+	public StyleBox sliderStyleBox { get; private set; }
+
+	public Civ3HSlider AddGrabber(ImageTexture texture) =>
+		SetIcon("grabber", texture, t => grabber = t);
+
+	public Civ3HSlider AddGrabberHighlight(ImageTexture texture) =>
```
ajhalme: _[Style note]_

Unless I'm missing some object life cycle thing, the lambda `Action<ImageTexture>` constructs for these slider actions feel a little unnecessary to me - I doubt we'll be calling `SetIcon(..)` with a range of different Actions.

Also not sure what we are gaining by swallowing null texture references. Is it nicer in the editor or something?

So instead of:
```csharp
	SetIcon("grabber_highlight", texture, t => grabberHighlight = t);

	private Civ3HSlider SetIcon(string name, ImageTexture texture, Action<ImageTexture> assign) {
		assign(texture);
		if (texture != null)
			AddThemeIconOverride(name, texture);
		return this;
	}
```

I'd consider simply:
```csharp
    grabberHighlight = texture;
    SetIcon("grabber_highlight", texture);

	private Civ3HSlider SetIcon(string name, ImageTexture texture) {
	    AddThemeIconOverride(name, texture);
		return this;
	}
```

stavrosfa:
_Leaving this comment here as it will be removed from the code, but I think it's still helpful._

Unlike Civ3TextureButton or Civ3TextureRect where properties like Texture, TextureNormal, etc, are exposed in their parent classes, when we override a theme, godot seems to do some behind the scenes (pun not intended) magic .

So it's not enough to expose our custom texture fields, and then use _ValidateProperty and Util.ApplyNoSaveFlag to prevent godot from serializing the image textures.

We do this "hacky" thing where before godot saves the scene in the editor, we remove all the theme overrides, godot saves the scene, and re-applies them.It needs a bigger setup than I would prefer, but currently seems to be the most reliable way of preventing godot from serializing the textures. 

This is the post I found that helped with this (had to use some chatgpt help to translate it to c#): https://forum.godotengine.org/t/prevent-generated-resources-from-being-serialized-in-the-tscn/46562/3

stavrosfa:
Sorry for changing things again a little bit, but i think this is for the better, given the situation.

You 'll see now that the Civ3HSlider (and while I was at it, I added the Civ3VSlider equivalent) has an interface, which might not make much sense now, but while I am working on the side on the custom scrollbar, I realised that I needed a way to get a generic ScrollBar node so I can apply the theme (I can't expose the interface to Godot).

The Civ3RangeTheme is a way to have a centralized implementation for the various Civ3V-Range-Like and Civ3H-Range-Like stuff.

I cycled through many different variations of this, and I had some back and forth with ChatGPT, which has helped a lot with the Godot UI stuff. This is far from vibe-coded, I haven't used any AI for this project in a long time, just making sure I am transparent I guess..

Let me know if you have any quiestions/suggestions.

C7/UIElements/Civ3HSlider.cs
stavrosfa: Lambdas ammended in the newer version.

As for the null checks, I removed them at some point, and godot was throwing some kind of error. I don't think we gain much either way, in my mind it's a simple object null check, if we can't see a texture, we propably aren't loading it right.

stavrosfa:
> The Godot editor event detection hack in Civ3HSlider is not to my liking, but at least it's neatly tucked away in that class. I suppose every engine has it's little quirks.

I don't like it much either, but as I said it seems to be the only way to not have theme overrides serialize the textures.

C7/UIElements/Civ3RangeTheme.cs
```diff
@@ -0,0 +1,190 @@
...
+		if (styleBox != null)
+			this.range.AddThemeStyleboxOverride(name, styleBox);
+	}
+
+	public void ClearAndRestoreOverrides(int what) {
+		if (what == Node.NotificationEditorPreSave)
+			this.RemoveOverrides();
+
+		if (what == Node.NotificationEditorPostSave)
+			this.Apply();
```
ajhalme: I'd rename `Apply()` to `ApplyOverrides()` -- or even `ReapplyOverrides()` -- for symmetry and clarity.

Both remove and apply operations should probably be private as well.

`ClearAndRestoreOverrides` could do with a comment line that documents this theme component juggling mechanism, and maybe points at the fluent-style methods one is supposed to call when parameterising this custom range theme with textures and stylesboxes.

We could also try and go further in hiding the Godot hack, by renaming `ClearAndRestoreOverrides` to something more generic like `HandleEditorNotification`. Not sure if worth it, though.

ajhalme (approved):
Working fine, changes look reasonable. Left a renaming suggestion.

C7/UIElements/Civ3RangeTheme.cs
stavrosfa: Thanks, I updated the code.

>and maybe points at the fluent-style methods one is supposed to call when parameterising this custom range theme with textures and stylesboxes.

Instead of trying to explain it with words, I think the best way for someone looking at it for the first time and trying to understand it, is to look at how it's already being used. Is something not clear about how it's used?

>We could also try and go further in hiding the Godot hack, by renaming ClearAndRestoreOverrides to something more generic like HandleEditorNotification. Not sure if worth it, though.

I think we need to be clear as possible about what this does, because it can be confusing but necessary, so I don't want to try and hide it.

ajhalme:
🚢

stavrosfa:
Ahoy!

PR #983, opened 2026-08-07
Victory status, towards victory conditions

ajhalme:
Adds victory conditions and various scaffolding needed for their evaluation.

- Extend ruleset with victory conditions
- Add Civ3Checkbox UI element (started from a clone of the Civ3MenuButton)
- Extend Victory Status view with a display grid on which we can draw victory status rows
- Render "registered" victory conditions in the VS view, depending on game mode
- Add New Game Player Set widget for selecting games rules, add a couple of victory condition selectors as a demo
- Add age multiplier to culture-per-turn calc (sketch) 
- Model victory conditions in OOP style as implementations of an IVictory interface
- Extend SaveGame with victory conditions and score history, load from save
- Wire score history maintenance into main game loop
- Add a victory status calculator and sketch out some victory calculations
- Add some util fuctions for victory calcs in Tile, TileKnowledge, City, GameData, Player, TerrainType, Building

Victory conditions can be evaluated (and are, for VictoryStatusView), but the game cannot be won yet. Full game loop wiring and the mechanics of winning the game need more thought.

Per turn score calculation looks pretty good, per turn culture calc is not quite there, per turn power is unimplemented. (Per turn power formula is not known to the modding community, AFAICT). 

Checkboxes from https://github.com/C7-Game/Assets/pull/14


ajhalme:
This victory conditions PR is now open for reviews. See first post for changes.

Plan is to implement victory conditions and related scaffolding here, and then take on actual game winning wiring in a separate effort.

C7Engine/C7GameData/Player.cs
```diff
@@ -1130,6 +1130,37 @@ public int ShieldCost(IProducible producible) {
...
+			if (!gameData.history.ContainsKey(id.ToString()))
+				return;
+
+			int n = gameData.history[id.ToString()].Count;
+			HistTurnRecord lastTurn = gameData.history[id.ToString()].LastOrDefault();
+
+			// "Power is an amalgam of cities, gold, culture, advances, resources, military strength,
+			// nuclear weapons, and wonder."
+			// Exact formula is unknown, so we repeat the previous value.
+			// TODO: Figure out power formula
```
stavrosfa, line 1146: I would add a TODO here to make this formula/score calculations moddable (where it makes sense)

C7Engine/C7GameData/ImportCiv3.cs
```diff
@@ -894,6 +959,13 @@ private void ImportSavLeaders() {
 			}
 		}
 
+		private static bool IsDefeated(SavePlayer player, QueryCiv3.Sav.LEAD leader) {
```
stavrosfa, line 963: Is this really that simple in the base game, or should we add a few TODOs here?

C7/UIElements/NewGame/PlayerSetup.cs
```diff
@@ -99,6 +105,25 @@ public override void _Ready() {
...
+		victoryConditions = VictoryConditions.WarMongerDefault();
+		rulesContainer.Columns = 2;
+		rulesContainer.AddThemeConstantOverride("v_separation", 0);
+
+		var dominationVictory = new Civ3Checkbox() { Text = "Allow Domination Victory" };
+		dominationVictory.SetPressed(victoryConditions.AllowDominationVictory);
+		dominationVictory.Toggled += (state) => { victoryConditions.AllowDominationVictory = state; };
+		rulesContainer.AddChild(dominationVictory);
+
+		var placeholder1 = new Civ3Checkbox() { Text = "Placeholder" };
```
stavrosfa: Not sure what's going on here. Why the placeholder?

C7/UIElements/GameViews/VictoryStatusView.cs
```diff
@@ -28,12 +38,113 @@ private void CreateUI() {
...
+
+	private void ClearGrid() {
+		foreach (Node child in gridHeader.GetChildren())
+			child.QueueFree();
+
+		foreach (Node child in grid.GetChildren())
+			child.QueueFree();
+	}
+
+	private const string HeaderPadding = "     ";
```
stavrosfa, line 93: Not exactly sure where this is being used, but isn't there a better way to add padding?
I also think that these views should be done directly in Godot, as it makes it friendlier in general (not so much for the one to set this up at first :P)
No need for this to be done now, a TODO is fine.

stavrosfa (commented):
As far as I can tell, without being able to win/lose, it looks ok. I have left some comments.

I would also rather in the future that we split such big PRs into smaller, more managable chunks, so that they can be evaluated better and be easier to revisit in the future.

C7/UIElements/NewGame/PlayerSetup.cs
ajhalme: Godot wants to fill a GirdContainer by row. Couldn't immediately figure out how to make it fill by column. The placeholder makes it so the first and third elements go in the first column.

I've removed the placeholder: having just the two win conditions on the same row is fine for now.

C7Engine/C7GameData/ImportCiv3.cs
ajhalme: Added a TODO.

Probably good enough for now. No cities means civ is almost certainly dead, except for the very beginning of the game - hence the unit counter. I think civs that are killed in the original do have their remaining units wiped, so this should be reasonably correct for Civ3 save files.

C7Engine/C7GameData/Player.cs
ajhalme: Added TODO

ajhalme:
Yeah, not happy with how much stuff there is in this one. I think I'll slice off some of the util stuff and UI bits in separate PRs.

C7/UIElements/NewGame/PlayerSetup.cs
stavrosfa: Did a quick search, it seems that indeed Godot doesn't offer that out of the box, it's only been 2+ years since there was a proposal/PR for this, so fingers crossed, we might be getting this in this decade 😆 

As for our thing, it's fine, I just wasn't sure what the purpose was.

C7Engine/C7GameData/ImportCiv3.cs
WildWeazel: The original condition is no cities and no settlers (or more accurately, units with the build city action). All units are wiped on defeat, as well as any remaining cities if defeated under elimination rules, which obviously is a separate case.

ajhalme:
I've patched up this PR: the game is now winnable.

I've simplified things a bit, so that "Turn Limit" is the only supported victory type. The player with the highest score at the end of the last turn wins. End of the game is marked only with a simple popup on the UI side, for now.

PR #984, opened 2026-08-27
Towards moddable audio

ajhalme:
- Replicate TextureLoader pattern for audio
- Lua config for reference based audio file targeting
- `audio.lua` files for initial Lua-based modding/override
- General audio refactor

-> https://github.com/C7-Game/Assets/pull/15

### DIFF
--- a/C7/Audio/AudioLoader.cs
+++ b/C7/Audio/AudioLoader.cs
@@ -14,10 +14,6 @@ public static class AudioLoader {
 	private static Dictionary<string, AudioStream> configKeyCache = [];
 
 	static AudioLoader() {
-		// We need to register the "Type" type to be able to inspect
-		// the types of C# objects in the Lua code
-		UserData.RegisterType<Type>();
-
 		// Initialize when running in the editor
 		// In game it is done by GlobalSingleton, but it's not accessible in the editor
 		if (Engine.IsEditorHint()) {
--- a/C7/Audio/AudioLoader.cs
+++ b/C7/Audio/AudioLoader.cs
@@ -1,0 +1,132 @@
+using System;
+using Godot;
+using System.Collections.Generic;
+using System.IO;
+using MoonSharp.Interpreter;
+using Script = MoonSharp.Interpreter.Script;
+using C7Engine.Lua;
+
+public static class AudioLoader {
+
+	private static Script lua;
+	private static Table audioConfig;
+
+	private static Dictionary<string, AudioStream> configKeyCache = [];
+	private static Dictionary<(string configKey, object obj), AudioStream> objectMappingCache = [];
+
+	static AudioLoader() {
+		// We need to register the "Type" type to be able to inspect
+		// the types of C# objects in the Lua code
+		UserData.RegisterType<Type>();
+
+		// Initialize the TextureLoader when running in the editor
+		// In game it is done by GlobalSingleton, but it's not accessible in the editor
+		if (Engine.IsEditorHint()) {
+			GameMode gameMode = GameMode.Load(GamePaths.GameModesDir, GamePaths.basic);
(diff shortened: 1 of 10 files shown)


C7/Audio/AudioLoader.cs
```diff
@@ -0,0 +1,113 @@
...
+
+	private static Script lua;
+	private static Table audioConfig;
+
+	private static Dictionary<string, AudioStream> configKeyCache = [];
+
+	static AudioLoader() {
+		// We need to register the "Type" type to be able to inspect
+		// the types of C# objects in the Lua code
+		UserData.RegisterType<Type>();
```
WildWeazel: Is this unique to audio, or does it matter if it's done multiple times? Just wondering if it should be raised up to a higher level constructor.

C7/Audio/AudioLoader.cs
```diff
@@ -0,0 +1,113 @@
...
+			return simplePath;
+		}
+
+		throw new ArgumentException($"Invalid audio config format: {entry?.GetType().Name ?? "null"}");
+	}
+
+	private static AudioStream LoadFromPath(string path) {
+		string ext = Path.GetExtension(path).ToLowerInvariant();
+
+		return ext switch {
```
WildWeazel, line 83: This is fine for now but maybe it would be better to query the file format to account for variants like .oga or just misnamed files in mods

C7/Audio/AudioLoader.cs
```diff
@@ -0,0 +1,113 @@
...
+		string[] parts = configKey.Split('.');
+		object current = audioConfig;
+
+		foreach (string part in parts) {
+			if (current is Table table && table[part] != null) {
+				current = table[part];
+			} else {
+				return null;
+			}
+		}
```
WildWeazel, line 102: I'm confused, this will end up returning the last such table entry, correct? We would iterate backwards and return the first match in that case. But why would multiple substrings be found in the table anyway?

C7/Audio/AudioLoader.cs
ajhalme: Probably harmless to duplicate the registrations, but I've hoisted the UI-side Lua type registrations from AudioLoader and TextureLoader into GlobalSingleton.

C7/Audio/AudioLoader.cs
ajhalme: Added `.oga` as an explicit variant for `.ogg`. File content based (MIME?) type resolution is at least as fiddly.

C7/Audio/AudioLoader.cs
ajhalme: This is identical to TextureLoader. Inlined it here because it's a compact static helper.

I'm not sure I follow your interpretation. This function resolves the _value_ from a nested Lua structure:
```
audio.menu = {
  main_menu_1 = "/my/path/file.mp3"
}
```
```
Parts:       "menu.main_menu_1" --> ["menu", "main_menu_1"]
init:        current := audio
1st pass:    current["menu"] --> current := {  main_menu_1 = "/my/path/file.mp3" }
2nd pass:    current["main_menu_1"] --> current := "/my/path/file.mp3"
return:      "/my/path/file.mp3"
```

The modding Lua code makes a second lookup to see if there's a direct key-based override for the structured key, using it as a flat key-value lookup. That is, there is no structure to parse in the modded code. We could make the moddable stuff structured as well, but there's no need for now.

The reason to have structured Lua data in the first place is to enable not just simple values but full objects with properties to be retrieved by a single key.

WildWeazel: Got it, I missed that this is recursing into the table structure

WildWeazel (approved):

Issue #963, opened 2026-07-15
Keyboard interacts with game when writing save file name


FranklinFFR:
I have observed the game move units and end turn while I am typing into the file name box when saving. 

This should hopefully be as easy as creating a flag that gets turned on when in a menu and making the key presses ignored when that flag is set to true.

IcanBENCHurCAT:
## Speckit Analysis Complete

A comprehensive implementation plan has been created using the speckit workflow. This is a quick fix for the keyboard input bug.

### Spec Summary (`.specify/specs/963-save-dialog-input-blocking.md`)
### Plan Summary (`.specify/plans/963-save-dialog-input-blocking.md`)

### Root Cause
The save dialog uses Godot's native `FileDialog` (via `Civ3FileDialog.cs`), which inherits from `Window` - not `Popup`. The `Game._UnhandledInput()` guard checks `HasVisibleModal()` which doesn't detect FileDialog, so keyboard input meant for the save filename is forwarded to the game layer, triggering unit moves, view toggles, etc.

### Proposed Fix (tiny - 3 lines across 2 files)
1. **`Civ3FileDialog.cs`**: Add `public bool DialogVisible => Visible;` property
2. **`Game.cs`**: Add `private Civ3FileDialog loadDialog` field (wired via `GetNode`)
3. **`Game.cs` (_UnhandledInput)**: Extend guard with `|| loadDialog?.DialogVisible == true`

### Key Findings
- Exact line numbers and diff format provided for each change
- 10 manual verification tests planned (save, load, cancel, confirm, main menu, function keys, shortcuts)
- 3 edge cases covered (rapid open/close, ComputerTurn, multi-dialog)
- Risk assessment: all rated Low with specific mitigations
- Rollback plan: 3-step revert with no data impact

The plan is ready for implementation. This should be a quick 1-2 hour task.

PR #986, opened 2026-08-31
GameMenu rewiring, FileDialog tweaks

ajhalme:
- Rewire GameMenu, use signals more consistently
- Fix FileDialog UI event bugs, including #963
- Fix focus bugs that result in Space key triggering buttons when wanting to end turn / skip unit after clicking menu
- Widen FileDialog a bit, so it's less annoying to use


stavrosfa (approved):
This looks fine.
I have been working on having more consistent pop ups (with & without advisors) for some time now, but it will be a while before I actually push something.
The scrollbar PR I have opened is kinda part of that, so this is a good oppurtunity to ask, should we keep the Godot native file dialog, or should we re-create the one we see in the original?
I lean towards the latter, since we can have full control over looks & functionality.

PR #987, opened 2026-08-31
Tech box UI fix

stavrosfa:
Remove the old tech box items, before we add new ones as the new ones are being drawn over the existing ones without the old ones ever being deleted.

(before, courtesy of @WildWeazel )
image


ajhalme (approved):
ok

PR #991, opened 2026-09-11
Extract recursive mouse filter into NodeExtensions; fix popup event leak

### LINKED ISSUE #452: Find a canonical way to block input to the whole screen and/or make blocking modals

Billytifft:
Closes #452.

## Summary

**1. Make modal mouse-filter blocking reusable (#452)**
`PopupOverlay` had a private static `SetMouseFilter` that recursed through a subtree setting the Godot `MouseFilter`. This moves it to a public extension method, `NodeExtensions.SetMouseFilterRecursive(this Node, Control.MouseFilterEnum)`, so any code can generically block/restore input to a subtree. `PopupOverlay.Isolate()`/`Reconnect()` now call it; the private helper is removed (along with a redundant `control.MouseFilter` assignment that the recursion already handled).

**2. Fix a popup signal-handler leak (found while testing #452)**
`TileInfoPopup` subscribes to `PopupOverlay.Click` to dismiss on an outside click (`overlay.Click += Close`) but never unsubscribes. Because the closed popup is only removed from the tree (not freed) and the delegate keeps it alive, the stale handler persists. Concretely: after opening any tile-info popup once, clicking outside a later modal (e.g. the Escape quit menu) would invoke the stale handler, which calls `HideTileInfo()` → `OnHidePopup()`, closing the modal that should have stayed open. This unsubscribes in `_ExitTree` (guarded by `IsInstanceValid` for app teardown), so the tile-info dismiss behavior remains but can no longer affect other popups.

**3. Ignore local agent docs**
Adds `/doc/agent/` and `/AGENTS.md` to .gitignore — local contributor/assistant documentation kept out of the repo until maintainers want it.

## Testing

- `dotnet build C7/C7.sln`: 0 errors.
- `dotnet test C7/C7.sln` with `CIV3_HOME` set: 67 passed, 0 failed, 0 skipped.
- Manual, in-game (Godot 4.4.1):
  - Modal popups (Escape quit menu, build-city, advisors) still block background input while open and restore it on close.
  - Tile-info popup still dismisses on an outside click.
  - Regression for the leak: open a tile-info popup, dismiss it, then open the Escape quit menu — clicking outside now correctly leaves the menu open.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,7 @@ msbuild.wrn
 
 # Visual Studio 2015
 .vs/
+
+# Local agent/contributor docs (keep out of the repo until maintainers want them)
+/doc/agent/
+/AGENTS.md
--- a/C7/UIElements/NodeExtensions.cs
+++ b/C7/UIElements/NodeExtensions.cs
@@ -1,0 +1,17 @@
+using Godot;
+
+public static class NodeExtensions {
+	/// <summary>
+	/// Recursively set MouseFilter on the node and all of its descendants.
+	/// </summary>
+	/// <param name="node">Root of the subtree to update.</param>
+	/// <param name="filter">MouseFilterEnum to apply to every Control in the subtree.</param>
+	public static void SetMouseFilterRecursive(this Node node, Control.MouseFilterEnum filter) {
+		foreach (var child in node.GetChildren()) {
+			child.SetMouseFilterRecursive(filter);
+		}
+		if (node is Control control) {
+			control.MouseFilter = filter;
+		}
+	}
+}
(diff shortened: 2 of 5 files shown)


.gitignore
Billytifft, line 1: These files/directories are used locally for documents used by AI coding assistants. They can be added to the repo and refinded with group consensus.

PR #992, opened 2026-09-14
Add support for impassable terrain

### LINKED ISSUE #362: Add support for "Impassable" flag
While testing movement in WWII in the Pacific, I noticed that units could move infinitely on desert.  After enough testing to determine this was probably not just a bug, I checked the BIQ, and found that desert has 0 movement cost, but is impassable.  Which explained the C7 behavior.

Eventually we should add impassable by wheeled too, but just plain impassable is where to start.

Billytifft:
Closes #362

## Summary
Adds the `Impassable` terrain flag from the BIQ `TERR` section to the terrain model and makes unit movement respect it. This fixes terrains like the WWII Pacific scenario's desert, which has a movement cost of 0 but must block entry entirely.

`ImpassableByWheeled` is deliberately left for a follow-up, as scoped in the issue.

## Changes
- **TerrainType**: new `impassable` property, imported from `civ3Terrain.Impassable`
- **Tile**: `IsImpassable()` helper, reading the overlay terrain (consistent with `MovementCost()`)
- **MapUnit.ResolveIntent**: returns `Disabled` on impassable tiles, which gates every entry path:
  - manual movement (`MapUnit.Move`)
  - go-to pathing (A* `CanEnterPeacefully`/`CanEnterForcefully`)
  - go-to execution (`MoveAlongPath`)
  - go-to UI (goto cursor / hover cost)
- **Tests**: `MakeImpassableDesertTile()` helper plus two tests — a land unit can't enter an impassable tile (even forcefully) and can't path *through* one

## Verification
- `dotnet build C7/C7.sln` — 0 errors
- `dotnet test C7/C7.sln` with `CIV3_HOME` set — 69/69 pass, 0 skipped (real Civ3 scenario/save tests included)
- `dotnet format C7/C7.sln whitespace --verify-no-changes` — clean

Standard BIQs have no impassable terrain, so this is a no-op for normal games; it only affects mods that flag terrain impassable.


EngineTests/Utils/MapBase.cs
```diff
@@ -182,6 +182,12 @@ protected Tile MakeDesertTile() {
 			overlayTerrainType = new() { Key = "desert", movementCost = 1 }
 		};
 	}
+	protected Tile MakeImpassableDesertTile() {
+		return new(ID.None("")) {
+			baseTerrainType = new() { Key = "desert" },
+			overlayTerrainType = new() { Key = "desert", movementCost = 1, impassable = true }
```
ajhalme, line 188: #362 explicitly mentions 0 movementCost. That could be a good test case to check.

PR #994, opened 2026-09-15
Allow barbarian camp tiles to be entered on impassable terrain

### LINKED ISSUE #993: Allow barbarian units to enter impassable terrain
Follow-up to [#992](https://github.com/C7-Game/OpenCiv3/pull/992).

PR #992 added impassable terrain support, which prevents all units from entering tiles flagged impassable. However, barbarian units spawn on impassable terrain in some scenarios (e.g. WWII Pacific desert camps) and need to be able to re-enter those tiles after moving off.

**Proposed fix:** Add an `IgnoresImpassable` flag to `UnitPrototype` (via `SaveUnitPrototype.Flag`), set it on barbarian unit prototypes by default, and gate the impassable check in `ResolveIntent` on it. This keeps the behaviour configurable: a mod author can add or remove the flag per unit type in the save JSON.

Billytifft:
Closes #993. Follow-up to #992.

Barbarian camps can be placed on impassable terrain (e.g. the WWII mod's desert) and PR #992 blocked all units from entering impassable tiles, stranding a camp's garrison once it leaves its tile — spawning works, but moving back onto the camp does not.

Per review, this treats the exemption as a *camp* property (like a city being implicitly reachable) rather than a barbarian-unit property: in `MapUnit.ResolveIntent`, impassable terrain is only a barrier when the tile has no barbarian camp.

**Change:**
- `MapUnit.ResolveIntent`: `tile.IsImpassable() && !tile.hasBarbarianCamp` → disabled. Impassable terrain stays impassable everywhere except the camp tile itself.
- Engine test verifying a camp tile on impassable terrain can be entered and pathed to.

The broader tag/rule-driven "can a unit enter a tile" engine discussed in the issue is tracked as a follow-up design note (directional: ordered unit-tags × tile-tags allow/deny rules, Lua-configurable).


ajhalme:
Looks reasonable.

If we do want to make impassable tiles traversable by barbarians, that should be straightforward to implement via the owner. (This would be problematic as you could then have barb camps that couldn't be easily destroyed as they are unreachable to non-barb land units.)

Issue #864, opened 2026-02-04
City commerce


stavrosfa:
There is something weird going on with how we calculate commerce. 

For example, if we have 5 corrupt gold, we both add it and subtract it from our total income.

The same is true for beakers and happiness.

Check the Inflows() and Outflows(), they both take these values into account
https://github.com/C7-Game/Prototype/blob/9e90b4df5d32097814c6da1379057e4e79265434/C7Engine/C7GameData/Player.cs#L24

I don't really know how commerce should work in game (embarrassing, I know, but it's only been 20 or so years 😅), so I am not sure what changes to make.

ethhics:
If I'm reading it correctly, this is having to do with the ledger calculations for the domestic advisor?

If so, then note that the domestic advisor has a "From cities" category. We don't know, from the available variables, how much gold comes from cities. But, we do know that every city turns its gold exclusively into taxes, beakers, happiness, and corruption. So if you sum those together, they yield the total city gold.

So yes they're double-counted, but it's an artifact of accounting more than anything. The inflow from those sources gets passed directly into outflows as they are really expenses at the end of the day.

Billytifft:
Closing as working-as-intended.

Corruption, science and entertainment being in both `Inflows()` and `Outflows()` is deliberate and matches Civ3 on the Domestic Advisor screen:

- Income shows **gross** city commerce (the full figure, including the part ultimately lost to corruption).
- Science, entertainment and corruption are listed as outflows as well.
- These duplicated terms cancel out in `Netflows()` (e.g. in a probe save: income 19 = corruption 6 + science 9 + entertainment 2 + taxes 2; outflows have the same 6+9+2 plus maintenance, giving a correct net gold-per-turn), so the treasury math was never wrong.

We verified this against screenshots of the same save in Civ3 and OpenCiv3 (currently identical). No code change needed; the ledger semantics can be locked in with regression tests when the advisor is further developed.

PR #995, opened 2026-09-15
Fix double-counting of commerce in treasury flows (#864)

Billytifft:
Fixes the broken income/expense totals on the Domestic Advisor described in #864.

## Problem

`PlayerCommerceBreakdown.Inflows()` and `Outflows()` both included `corrupted`, `beakers`, and `happiness`. Those are allocations of a city's commerce, not treasury gold, so they appeared twice: once on the income side and once on the expense side of the advisor's totals. `Netflows()` happened to be numerically correct because the duplicates cancelled out, so gold-per-turn was never actually wrong.

## Fix

- `Inflows()` is now gold income only: `taxes + taxmenTaxes + fromOtherCivs + interest + wealthProduction`
- `Outflows()` is now gold expenses only: `toOtherCivs + maintenance + unitSupport`
- Added `CommerceTotal()` for the full pre-split city commerce figure
- Domestic Advisor: "From cities" shows `taxes + wealthProduction` (real gold from cities) instead of the old total-commerce `CityInflows()`, and the expense lines keep the exact line order Civ3 uses (Science, Entertainment, Corruption, Maintenance, Unit costs, To other civs)

## Tests

Added `PlayerCommerceBreakdownTest` covering the gold-only inflow/outflow math and `CommerceTotal`. Full suite passes locally (73 tests, Civ3 install enabled).

No behavioral change to gold-per-turn; only the reported breakdown is corrected.


ajhalme:
I don't think this is quite right.

We want to show gross income for cities -- i.e., an income including the part that is ultimately consumed by corruption -- in the "inflows" in Domestic Advisor. This change, as propsed, makes it show the net value, which we already show as gold per turn.

**Civ3**
Screenshot 2026-09-19 at 14 19 06

**OC3 current**
Screenshot 2026-09-19 at 14 38 42

**OC3 proposed**
Screenshot 2026-09-19 at 14 39 20

Billytifft:
youre absolutely right. I shouldve tested this code before putting the PR up. that was my bad. working on the correct fix.

Billytifft:
Closed: current behavior matches Civ3 (gross city income). No code change needed.

PR #998, opened 2026-09-17
Make TileDirection.INVALID safe to use (#397)

### LINKED ISSUE #397: Adding a new `TileDirection.INVALID` value causes existing code to crash
I attempted to add `TileDirection.INVALID`, and without changing any other code, the game crashes with the following error when trying to move a unit to a new tile:

```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'Invalid TileDirection')
   at C7GameData.Tile.HasRiverCrossing(TileDirection dir)
```

This is strange because I did not reference the `INVALID` enum anywhere in code other than adding it to the enum declaration.

Billytifft:
Closes #397

## Summary

Adding a `TileDirection.INVALID` value used to crash the game (e.g. `System.ArgumentOutOfRangeException` in `Tile.HasRiverCrossing`) because several places iterated the enum with `Enum.GetValues(typeof(TileDirection))`, leaking the new member into the map's `tile.neighbors` dictionary and from there into movement/pathing.

This PR:

1. Adds `TileDirection.INVALID` — appended at the **end** of the enum, so the existing `% 8` arithmetic in `MapGenerator` and the serialized `facingDirection` values in `SaveUnit` are unaffected.
2. Adds `TileDirectionExtensions.All` (an explicit array of the 8 real directions) and switches `GameMap.computeNeighbors` and `ImportCiv3` to iterate it instead of `Enum.GetValues(...)`, so `INVALID` can never be injected into the neighbors dictionary or tile knowledge.
3. Makes the switch-style helpers `INVALID`-safe instead of throwing: `HasRiverCrossing` returns `false`, `Reversed` / `RotatedCounterClockwise90Degrees` return `INVALID`, and `ToCoordDiff` returns `(0, 0)`.

Verified locally: `dotnet build` (0 errors), `dotnet test` (56 passed, 13 skipped), `dotnet format C7/C7.sln whitespace --verify-no-changes` (clean).

## Question for reviewers

`C7/Animations/AnimationManager.cs:143-144` carries a TODO left over from this exact issue — the author wanted to add `TileDirection.INVALID` as the fallback for `flicRowToAnimationDirection` but couldn't because adding the value crashed the game (the very bug #397 tracks). That fallback currently silently returns `TileDirection.SOUTHEAST`.

With this PR, `INVALID` is safe to use, so that TODO can finally be implemented (the fallback could return `TileDirection.INVALID` instead of guessing `SOUTHEAST`). I left `AnimationManager.cs` untouched to keep this PR small — would you prefer that TODO be implemented **in this PR** or in a **separate one**?


ajhalme:
Seems reasonable. Not sure if an explicit Invalid enum is that much better than just raising an error, which we currently have, but I suppose this can reduce crashing in some unexpected situations.

The AnimationManager change should go in this PR as well, as that's the driver for the change, the comment, and the GH issue. Returning `INVALID` instead of `SOUTHEAST` for the animation "row" seems reasonable: if there's an unexpected animation that we don't handle, it shouldn't interfere with the others.

Billytifft:
Ok. ill work on adding the animation manager change as well.

Billytifft:
Added the `AnimationManager` change in commit `46c2b27f` and updated the PR description: `flicRowToAnimationDirection` now returns `TileDirection.INVALID` for unexpected rows (instead of silently guessing `SOUTHEAST`), resolving the TODO left over from #397.

